PR #2, opened 2016-09-02
WP 2 Upgrade RxJS to v5

mmcgahan:
https://meetup.atlassian.net/browse/WP-2


middleware/auth.js
```diff
@@ -19,14 +18,6 @@ import {
 export const ANONYMOUS_AUTH_APP_PATH = '/anon';
 
 /**
- * login sub responds to only the most recent login request, and can be disposed
- * by a logout
- * @const
- */
-const loginSub = new Rx.SerialDisposable();
-loginSub.setDisposable(Rx.Disposable.empty);
```
mmcgahan: this was orphaned code - no longer being used.

middleware/sync.js
```diff
@@ -23,11 +23,12 @@ import { activeRouteQueries$ } from '../util/routeUtils';
...
- * navRenderSub lives for the entire request, but only responds to the most
- * recent routing action, so it's a module-scoped 'SerialDisposable', which
- * will take care of disposing previous subscriptions automatically
+ * We want to make sure there is always and only one data-fetching subscription
+ * open at a time, so we create an empty subscription that is scoped to the
+ * request. When a new fetch happens, this subscription will get replaced with
+ * a new subscription.
  */
-const navRenderSub = new Rx.SerialDisposable();
+let apiFetchSub = new Rx.Subscription();
```
mattkime: `const` 😄

mmcgahan: not here - it gets re-assigned to a new subscription each time we need to do a new fetch, so it has to be `let`

mattkime: :+1:

mattkime: we'll need to go over what this code does and why it does it. i should probably spend some time stepping through it first so i'm somewhat familiar.

middleware/sync.js
```diff
@@ -23,11 +23,23 @@ import { activeRouteQueries$ } from '../util/routeUtils';
 import { fetchQueries } from '../util/fetchUtils';
 
 /**
```
sadafie: Are we doing jsdoc in this repo?

middleware/sync.js
```diff
@@ -36,51 +48,58 @@ const navRenderSub = new Rx.SerialDisposable();
...
+	const doFetch = serialSubscription(
+		actions.apiSuccess,
+		actions.apiError,
+		actions.apiComplete
+	);
 
-		activeQueries$.subscribe(dispatchApiRequest);
-	}
+	return next => action => {
+		if (action.type === LOCATION_CHANGE ||  // client nav
```
sadafie: Something that would be helpful is to have more inline comments that describe what each block of code is doing. i.e., what is this return statement? or why is a locationSync happening on line 84? etc.

sadafie:
The code changes themselves LGTM. I'm still also a little unclear with a lot of the RxJS code, but I think I understand it at a high level.  Perhaps we can do various "info sessions" on the code in this repo. i.e., a session per middleware or something, as we did with docker?

middleware/sync.js
mmcgahan: It's not set up to actually generate docs right now, and I'm ambivalent about adding that in, but I can open a task for it if we decide it would be valuable. I like the amount of "let's add some code comments about this" comments coming through PRs at the moment.

sadafie:
it's been open for awhile now. going to merge.

PR #3, opened 2016-09-03
WP 3 redux-observer middleware

mmcgahan:
This PR converts SyncMiddleware to be a set of ["Epics"](https://github.com/redux-observable/redux-observable/blob/master/docs/basics/Epics.md) that are consumed by the [epicMiddleware](https://github.com/redux-observable/redux-observable/blob/master/docs/basics/SettingUpTheMiddleware.md) created by [redux-observer](https://github.com/redux-observable/redux-observable)

The best way to review what is going on is to put the old `middleware/sync.js` code alongside the new `epics/sync.js` code - the functionality is identical, but the new epic-based code is completely based on RxJS/observables rather than a mix of imperative and functional code.

Part of this change is just stylistic - the old code used RxJS for certain async functionality, but redux-observable adapts the whole Redux Action-dispatching functionality to a functional/reactive structure. I really like the new setup because it feels more 'pure' and explicit - there is minimal state and each operation is encapsulated in a sequence of transformations, so the input-output operations are clearly laid out.

The main downside is that by using RxJS more, there is a heavier burden on developers to understand RxJS and what the [Observable operators](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html) do. Ultimately, I think this is a worthwhile dev requirement and will yield more bug-free code.


mmcgahan:
When this merges, we'll need to bump the version to `0.3.0`. The follow-on bugs can be included in that tag if they're reviewed quickly, otherwise they can just be included in patch versions.

mmcgahan:
Closing to re-create with `master` target branch

PR #4, opened 2016-09-16
WP 5 move the renderers over from mup-web

mmcgahan:
https://github.com/meetup/mup-web/pull/308 refactored the browser and server rendering functions to be app-agnostic so that they could be moved here to meetup-web-platform. This PR is just a copy-paste of those already-reviewed modules, with updates to the import paths so that they are internally consistent.

For review, please double check that the locations of these scripts make sense - if you think there's a better way to organize them, that's fine.


mmcgahan:
this will need a couple updates once https://github.com/meetup/mup-web/pull/303 merges, since that updates `dom.jsx` and a couple other bits and bobs that are moving into the platform in this branch

sadafie:
@mmcgahan is this no longer prelim since https://github.com/meetup/mup-web/pull/303 merged?

components/dom.jsx
```diff
@@ -0,0 +1,69 @@
...
+		<html>
+			<head>
+				{head.title.toComponent()}
+				{head.meta.toComponent()}
+				{head.link.toComponent()}
+			</head>
+			<body style={{ margin: 0, fontFamily:'sans-serif' }}>
+				<div id='outlet' dangerouslySetInnerHTML={getInnerHTML(appMarkup)} />
+				<script dangerouslySetInnerHTML={getInnerHTML(`window.INITIAL_STATE=${JSON.stringify(INITIAL_STATE_SAFE_JSONABLE)};`)} />
+				<script src={clientPath} />
```
sadafie: Looks like this might need some updates from https://github.com/meetup/mup-web/pull/303 ?

renderers/browser-render.jsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
```
sadafie: Could you remind me where this is used? I'm guessing in `mup-web` app code somewhere?

renderers/server-render.jsx
```diff
@@ -0,0 +1,172 @@
+import Rx from 'rxjs';
```
sadafie: As far as directory structure goes, since this is the server entry point, I personally find it a little strange/confusing that it lives in this sub-directory alongside browser-render.jsx. 

Could this file just live top level, and maybe move `browser-render` until `utils/` or somewhere else?

util/createStore.js
```diff
@@ -0,0 +1,49 @@
+/**
```
sadafie: Will these files under `util/` be removed from `mup-web` or do they now live in both places?

sadafie: I just looked at the corresponding PR in `mup-web` and now understand.

renderers/server-render.jsx
mmcgahan: the server entry point(s) are in the individual apps, which have `src/server.js` or `src/server-locale` as the entry point for the locale bundles and `scripts/start.js` or whatever for the app server bundle.

`server-render.jsx` and `browser-render.jsx` just provide rendering functions to the app's entry points

components/dom.jsx
mmcgahan: I think we're good now - mup-web#303 took out the `CONFIG` global var and made it possible to inject `clientPath` as a prop for `<Dom>`

renderers/browser-render.jsx
mmcgahan: Yup, in the client.js entry point: https://github.com/meetup/mup-web/blob/master/src/client.jsx#L1

renderers/server-render.jsx
sadafie: Okay, sorry. Some of this stuff is getting a little confusing w/ the refactoring + two repos.  I did some digging to make sure I actually understand this.

- I *think* one confusion for me was that in the mup-web app we build `build/server-locales/<locale>/serverRender.js` as the app's server entry which I assumed was `src/server-render.jsx`, but I see now is `src/server.js`.

- Another source of confusion:

> the server entry point(s) are in the individual apps, which have src/server.js or src/server-locale as the entry point for the locale bundles and scripts/start.js or whatever for the app server bundle.

I didn't even realize `build/server-locale/` was a thing now! 

- My takeaway:

I think it just comes down to the names of the source code files and the build output files in the application code at this point, since `build/server.js` corresponds to `scripts/start.js` and `build/server-locale/<locale>/serverRender.js` corresponds to `scripts/server.js`. Could we have corresponding output names?

- instead of `scripts/start.js` have `scripts/server.js` => `build/server.js`
- instead of `scripts/server.js` have `scripts/serverRender.js` => `build/server-locale/.../serverRender.js`

Or something similar.

mmcgahan: yes, that makes sense - I'll update this PR and the corresponding files in mup-web#312

mmcgahan: Since these scripts relate closely to the scripts that are affected by the parallel build PR ([mup-web#321](https://github.com/meetup/mup-web/pull/321)), I've made the name changes there. I think these are really good updates for clarifying what goes where, although it has made the mup-web PR a bit larger. I think that was very close to being mergeable, but hopefully that can happen soon so there are fewer big file moves and var name changes hanging around.

sadafie: yep yep. let me do one more quick pass at this. i think this is mergeable. then i'll jump over to mup-web.

sadafie:
LGTM!

PR #6, opened 2016-09-23
WP 7 runtime publicpath

mmcgahan:
Consume client filename and asset server public path in server-render so that these values can be applied at runtime when building the app.


sadafie:
LGTM

PR #7, opened 2016-09-26
Tracking uuid and cookie logic

sadafie:
Fixes https://meetup.atlassian.net/browse/WP-8

Request logic to create tracking cookie with a random uuid.

### DIFF
--- a/actions/configActionCreators.js
+++ b/actions/configActionCreators.js
@@ -5,3 +5,9 @@ export function configureApiUrl(url) {
 	};
 }
 
+export function configureTrackingId(id) {
+	return {
+		type: 'CONFIGURE_TRACKING_ID',
+		payload: id
+	};
+}
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "jest-cli": "^14.1.0",
     "js-cookie": "^2.1.2",
     "node-fetch": "^1.6.0",
+    "node-uuid": "^1.4.7",
     "react": "^15.3.1",
     "react-addons-test-utils": "^15.3.1",
     "react-helmet": "^3.1.0",
--- a/renderers/server-render.jsx
+++ b/renderers/server-render.jsx
@@ -5,18 +5,22 @@ import React from 'react';
 import ReactDOMServer from 'react-dom/server';
 import RouterContext from 'react-router/lib/RouterContext';
 import match from 'react-router/lib/match';
-import { Provider } from 'react-redux';
 import RedBox from 'redbox-react';
+import { Provider } from 'react-redux';
 
+import createStore from '../util/createStore';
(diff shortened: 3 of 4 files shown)


sadafie:
@mmcgahan  From [Mup-web #300](https://github.com/meetup/mup-web/pull/300#issuecomment-247134315):

>If you want to look into writing a standalone Hapi plugin, check out the docs here and use anonAuthPlugin as an example, but let's maybe set up a meeting to talk through options once you've looked through it a bit more

I spent some time looking more closely at this and the ideas you suggested for the Request State Manager. I'd like to do this part as an immediate follow on if you think that's okay. I'll set up a time for us to chat tomorrow?

mmcgahan:
> do this part as an immediate follow on

sounds good

mmcgahan:
I think this needs a spec for its lifecycle - currently it doesn't change in response to log in/out, but I think that's a critical feature? We'll probably want to hold off until we can align/consolidate this with the auth logic, which has an almost identical lifecycle

sadafie:
>I think this needs a spec for its lifecycle - currently it doesn't change in response to log in/out, but I think that's a critical feature? We'll probably want to hold off until we can align/consolidate this with the auth logic, which has an almost identical lifecycle

Yes, we had a brief chat with @bgruber last week. Per his suggestion, I'm going to be setting up some time with Tom M on the data analytics side. We'll probably want things to work differently on this platform than it does in chapstick, and he probably has some ideas on what he'd like that to look like.

Depending on how long it takes to nail it down, I'd rather not keep this open for a long time. Since this is just setting a trackingId cookie that we aren't consuming for anything, I'm not concerned of merging as is. LMK what you think, and if you want to be in the tracking discussions with the data team.

mmcgahan:
it sounds like at a minimum the cookie should correspond to a user session, which it doesn't do right now because a log in/out event will change the user without changing the cookie, so i think this code change is not ready to merge.

sadafie:
>it sounds like at a minimum the cookie should correspond to a user session, which it doesn't do right now because a log in/out event will change the user without changing the cookie,

just fyi, you would think that's a valid assumption! but i recently learned that in chapstick (for example) that doesn't happen. we at some point intentionally decided to keep the tracking cookie on logout to continue the logged out tracking (and yes, it can result in edge cases like dupe ids for different members). that being said, i'm pretty sure that's a thing we no longer want to do, but it just seems unclear to me until we have a follow up with tom and brian.

i can link the cookie to the auth in this branch, but in my mind there's a good chance that whatever we do here will change after we do some more investigation.

mmcgahan:
ah, that's a behavior i definitely would not have guessed. will look over the code one more time but will probably go ahead and merge

PR #8, opened 2016-09-27
WP 12 cache observable

mmcgahan:
Same as WP-3, but for Cache Middleware

**Note** this PR currently targets #3, so probably best if it merges there _after_ WP-3 is approved, but _before_ it merges to master. If that's awkward, I'll just create a new PR targeting master after #3 merges.

This looks like a lot of changes, but it's really just a refactor with almost no effect on functionality.

The refactor makes it clear that the cache can only do 3 things:

1. Clear (destroy completely)
2. Set
3. Query (this is an async getter, but it specifically responds to reactive queries) 

Features of note:

1. Some of the utility functions for CRUD'ing the cache have been moved to `util/cacheUtils.js` to keep the cache epic a little more well-defined. Bonus: if we want other caches for other parts of the app, we can import the utilities directly. _Nothing was changed in these functions_
2. I've eliminated direct dispatch of some of the 'intermediate' actions like `CACHE_CLEAR` and `CACHE_SET`, which were kind of helpful for cataloging activity within the application, but meant that the cache middleware was adding some noise to the action logs without much practical benefit. The cache epic is now a little more isolated from the rest of the application than before, with the expectation that only `CACHE_SUCCESS` is ever emitted from any of the Epics, because it's the only action that has implications for the rest of the application. `cacheClear` and `cacheSet` still exist as action creators so that there is a direct way of triggering cache activity (e.g. for testing or clearing the cache on logout).


mattkime:
👍

epics/cache.js
```diff
@@ -0,0 +1,100 @@
...
+export function checkEnable() {
+	if (typeof window !== 'undefined' && window.location) {
+		const params = new URLSearchParams(window.location.search.slice(1));
+		return !params.has('__nocache');
+	}
+	return true;
+}
+
+/**
+ * The cache middleware triggers a 'set'/store action when new data is received
```
sadafie: s/middleware/epic/

epics/cache.js
```diff
@@ -0,0 +1,100 @@
+/**
+ * Provides a cache outside of Redux state that can optimistically update state
+ * before an asynchronous API call returns
+ *
+ * @module CacheMiddleware
+ */
+import Rx from 'rxjs';
```
sadafie: Nitpick, in some epics you import `Rx` others you import `Observable` directly. Do we care to be consistent? (see #10 )

epics/cache.js
```diff
@@ -0,0 +1,100 @@
...
+ * `{ queries, responses }` payload
+ *
+ * API_SUCCESS means there is fresh data ready to be stored - extract the
+ * queries and their responses, then dispatch `CACHE_SET` actions with each
+ * pair
+ *
+ * Not that this will set the cache without emitting an action
+ */
+export const cacheSetEpic = action$ =>
+	action$.ofType('API_SUCCESS', 'CACHE_SET')
```
sadafie: For these cache epics, you mentioned "...without emitting an action". Does that mean that the incoming actions in `action$` never reach the reducer? Or do they and the reducers are now a no-op? Just curious of the order of operations with these actions + epic middleware + redux reducer.

epics/cache.js
```diff
@@ -0,0 +1,100 @@
+/**
```
sadafie: Does this epic need a test file?

epics/cache.js
mmcgahan: That's my ambivalence on import style leaking through. I don't have a strong preference for either style, or a particularly strong need to be consistent, _but_ I'm pretty sure there are some ways of updating these imports so that you don't have to import the whole library, which can make the bundle smaller - I'll open a Jira ticket to review the imports and we can hopefully get everything nice and tidy then

epics/cache.js
mmcgahan: From https://github.com/redux-observable/redux-observable/blob/master/docs/basics/Epics.md:

> Epics run alongside the normal Redux dispatch channel, after the reducers have received them, so you cannot "swallow" an incoming action. When you map one of those actions to another, you're not preventing the original action from reaching the reducers; that action has already been through them!

epics/cache.js
mmcgahan: I've put all the epic tests into `middleware/epic.test.js` in order to simulate `dispatching` the actions, but I've just realized that's not necessary since Epics are pure function that can be tested independently of the middleware... is it alright with you if I file a followup with convert the `epic.test.js` unit tests to epic-specific test scripts?

epics/cache.js
sadafie: oh interesting. the video i just watched by jay phelps mentioned that the order is set by your middleware. which i guess i interpreted as the order is set by the epic. if it's set by the middleware, it must mean the epic middleware is sending it through the reducer first then through the epic

PR #9, opened 2016-09-27
WP 13 convert auth middleware to auth epic

mmcgahan:
**Please review/merge #10 before this PR** - this branch includes the commits from #10 

Auth epic using redux-observable. Direct conversion, no news.

This will conflict with #8, but only because both of them are adding a new epic to the epic middleware - resolution should be simple (take both updates together), but just a heads up.


epics/auth.js
```diff
@@ -0,0 +1,114 @@
+/**
+ * Auth middle handles all the server communication and cookie managmenet
```
sadafie: s/middle/epic/

PR #10, opened 2016-09-27
WP 3 redux-observable middleware

mmcgahan:
**Please review/merge #8** before merging this PR

When this merges, we'll need to bump the version to 0.3.0. The follow-on bugs can be included in that tag if they're reviewed quickly, otherwise they can just be included in patch versions.

This PR converts SyncMiddleware to be a set of ["Epics"](https://github.com/redux-observable/redux-observable/blob/master/docs/basics/Epics.md) that are consumed by the [epicMiddleware](https://github.com/redux-observable/redux-observable/blob/master/docs/basics/SettingUpTheMiddleware.md) created by [redux-observer](https://github.com/redux-observable/redux-observable)

The best way to review what is going on is to put the old `middleware/sync.js` code alongside the new `epics/sync.js` code - the functionality is identical, but the new epic-based code is completely based on RxJS/observables rather than a mix of imperative and functional code.

Part of this change is just stylistic - the old code used RxJS for certain async functionality, but redux-observable adapts the whole Redux Action-dispatching functionality to a functional/reactive structure. I really like the new setup because it feels more 'pure' and explicit - there is minimal state and each operation is encapsulated in a sequence of transformations, so the input-output operations are clearly laid out.

The main downside is that by using RxJS more, there is a heavier burden on developers to understand RxJS and what the [Observable operators](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html) do. Ultimately, I think this is a worthwhile dev requirement and will yield more bug-free code.


epics/sync.js
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Navigation actions will provide the `location` as the payload, which this
+ * epic will use to collect the current Reactive Queries associated with the
+ * active routes.
+ *
+ * These queries will then be dispatched in the payload of `apiRequest`
+ * @param {Object} routes The application's React Router routes
+ * @returns {Function} an Epic function that emits an API_REQUEST action
+ */
+export const getNavEpic = routes => {
```
sadafie: I thought epics had to be "actions in, actions out", but this is taking the routes as an argument instead of a stream of active routes.

sadafie: oh pfffttt, ignore me, i just saw this is returning a function

mmcgahan: Yeah - the only flag that this is a higher order function is the `get` prefix that isn't on the other epics, but hopefully it won't throw people off too much.

sadafie: Yeah, sorry I totally missed that. Makes sense.

mmcgahan:
> If you don't understand / haven't used one or both, this code is just not going to make any sense

That's true, and I think in our discussions as a team we've leaned toward saying that, for the _platform_ code, understanding RxJS and Redux is a requirement - the platform is always going to have a higher expectation for its maintainers because it generally deals with more async data transformation than the apps it supports. The one thing that I hope will help is that each step of each observable stream is a pure function, which should, in isolation, be understandable for anyone, regardless of third-party framework/library knowledge.

> I wonder if this is a good use case for epic middleware.

I think the platform middleware meet the criteria for conversion to Epics, in part because they will only become more complex over time - at the moment, they're very much MVP, although they're already leveraging some nice features of redux-observable like cancelation for API requests. For me, epics are so much easier to follow than middleware, mainly because the imperative style of middleware is so much less organized by default.

The alternative is having some middleware and some epics and trying to make those available in a coherent way as modules for apps - I'm kind of thinking that once all the middleware is converted, we can essentially expose the 'EpicMiddleware' as 'PlatformMiddleware', and apps can import that without having to think too much about what it contains and how it's implemented... but that's a discussion for later.

PR #11, opened 2016-09-28
WP 16 get travis up and running to unit test PRs

mmcgahan:
`Prelim` until I've tested a few times to see what's appearing in the logs and how it behaves in general

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,69 @@
+udo: required
+language: bash
+
+# exclude tags
+branches:
+  except:
+    - /^\d+$/
+    - /^v[\d.]+$/
+
+services:
+  - docker
+
+env:
+  global:
+    # build number used by blt pipeline
+    - CI_BUILD_NUMBER=$TRAVIS_BUILD_NUMBER
+    # disables gcloud installation prompts
+    - CLOUDSDK_CORE_DISABLE_PROMPTS=1
+
+before_install:
+  # Install envtpl for easy templating.
+  - sudo pip install envtpl
+
+  # Login to mup.cr to publish artifacts
+  - docker login --username $DOCKER_LOGIN --password $DOCKER_PASS mup.cr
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -0,0 +1,65 @@
...
+language: bash
+
+# exclude tags
+branches:
+  except:
+    - /^\d+$/
+    - /^v[\d.]+$/
+
+services:
+  - docker
```
mattkime: won't need this!

Makefile
```diff
@@ -0,0 +1,21 @@
+NPM_TOKEN ?= '00000000-0000-0000-0000-000000000000'
+CI_BUILD_NUMBER ?= $(USER)-snapshot
+VERSION ?= 0.2.$(CI_BUILD_NUMBER)
+
+test:
+	npm install --silent
+	npm test
+
+package:
+	docker run \
```
mattkime: Looks like you need a Dockerfile...

mmcgahan:
Need to investigate how `package.json` is getting updated.

Makefile
mmcgahan: I thought I had left a reply to this, but looks like it didn't stick.

With the way this was set up, there was no need for a Dockerfile or `docker pull` - the `docker run` command took care of all that (I think the `docker pull` was unnecessary because `node:6` is in the Docker registry so it was pulled automatically from there).

However, I've gone a completely different route with this that makes it a lot simpler by just using Travis's [built-in support](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/) for Node-based projects that just need to run `npm test` - no need for Docker at all, although I have kept the Makefile just for the npm versioning helper. I'd like to avoid the complication of Docker for this repo in part because it's not "runnable", so containerization is a little overkill for running tests - I'm much more comfortable with the new setup.

PR #15, opened 2016-09-30
WP 22 jest upgrade to v15 and add coverage reports

mmcgahan:
upgraded a bunch of npm packages by minor versions, plus Jest to v15.

Interestingly, I found that Travis integrates really well with Coveralls and so it doesn't need the COVERALLS_REPO_TOKEN to be explicitly set as an environment variable - we can potentially update our Travis settings for mup-web to remove that private env var

https://coveralls.io/github/meetup/meetup-web-platform


sadafie:
Cool, seems straightforward. 👍

PR #19, opened 2016-10-06
Unit tests for routes

mmcgahan:
The meetup-web-platform `routes.js` was almost completely untested, and requires mocking a Hapi request using `server.inject`, which we haven't done before.

I went ahead and set up some basic tests that greatly increase test coverage, but there's more to do.

@sadafie it would be good to add some tests for `initTrackingCookie`

`Prelim` until I can review everything and add some documentation. Lots of opportunities for refactoring into more testable units...

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "description": "A collection of Node web application utilities developed for the Meetup web platform",
   "scripts": {
     "coveralls": "cat coverage/lcov.info | coveralls",
+    "lint": "eslint . --ext .js,.jsx --fix",
     "test": "jest --silent --config jest.config.json --coverage"
   },
   "repository": {
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "request": "2.75.0",
     "rxjs": "^5.0.0-beta.11",
     "source-map-support": "0.4.3",
+    "tough-cookie": "2.3.1",
     "url-search-params": "0.6.1"
   }
 }
--- a/routes.js
+++ b/routes.js
@@ -50,10 +50,11 @@ export default function getRoutes(
 	{
 		API_SERVER_ROOT_URL,
 		PHOTO_SCALER_SALT,
-	}) {
+	},
+	apiProxyFn$ = apiProxy$) {
 
 	console.log(chalk.green(`Supported languages:\n${Object.keys(renderRequestMap).join('\n')}`));
-	const proxyApiRequest$ = apiProxy$({
+	const proxyApiRequest$ = apiProxyFn$({
 		baseUrl: API_SERVER_ROOT_URL,
 		duotoneUrls: getDuotoneUrls(duotones, PHOTO_SCALER_SALT),
 	});
--- a/routes.test.js
+++ b/routes.test.js
@@ -1,106 +1,62 @@
-import { Cookie } from 'tough-cookie';
-import { Observable } from 'rxjs';
-import Hapi from 'hapi';
 import getRoutes from './routes';
 import getConfig from './util/config';
+import {
+	MOCK_API_RESULT,
+	MOCK_OAUTH_COOKIES,
+	MOCK_renderRequestMap,
+	MOCK_API_PROXY$,
+	MOCK_RENDER_RESULT,
+	MOCK_REQUEST_COOKIES,
+} from './util/mocks/app';
+import {
+	parseCookieHeader,
+	getServer,
+} from './util/testUtils';
 
-const server = new Hapi.Server();
-
-server.connection();
-
-// mock the anonAuthPlugin
-server.decorate(
-	'request',
-	'authorize',
-	request => () => Observable.of(request),
-	{ apply: true }
-);
-
+// RegEx to verify UUID
 const UUID_V4_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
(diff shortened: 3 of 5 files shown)


sadafie:
> @sadafie it would be good to add some tests for initTrackingCookie

d'oh. yes, will do that today.

package.json
```diff
@@ -47,6 +47,7 @@
     "request": "2.75.0",
     "rxjs": "^5.0.0-beta.11",
     "source-map-support": "0.4.3",
+    "tough-cookie": "2.3.1",
```
sadafie: 💪 🍪

routes.test.js
```diff
@@ -0,0 +1,89 @@
...
+);
+
+const MOCK_RESULT = 'asdf';
+const MOCK_RENDER_REQUEST = () => Observable.of({ result: MOCK_RESULT, statusCode: 200 });
+
+const MOCK_renderRequestMap = {
+	'en-US': MOCK_RENDER_REQUEST,
+};
+
+const MOCK_OAUTH_COOKIES = {
```
sadafie: should some of these mock data objects go in `util/mocks/`?

routes.test.js
```diff
@@ -0,0 +1,89 @@
...
+	refresh_token: 'asdf',
+	anonymous: true
+};
+
+const MOCK_API_RESULT = {
+	queries: [],
+	responses: [],
+};
+
+const serverReady = getConfig()
```
sadafie: Unrelated to this PR, but this is the first time i'm noticing that `config.js` is a Promise based interface. Any plans on using observables here instead?

routes.test.js
```diff
@@ -0,0 +1,89 @@
...
+	.then(config => getRoutes(MOCK_renderRequestMap, config, () => () => Observable.of(MOCK_API_RESULT)))
+	.then(server.route.bind(server));
+
+describe('routes', () => {
+	it('serves the homepage route', () =>
+		serverReady
+			.then(() => server.inject({ url: '/' }))
+			.then(response => expect(response.payload).toEqual(MOCK_RESULT))
```
sadafie: i think i'm just not following this setup. how is the `response.payload` getting set?

routes.test.js
mmcgahan: Yeah I wanted to stub them out here and see what I really needed, but I'll move the final set over to `util/mocks/`

routes.test.js
mmcgahan: In this case, there's not a big advantage to observables since we only care about one 'future' value and we don't need to do any sophisticated transformations.

routes.test.js
mmcgahan: Yeah the tools for mocking Hapi requests are a little opaque, unfortunately, but basically `server.inject` on the previous line creates a fake [Hapi request object](http://hapijs.com/api#request-object), which allows you to inspect the response of your routes for the expected features.

I'll add a bunch of code comments here before I pull off the `Prelim` label

mmcgahan:
re: tests for `initTrackingCookie` - I think it would be best to do that in a branch off of this branch, which created the `routes.test.js` file

sadafie:
>re: tests for initTrackingCookie - I think it would be best to do that in a branch off of this branch, which created the routes.test.js file

Done: #20

routes.test.js
sadafie: Yep, went back and looked the hapi docs. thanks!

routes.js
```diff
@@ -50,10 +50,11 @@ export default function getRoutes(
 	{
 		API_SERVER_ROOT_URL,
 		PHOTO_SCALER_SALT,
-	}) {
+	},
+	apiProxyFn$ = apiProxy$) {
```
sadafie: it feels unkosher to have to change the signature of a method in order to support test cases, but i don't know enough about jest to know if there's any way around this.

util/testUtils.js
```diff
@@ -22,3 +25,30 @@ export const middlewareDispatcher = middleware => (storeData, action) => {
 	return dispatched;
 };
 
+export const parseCookieHeader = (cookieHeader) => {
```
sadafie: nitpick, but i'd like to see consistent function declaration syntax at least within a file? thoughts?

routes.js
mmcgahan: this is actually a big improvement prompted by writing tests IMO - we can make this function more 'pure' _and_ more testable by making this a function arg rather than consuming it from the module scope. By using ES6 default args, it's also not a breaking change, just more flexibility - in theory, you could use this module with any API proxying routine now.

mmcgahan: That said, I'm not super happy with the overall signature for this function now - I think we should probably refactor it to be a single config object arg at some point, but since that's a breaking change I don't want to do it now.

util/testUtils.js
mmcgahan: made the update for this file, although I think in general there are good reasons to use different function declaration syntaxes in different cases. Arrow functions are really nice for currying (`const fn = x => y => z => { ... }`) and when you return immediately (`const fn = x => Object.keys(x).map(key => x[key])`), but I tend to have a slight preference for `function fn(x)` when you have a beefy function block.

This is all mostly a mild aesthetic pref, though, so I'm happy to adjust wherever

sadafie: I don't really have a strong preference. Sometimes with the arrow notation, it's slightly less clear at first glance, so I could also have a preference for explicit function blocks, but I don't truly care that much. 

I just have an OCD around readability when styles are intermixed.

PR #20, opened 2016-10-06
Feature/tracking cookie tests

sadafie:
Add test case for initTrackingCookie logic.

Includes/Depends on #19 

The test will need to be updated as we figure out more of the tracking logic. Works with the code that's in place now.


mmcgahan:
This looks really good and I like how you've pulled out some self-contained functions to clarify what's happening. I think I will merge this into #19 to avoid conflicts while I add comments and such there.

PR #21, opened 2016-10-07
WP 29 convert PostMiddleware to PostEpic

mmcgahan:
This is the last piece of the conversion from middleware to redux-observable epics in the platform. The final epic implementation is almost identical to the previous middleware implementation, with a little more simplicity and terseness provided by destructuring function arguments.

This is a breaking change because it removes the Post Middleware and converts `getEpicMiddleware` to `getPlatformMiddleware`. I actually don't think we'll be adding much more middleware to the platform at all, for better or worse.

As a side note, the very structured nature of the arguments passed around a Redux system (mostly Actions) makes a good argument in favor of introducing some stronger variable typing into the codebase, either through [Flow](https://flowtype.org/) or Typescript (I'd favor Flow)...


sadafie:
This seems pretty straightforward. Nothing stands out as an issue ---- should I go ahead and merge a "breaking change" or does something else need to happen in conjunction?

mmcgahan:
This is mergeable - I think as long as there's a CHANGELOG entry and the minor version is bumped in the `Makefile`, breaking changes are mergeable. I'll file a ticket to upgrade mup-web

PR #22, opened 2016-10-08
WP 31 serve a skeleton version of the app

mmcgahan:
This code update looks for a `?skeleton` query param on any request and returns a bare-bones (get it?) version of the app that will be rendered entirely on the client, which will be the entry point for an offline-first service worker (or if you just hate server rendering, here's your chance to beat the system!).

This still requires setting up the global configuration properties in the initial Redux state so that the client-side app can communicate with the API using server-provided oauth data.


sadafie:
> bare-bones (get it?)

😂😂😂😂

sadafie:
@mmcgahan I was a little disconnected last week, but is this just an investigatory branch or are we for sure going this route?

mmcgahan:
@sadafie we can think about other methods of requesting the skeleton route, but the ability to serve it so that the SW can cache the skeleton is critical to having a 'good' offline/fast load experience. I'm pretty happy with the change for now, though.

renderers/server-render.jsx
```diff
@@ -49,55 +61,53 @@ const DOCTYPE = '<!DOCTYPE html>';
...
-		result
+		return {
+			statusCode,
+			result
+		};
 	};
-}
 
 /**
  * Curry a Redux store and auth tokens to privde a function that can dispatch
```
sadafie: No longer a curry

PR #23, opened 2016-10-14
Fix RedBox rendering of server-side app render errors

mmcgahan:
Our existing setup for returning friendly error messages to the front end doesn't actually work - errors in the server-side rendering of the app markup don't get caught correctly in the Observable stream, and we end up with the super-unhelpful 500 error and impenetrable Observable error logs.

This fixes that.

screen shot 2016-10-15 at 8 14 12 am


mmcgahan:
Going to go ahead and merge this since it's going to conflict with #22 and it's mostly just a code move as opposed to changing functionality. We can revisit the error handling behavior later if we want.

PR #24, opened 2016-10-15
WP 37 - Auth/API Timeouts and better error messages

mmcgahan:
This PR applies auth and API endpoint timeouts that will throw errors when external services don't respond in a reasonable amount of time. It also `catch`es these timeout errors and attempts to gracefully continue request execution. It is up to the consuming application to render errors appropriately.

### DIFF
--- a/apiProxy/api-proxy.js
+++ b/apiProxy/api-proxy.js
@@ -233,7 +233,7 @@ export const apiResponseDuotoneSetter = duotoneUrls => {
  * @param {Object} baseUrl API server base URL for all API requests
  * @return Array$ contains all API responses corresponding to the provided queries
  */
-const apiProxy$ = ({ baseUrl, duotoneUrls }) => {
+const apiProxy$ = ({ API_TIMEOUT=5000, baseUrl, duotoneUrls }) => {
 	const setApiResponseDuotones = apiResponseDuotoneSetter(duotoneUrls);
 
 	return request => {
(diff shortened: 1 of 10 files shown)


apiProxy/api-proxy.js
```diff
@@ -244,14 +245,26 @@ const apiProxy$ = ({ baseUrl, duotoneUrls }) => {
 		// to build the query-specific API request options object
 		const apiConfigToRequestOptions = buildRequestArgs(externalRequestOpts);
 
+		const apiTimeout = 5000;
```
sadafie: Can we stick this at the top, or somewhere where it's more easily discoverable?

plugins/anonAuthPlugin.js
```diff
@@ -2,6 +2,7 @@ import Boom from 'boom';
 import chalk from 'chalk';
 import Rx from 'rxjs';
 
+const AUTH_TIMEOUT = 5000;
```
sadafie: Actually can we use a configuration for this value? So it's defined in fewer spots.

apiProxy/api-proxy.js
```diff
@@ -3,7 +3,6 @@ import Rx from 'rxjs';
 const externalRequest$ = Rx.Observable.bindNodeCallback(externalRequest);
 
 import * as apiConfigCreators from './apiConfigCreators';
-import { catchAndReturn$ } from '../util/rxUtils';
```
sadafie: Method no longer used. Remove from rxUtils?

plugins/anonAuthPlugin.js
mmcgahan: yes - I'll put it [here](https://github.com/meetup/meetup-web-platform/blob/master/util/config.js)

mmcgahan:
oooh, gotta set up standalone tests for `fetchUtils.js` to keep up the coverage! I'll do that in a separate branch, but this PR won't be mergeable until that merges and addresses the coverage decline

mmcgahan:
`Prelim` because I'd like #25 to be merged first (its code changes have been pulled into this branch)

sadafie:
Merged 25, this is looking good too

PR #27, opened 2016-10-21
Redux little router - POC for discussion

mmcgahan:
For your edification: a conversion from React Router to [redux-little-router[(https://github.com/FormidableLabs/redux-little-router), a routing solution that is more "React native" than React Router, and more "Redux native" than "react-router-redux", yielding a simpler, more composable routing/rendering architecture.
### Highlights:

**The `routes` object makes no reference to React components**
- more declarative, fine grained component nesting - instead of having a single `{children}` outlet for nested route components, you can put route-matching `<Fragment>`s anywhere you want, e.g. this is not possible in React Router (v2):
  
  ``` jsx
  <Container>
    <ChildA forRoute='/:nestedRoute' />
  
    <div>Content in between children</div>
  
    <ChildB forRoute='/:nestedRoute' />
  </Container>
  ```
  
  but it's trivial in redux-little-router
- unlike React Router v4, there is still a single `routes` object on which we can hang our Reactive Queries - and unlike with React Router v2, the matching routes are known as soon as the store is initialized - we don't have to go back and do a separate `match` routine

**The `store.getState().router` is a first-class app state citizen** - it is _the_ place to look for routing information, unlike React Router Redux, which encourages users _not_ to reference the routing information stored in state

**Fewer + smaller dependencies**

**Possibly less dependent on consumer architecture** - from this first pass, I think it will be easier to separate out the router-dependent platform code from the router-independent platform code, which might allow consumers to skip routing altogether or choose a different routing architecture if they so choose.

**Sync epic doesn't need any routes information** - the `LOCATION_CHANGED` action will directly provide all the info it needs
### Lowlights

**We will need a followup bug to determine the best pattern for async loading routes**, probably using an additional component that uses `System.import` to fetch its children on `componentDidMount`. This is goodnews/badnews in that it's a more orthodox React way of dealing with async loading, but it's also not 'free' - comparable in difficulty to React Router v2's `getChildRoutes`.

**Major breaking change** - the routing setup expected from the consumer is significantly different than that of React Router v2 (same cost associated with going to React Router v4 though)


mmcgahan:
I'd like to add tests for `routeUtils.js`, which has pretty poor coverage ATM

mmcgahan:
Also lowlight that needs a feature that was _just merged_ into redux-little-router - matching routes based on priority so that overlapping patterns don't all activate at once - will need to implement and check functionality

package.json
```diff
@@ -50,15 +50,19 @@
...
     "redux": "^3.5.2",
+    "redux-little-router": "12.0.0",
     "redux-observable": "0.12.0",
     "request": "2.75.0",
     "rxjs": "^5.0.0-beta.11",
     "source-map-support": "0.4.3",
     "tough-cookie": "2.3.1",
     "url-search-params": "0.6.1"
+  },
+  "peerDependencies": {
```
sadafie: What's the purpose/advantage of making these peer dependencies?

src/components/Link.js
```diff
@@ -0,0 +1,4 @@
+import { Link } from 'redux-little-router';
```
sadafie: Why do we need this and `RouterProvider.js`?

src/epics/sync.js
```diff
@@ -1,33 +1,27 @@
...
-export const getNavEpic = routes => {
-	const activeQueries$ = activeRouteQueries$(routes);
-	return (action$, store) =>
-		action$.ofType(LOCATION_CHANGE, '@@server/RENDER', 'LOCATION_SYNC')
-			.map(({ payload }) => payload)  // extract the `location` from the action payload
-			.flatMap(activeQueries$)        // find the queries for the location
-			.map(apiRequest);               // dispatch apiRequest with all queries
-};
+export const navEpic = (action$, store) =>
+	action$.ofType(LOCATION_CHANGED, 'LOCATION_SYNC')
```
sadafie: You don't need the `@@server/RENDER` action here anymore?

src/epics/sync.js
```diff
@@ -56,15 +50,15 @@ export const getFetchQueriesEpic = fetchQueriesFn => (action$, store) =>
...
 				.catch(err => Observable.of(apiError(err)));  // ... or apiError
 		});
 
-export default function getSyncEpic(routes, fetchQueriesFn=fetchQueries) {
+export default function getSyncEpic(fetchQueriesFn=fetchQueries) {
 	return combineEpics(
-		getNavEpic(routes),
+		navEpic,
 		resetLocationEpic,
 		getFetchQueriesEpic(fetchQueriesFn)
```
sadafie: nitpick, but can you drop the `get` on here and call it `fetchQueriesEpic` for consistency?

src/util/createStore.js
```diff
@@ -19,22 +23,35 @@ const noopMiddleware = store => next => action => next(action);
  * @param {Function} middleware (optional) any app-specific middleware that
  *   should be applied to the store
  */
-function finalCreateStore(routes, reducer, initialState=null, middleware=[]) {
-	// **All** middleware gets added here
-	const middlewareToApply = [
-		getPlatformMiddleware(routes),
-		typeof window !== 'undefined' && window.mupDevTools ? window.mupDevTools() : noopMiddleware,
+function createEnhancedStore(router, reducer, initialState=null, middleware=[], enhancers=[]) {
```
sadafie: What does `EnhancedStore` mean in this context?

package.json
mmcgahan: After reading more about how peerDependencies are supposed to be used, I think this is the correct way - basically, the platform is going to import some libs that it uses to set up the render, and the consumer is going to import the same libs for some of its rendering scripts, so the platform needs to make sure the same lib is being imported in both places by specifying them as `peerDependencies`.

Without this, the server-side `dom.jsx` might be rendered with React version 15 and the client will supply an `<App>` that renders with React version 20, which would be a problem.

An alternative that I've experimented with is for the platform to be the _only_ part of the system that imports React, and then it can wrap it for export so that consumers always use the same lib that the platform is using. That would mean using `import React from 'meetup-web-platform/lib/react';`, which is a little weird, but would enforce consistency.

src/components/Link.js
mmcgahan: I can roll these exports back in favor of exporting the whole lib at once in `src/router.js`, but the underlying issue is that redux-little-router creates a 'cache' to store matched routes when it is first imported, and the `<Fragment>` and `<Link>` components in the consumer therefore need to use the _same_ import as the platform. If the consumer imports its own `Link` and `RouterProvider` directly from `redux-little-router` itself, it will get a new cache independent of the platform cache, and things won't work.

It's not a system I particularly like, and it makes me more skeptical of redux-little-router

src/epics/sync.js
mmcgahan: no - redux-little-router provides a way of manually triggering its `LOCATION_CHANGED` event with the initial location on the server, which is effectively what `@server/RENDER` was being used to signal.

src/epics/sync.js
mmcgahan: `get` is used to indicate that it is a function that returns the epic function (which receives the `action$` observable as an argument). The others are already epic functions, so they don't need to get called before being used in `combineEpics`.

src/util/createStore.js
mmcgahan: A `store` that has been created with [store enhancers](https://github.com/reactjs/redux/blob/master/docs/Glossary.md#store-enhancer)

mmcgahan:
The more I work with this lib, the less confident I am that it's a good fit for us at this point. The integration with Redux is great, and I really like the interface it provides for adding metadata to routes, but it really doesn't have a good solution for overlapping route patterns and 'index routes', which are critical for us. Testing is also a major pain - any component with a `<Link>` tag needs a complete Redux store with the router enhancer applied using a valid/mocked `location`, which is a pain.

package.json
sadafie: Ah okay, that totally makes sense. Peer dependencies seem like the more correct way to do it. Thanks for the explanation 👍

src/components/Link.js
sadafie: Ohhhhhhhhhhh. Yeah, I'm a bit skeptical as well if that's what is happening. If we do go this route, I'd suggest just dropping this as a comment in these components.

src/epics/sync.js
sadafie: gotchya

sadafie:
>The more I work with this lib, the less confident I am that it's a good fit for us at this point.

Based on some of the explanations you've given here, I think I agree. Also, there are so many [routing solutions out there](https://github.com/markerikson/redux-ecosystem-links/blob/master/routing.md) that I think we can try out a few others if we are so inclined. 

Have we already deemed that React-Router 4 is not a good fit for us either?

mmcgahan:
I've re-opened the investigation ticket for React Router - https://meetup.atlassian.net/browse/MW-358

RRv4 doesn't do anything with Redux, and doesn't provide an obvious way to attach query functions to routes, so we might need to figure out a different solution for that, but I think that might not be a bad thing long term. Many other features of RRv4 look really well-thought-out, though, so definitely worth a look.

PR #28, opened 2016-10-23
added logging of api request time

mattkime:
unsure of we have a format for this sort of thing yet


sadafie:
This looks good to me, but will leave it for @mmcgahan to take a look too.

mmcgahan:
LGTM - we'll probably override `console.log` and/or `request.log` when we determine standard formatting, so this will work. You _could_ add another 'tag'` to the `request.log` call to make it slightly easier to filter for timings, but since we don't really have a full logging or analytics story, not really necessary

PR #29, opened 2016-10-24
WP 45 transpile to CommonJS on publish

mmcgahan:
In order to avoid awkward Webpack configuration requirements for platform consumers, this PR updates the package configuration with a `build` script that transpiles everything from `src/` into `lib/` using Babel.

This command is run automatically during `npm publish` so that the published package uploaded to the npm registry will contain the CommonJS/ES5-compatible code. Git will ignore the built code.

Instead of importing with `meetup-web-platform/<moduleName>`, consumers can import individual es5 modules with `meetup-web-platform/lib/<moduleName>`, individual es6 modules with `meetup-web-platform/src/<moduleName>`, or _some_ core modules directly from the root component with `meetup-web-platform`, e.g.

```js
import { makeServerRenderer } from 'meetup-web-platform';
```


.gitignore
```diff
@@ -35,3 +35,6 @@ jspm_packages
 
 # Optional REPL history
 .node_repl_history
+
+lib
```
mattkime: would be nice to find an automated way to check in a built version

mmcgahan: if this line was taken out, Travis would automatically check in the build version when it ran `npm publish`, but I would prefer never to check in built files for various 'best practice' reasons.

mattkime: I'm confused about the goal. i thought the goal was to make this easier to use in a js version agnostic way. i'm unclear about the benefit of having users run the module's build task.

or maybe i'm conflating git check in with npm check in.

mmcgahan: ah, I'll make the description more clear - the user doesn't run the build task here, it's just defined so that `prepublish` can run it automatically during `npm publish` so that the version that is uploaded to `npm` _does_ have the built version in it. Git will keep track of the source code while npm will contain the built CommonJS-compatible code, similar to how React and some other projects are set up.

mattkime: aaah, thanks for explaining💡

mattkime:
lgtm. leaving open so others can review

PR #34, opened 2016-10-28
WP-54 Actually make API requests in parallel

mmcgahan:
The current code uses `concatMap` to make sure that API requests are returned in the order of their corresponding queries from the app. However, `concatMap` keeps order by making the requests serially rather than in parallel, which is not what we want. Although RxJava has `concatMapEager` that has the behavior we want, RxJS does not, so we have to use `flatMap` to make the calls in parallel and then sort the responses after they all return.

This should significantly improve time to first byte, particularly for deeply-nested routes. I'll try to get numbers from mup-web

for [this event url](http://localhost:8000/nyc-kanban/events/fkswhlyvpbxb):

- Previous TTFB: 2.4s (same for client-side API request)
- This branch TTFB: **0.8s**

### DIFF
--- a/src/apiProxy/api-proxy.js
+++ b/src/apiProxy/api-proxy.js
@@ -122,7 +122,7 @@ export const buildRequestArgs = externalRequestOpts => ({ endpoint, params }) =>
  *
  * @param {Object} apiResponse JSON-parsed api response data
  */
-export const apiResponseToQueryResponse = ([response, query]) => ({
+export const apiResponseToQueryResponse = query => response => ({
 	[query.ref]: {
 		type: query.type,
 		value: response,
@@ -219,6 +219,25 @@ export const apiResponseDuotoneSetter = duotoneUrls => {
 	};
 };
 
+export const makeApiRequest = (request, API_TIMEOUT, duotoneUrls) => {
+	const setApiResponseDuotones = apiResponseDuotoneSetter(duotoneUrls);
+
+	return ([requestOpts, query]) =>
+		externalRequest$(requestOpts)
+			.timeout(API_TIMEOUT, new Error('API response timeout'))
+			.do(([response]) => request.log(['api'], `${response.elapsedTime}ms - ${response.request.uri.path}`)) // log api response time
+			.map(([response, body]) => body)    // ignore Response object, just process body string
+			.map(parseApiResponse)             // parse into plain object
+			.catch(error =>
+				Rx.Observable.of({ error: error.message })
+			)
+			.map(apiResponseToQueryResponse(query))    // convert apiResponse to app-ready queryResponse
+			.map(setApiResponseDuotones);        // special duotone prop
+};
+
+const sortResponsesByQueryOrder = queries => responses =>
+	queries.map(({ ref }) => responses.find(response => response[ref]));
+
 /**
  * This function transforms a single request to the application server into a
  * parallel array of requests to the API server, and then re-assembles the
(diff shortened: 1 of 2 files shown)


src/apiProxy/api-proxy.js
```diff
@@ -249,27 +267,11 @@ const apiProxy$ = ({ API_TIMEOUT=5000, baseUrl, duotoneUrls }) => {
...
-			.catch(error =>  // if there has been an error, all 'responses' need to be the error
-				Rx.Observable.of({ error: error.message })
-					.repeat(queries.length)
-			)
+			.do(({ url }) => request.log(['api'], JSON.stringify(url)))  // logging
 			.zip(Rx.Observable.from(queries))   // zip the apiResponse with corresponding query
-			.map(apiResponseToQueryResponse)    // convert apiResponse to app-ready queryResponse
-			.map(setApiResponseDuotones)        // special duotone prop
-			.toArray();                         // group all responses into a single array - fan-in
+			.flatMap(makeApiRequest(request, API_TIMEOUT, duotoneUrls))  // parallel requests
```
mattkime: I'm mostly sure that you could use `concatMap` instead of manually sorting. 

http://fernandocejas.com/2015/01/11/rxjava-observable-tranformation-concatmap-vs-flatmap/ (scroll down to the appropriate marble diagram)

mattkime:
woohoo! in general, the code look good. just want to see if mergemap is the operator you were looking for.

mmcgahan:
unfortunately mergeMap has the same ordering behavior of flatMap - it won't guarantee the output appears in the same order as the input - if the first query takes a long time and the second one is fast, the response to the second query will be emitted by both flatMap and mergeMap before the response to the first query. I think mergeMap is essentially a higher-order flatMap, where instead of mapping onto Observables, you map onto a function that returns observables.

src/apiProxy/api-proxy.js
mmcgahan: yeah i saw that line and initially thought it might be what i wanted, but basically mergeMap with higher concurrency results in output where you can't guarantee the order of whatever is running concurrently

mmcgahan:
Actually I'm wrong about the relationship between mergemap and flatmap - `flatMap` is just an alias for `mergeMap` :|

src/apiProxy/api-proxy.js
mmcgahan: in RxJS, yeah - but in RxJava `concatMapEager` does both... go figure. Not sure why `concatMapEager` hasn't been implemented for RxJS yet. Might be a fun side project to create it and submit as a PR to RxJS at some point.

mattkime: Ug, I want to avoid the game of guessing at the correct RxJs usage.

This is good but I think we should circle back to this / create a new ticket to solve this problem.

mattkime: @mmcgahan completely agree.

i'm going to create an investigate task. its possible that this is an rxjs bug or unexpected behavior of some sort.

mattkime: i'm happy to merge whenever

mmcgahan: `concatMap` is working as designed - it buffers its input so that its handler function is only called when the previous value has been processed (the handler is evaluated lazily), whereas `concatMapEager` evaluates greedily and buffers its output until previous calls resolve (which could be _very_ computationally expensive with a talkative input). `concatMap` is just the conservative option that would be preferred in some cases.

mmcgahan: yeah, `.zip` is a lot simpler:

https://github.com/meetup/meetup-web-platform/compare/WP-54_parallel-api...parall-zip-simple

PR #35, opened 2016-10-29
WP 39 allow queries to inject 'mockResponse'

mmcgahan:
This update allows consumer apps to mock the expected response from the API _before_ the expected API endpoint is available.

Example:

A new feature will use a new API `/:urlname/candy` endpoint - it returns a new `type` of data called `candy` with an `id` and `flavor` property.

The query function would look like this:

``` js
function candyQuery({ params, location }) {
  const urlname = params.urlname;
  return {
    type: 'candy',
    params: { urlname },
    ref: 'candystate',
    mockResponse: {
      id: 1234,
      flavor: 'kiwi'
  };
}
```

the `apiConfigCreator` would need to be defined like this:

``` js
function candy(params) {
  return {
    endpoint: `${params.urlname}/candy`,
    params
  };
}
```

Then, even if the API server isn't handling `/:urlname/candy` yet, the app will load the `mockResponse` into Redux state at `state.app.candystate.value`.

When the API endpoint is ready, you just need to remove `mockResponse` from the query function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,51 @@ From the client-side application's point of view, it will always send
 the `queries` and recieve the `queryResponses` for any data request -
 all the API-specific translations happen on the server.
 
+### Faking an API response
+
+Sometimes, during development, you might want to set up your consumer app to
+query data from an API endpoint that is not yet set up. In this case, you can
+add a `mockResponse` property to your `query` object that will be used as the
+return value for the API call (you _have_ defined the API return values,
+right?). When the API endpoint is ready, simply remove the `mockResponse` from
+the query and the platform will call the API as configured in
+`src/apiProxy/apiConfigCreators.js`.
+
+**Example**
+
+A new feature will use a new API `/:urlname/candy` endpoint - it returns a new
+`type` of data called `candy` with an `id` and `flavor` property.
+
+The query function in the consumer app would look like this:
+
+``` js
+function candyQuery({ params, location }) {
+  const urlname = params.urlname;
+  return {
+    type: 'candy',
+    params: { urlname },
+    ref: 'candystate',
+    mockResponse: {
+      id: 1234,
+      flavor: 'kiwi'
+  };
+}
+```
+
+the `apiConfigCreator` would need to be defined like this:
+
+``` js
+function candy(params) {
+  return {
+    endpoint: `${params.urlname}/candy`,
+    params
+  };
+}
+```
+
+Then, even if the API server isn't handling `/:urlname/candy` yet, the app will
+load the `mockResponse` into Redux state at `state.app.candystate.value`.
+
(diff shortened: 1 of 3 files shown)


mattkime:
is it possible for this to be accomplished from the consumer repo? would be nice to avoid creating new revisions of meetup-web-platform

mmcgahan:
There are ways to hack together something similar in a consumer repo using custom actions and modifying the `app` reducer, but it would require much more setup and much more teardown/rewriting when the API endpoint was actually ready for real, which makes dev slower to write and review, and more bug-prone. Without support for a feature like this, consumer apps don't have a 'standard' way of writing their app code before an API endpoint is available, which could lead to confusion and awkward hacks.

One nice thing is that this is a non-breaking change, so consumers don't need to upgrade unless they want to use it. I just realized I need to add docs, though.

> would be nice to avoid creating new revisions of meetup-web-platform

In what way? Every Jira ticket will create a new revision, but I'm not sure what the cost is at the moment? Once we have actual in-production apps we'll need to try to slow down/group our breaking changes, but at the moment we should probably keep building.

mattkime:
>>In what way?

I'm thinking through the process of someone getting started with meetup-web-platform. Its a bit indirect for the first step of working on a new feature to be adding mock data to a module dependency.

mmcgahan:
Ah okay I think I haven't communicated this clearly - the platform code does _not_ have to change when new consumers want to mock the API response, since the query functions are only defined inside the consumer app code. This PR just makes sure that if that `mockResponse` is sent from the consumer to the platform, the app server won't call the actual API.

sadafie:
@mmcgahan this looks good! Could we also add that README content to mup-web?

mmcgahan:
I'm reluctant to add redundant README info since it will inevitably get out of date somewhere. I guess I can just add a link to the platform README, however.

sadafie:
> I'm reluctant to add redundant README info since it will inevitably get out of date somewhere. I guess I can just add a link to the platform README, however.

Fair, i'm just wondering if you're new to `mup-web` or any consumer app and needed this, you wouldn't necessarily know the functionality/documentation exists in the platform repo.

mmcgahan:
> you wouldn't necessarily know the functionality/documentation exists in the platform repo.

Yeah, I think we'll probably want to do a documentation roundup sometime soon. I'm actually thinking that we will probably want to move a bunch of docs from mup-web into the platform repo and the starter kit, and mainly have links from mup-web back to those 2 "source" repos.

sadafie:
> Yeah, I think we'll probably want to do a documentation roundup sometime soon.

Would be good. Your idea sounds good. Also, maybe it'd be nice to have a master Confluence T.O.C. or something that simply links out to GH or a confluence doc in the cases where we have one.

PR #38, opened 2016-10-31
add support for group communication GET endpoints

akdetrick:
### Description
Add support for mug communication-related queries

### Jira ticket
https://meetup.atlassian.net/secure/RapidBoard.jspa?rapidView=66&view=detail&selectedIssue=WP-56

### See also
In chapstick, `static/script/mu/core/api.js :392`

### DIFF
--- a/src/apiProxy/apiConfigCreators.js
+++ b/src/apiProxy/apiConfigCreators.js
@@ -32,25 +32,22 @@ export function group(params) {
 /**
  * group communication endpoints
  *
- * accepts an `method` property in params
+ * accepts an `apiMethod` property in params:
+ * 	`apiMethod: 'followers' - return followers for a given conversation id
+ * 	`apiMethod: 'suggested_members' - return suggested_members for a given conversation id
  */
 export function mugcomm(params) {
-	const path = 'communications';
-	const { urlname, conversationId } = params;
+	const { urlname, conversationId, apiMethod } = params;
 
-	let endpoint;
-	switch (params.method) {
-		case 'getAll':
-			endpoint = `${urlname}/${path}`;
-		case 'getId':
-			endpoint = `${urlname}/${path}/${conversationId}`;
-		case 'followers':
-			endpoint = `${urlname}/${path}/${conversationId}/followers`;
-		case 'suggestedMembers':
-			endpoint = `${urlname}/${path}/${conversationId}/suggested_members`;
-		default:
-			endpoint = `${urlname}/${path}/`;
-	}
+	const endpoint = [
+		urlname,
+		'conversations',
+		conversationId,
+		apiMethod
+	]
+		.filter(urlFragment => urlFragment) // only inlcude populated fragments
+		.join('/');
+
 	return {
 		endpoint,
 		params
--- a/src/apiProxy/apiConfigCreators.js
+++ b/src/apiProxy/apiConfigCreators.js
@@ -30,6 +30,34 @@ export function group(params) {
 }
 
 /**
+ * group communication endpoints
+ *
+ * accepts an `method` property in params
+ */
+export function mugcomm(params) {
+	const path = 'communications';
+	const { urlname, conversationId } = params;
+
+	let endpoint;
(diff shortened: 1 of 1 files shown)


src/apiProxy/apiConfigCreators.js
```diff
@@ -30,6 +30,34 @@ export function group(params) {
 }
 
 /**
+ * group communication endpoints
+ *
+ * accepts an `method` property in params
+ */
+export function mugcomm(params) {
```
mmcgahan: not sure if `mugcomm` is a term you're using internally, but most new code refers to mugs as `group`s and there's not really much advantage to abbreviating 'comm' here - I'd recommend `groupCommunications`, which maps semantically to the `/:urlname/communications/` endpoint paths

src/apiProxy/apiConfigCreators.js
```diff
@@ -30,6 +30,34 @@ export function group(params) {
...
+	const path = 'communications';
+	const { urlname, conversationId } = params;
```
mmcgahan: Since all the endpoints start with `${urlname}/${path}`, and the remaining path components are essentially pre-defined in the `params`, you can be DRYer here by writing the endpoint more directly and only having `method` values of `followers` and `suggested_members`:

```js
const { urlname, conversationId, method } = params;

const endpoint = [
  urlname,
  'conversations',
  conversationId,
  method
]
.filter(c => c)  // only include non-empty values
.join('/');  // create a url path string

return {
  endpoint,
  params,
};
```

src/apiProxy/apiConfigCreators.js
```diff
@@ -30,6 +30,34 @@ export function group(params) {
...
+
+	let endpoint;
+	switch (params.method) {
```
mmcgahan: `method` already has a distinct meaning when making HTTP requests - probably better to find a different var name in this context. The underlying HTTP method will be forwarded from the initial request from the client.

src/apiProxy/apiConfigCreators.js
```diff
@@ -30,6 +30,34 @@ export function group(params) {
...
+		case 'getAll':
+			endpoint = `${urlname}/${path}`;
```
mmcgahan: when using `switch`, don't forget to put on the `break`s. This `switch` will currently _always_ return ``endpoint = `${urlname}/${path}/`;``

src/apiProxy/apiConfigCreators.js
akdetrick: Loving all the functional styles I see in the new platform; this is an elegant way to handle creating a path based on which values are actually populated from `params`.

One concern I have is that we would lose the self-documenting aspect of the `method` values in the `switch` statement. The group communications v0 endpoint for example, is not documented in the public API yet. I suspect we'll run into this with a number of new features...

I could always list accepted values in a comment above, but compared to actual code, a comment is sometimes a lie waiting to happen.

mmcgahan: Yeah that's a valid concern for sure - I think one thing that would help is just renaming `method` to something like `convoInfo` (not a great name, but...). REST principles already give you a lot of the self-documenting advantages:

Without looking at any variable names, it's pretty clear what each endpoint should return

- `/:urlname/conversations` - all conversations in a group
- `/:urlname/conversations/:conversationId` - a single conversation in a group
- `/:urlname/conversations/:conversationId/{followers|suggested_members}` - pretty clear

IMO all that stuff is already self-documented, although I would loooove to be able to use something like typescript/flowjs to declare the supported values for `convoInfo`. In this case, that's the only 'ambiguous' variable, and I think it would be fine to just leave a code comment about what you're expecting, which also gives you a little bit of wiggle room to "automatically" support new types of `convoInfo` endpoints.

mmcgahan:
Now that this is merged, Travis will publish a new version of the package that will need to be updated in mup-web. You should be able to find out what the version number is when it's ready at npmjs.com/package/meetup-web-platform

PR #39, opened 2016-11-03
New testUtil: routeRenderer

mmcgahan:
Testing application routes requires a fair amount of boilerplate to set up, and all of the dependencies are already part of the platform, so it makes sense to provide a simpler interface for rendering a route in an app's unit tests.

This PR adds `util/testUtils.js:routeRenderer` - a function that takes an app's `routes` and returns a function that renders a particular location (url) and returns a promise that resolves with the rendered React app. This function can then be called repeatedly with different locations across multiple unit tests. I'll set up a PR for mup-web that demonstrates how much it simplifies the route testing.

### DIFF
--- a/src/util/testUtils.js
+++ b/src/util/testUtils.js
@@ -1,8 +1,13 @@
+import React from 'react';
 import { Observable } from 'rxjs';
 import { ActionsObservable } from 'redux-observable';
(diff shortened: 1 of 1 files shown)


mattkime:
lgtm

PR #40, opened 2016-11-03
WP-49 - request/parse feature flags from the API

mmcgahan:
The REST API provides a way of reading arbitrary feature flag values using headers in the API requests.

This PR adds a feature to the `query` parsing in the API proxy that will read an array of feature flag names and return a map of `{ [flagName]: [Boolean value] }` in the response so that the flags can be stored in Redux and consumed by the app's UI components

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,24 @@ server routes and plugins. See the code comments for usage details.
 
 ## API Adapter
 
+### Query
+
+The app server expects API requests to be provided in the form of `query`
+objects in the payload of a single request to the app server at the `/api`
+endpoint. A query takes the following shape:
+
+```js
+{
+	type: <string>,
+	params: {
+		<string>: <string>,
+		...
+	},
+	ref: <string>,
+	flags: [<string>, ...],
+}
+```
+
 When the application server receives the JSON-encoded array of queries,
 it uses an API adapter module to translate those into the configuration
 needed to fetch data from an external API. The application server is
@@ -96,6 +114,15 @@ This adapter is used to proxy all requests to `/api`.
 From the client-side application's point of view, it will always send
 the `queries` and recieve the `queryResponses` for any data request -
 all the API-specific translations happen on the server.
+
+### Feature flags
+
+The API adapter provides an interface into feature flag values - just pass
+an array of feature flag names in the `flags` array of your query, and the
+response from the API adapter will return an object mapping the flag names to
+their true/false valuse in the `flags` property of the response. Your
+application will probably add these flag values directly to the Redux `state`,
+where they can be consumed by components to affect the UI.
 
 ## Middleware/Epics
 
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ endpoint. A query takes the following shape:
 		...
 	},
 	ref: <string>,
-	flags: [<string>, ...],
+	flags?: [<string>, ...],
 }
 ```
 
@@ -114,6 +114,21 @@ This adapter is used to proxy all requests to `/api`.
 From the client-side application's point of view, it will always send
 the `queries` and recieve the `queryResponses` for any data request -
 all the API-specific translations happen on the server.
+
+A `queryResponse` takes the following shape:
+
+```js
+{
+	<string (ref)>: {
+		value?: <parsed API response JSON>
+		flags?: {
+			<string>: <Boolean>
+			...
+		},
+		error?: <string>
+	}
+}
+```
 
 ### Feature flags
 
--- a/src/apiProxy/api-proxy.js
+++ b/src/apiProxy/api-proxy.js
@@ -41,9 +41,9 @@ export const parseApiResponse = ([response, body]) => {
 	let value;
 	const flags = parseResponseFlags(response);
 
-	if ((response.status || response.statusCode) > 299) {
+	if (response.statusCode > 299) {
 		return {
-			value: { error: response.statusText },
+			value: { error: response.statusMessage },
 			flags,
 		};
 	}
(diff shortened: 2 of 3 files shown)


src/apiProxy/api-proxy.js
```diff
@@ -92,40 +113,48 @@ function urlFormatParams(params, doEncode) {
...
-		externalRequestOptsQuery.headers['content-type'] = 'application/x-www-form-urlencoded';
-		break;
-	}
+export const buildRequestArgs = externalRequestOpts =>
+	({ endpoint, params, flags }) => {
 
-	return externalRequestOptsQuery;
-};
+		// cheap, brute-force object clone, acceptable for serializable object
+		const opts = JSON.parse(JSON.stringify(externalRequestOpts));
```
mmcgahan: this was changed to do a 'deep clone' in a cheap and simple way. Previous code made a shallow clone that was getting previously-modified `headers`. Ah, immutability...

src/apiProxy/api-proxy.js
```diff
@@ -92,40 +113,48 @@ function urlFormatParams(params, doEncode) {
...
+		opts.url = `/${endpoint}`;
+
+		const formattedParams = urlFormatParams(params, opts.method === 'get');
+
+		if (flags) {
+			opts.headers['X-Meetup-Request-Flags'] = flags.join(',');
+		}
```
mmcgahan: this is the only 'new' block of code in this function - the rest is mostly an indentation change

src/apiProxy/api-proxy.js
```diff
@@ -219,11 +248,16 @@ export const apiResponseDuotoneSetter = duotoneUrls => {
 	};
 };
 
+const MOCK_RESPONSE_OK = {
```
mmcgahan: this is _not_ a testing mock, but a mock `[http.IncomingMessage](https://nodejs.org/api/http.html#http_class_http_incomingmessage)` object to simulate the results of a real `externalRequest$`

README.md
```diff
@@ -97,6 +115,30 @@ From the client-side application's point of view, it will always send
...
+			...
+		},
+		error?: <string>
+	}
+}
+```
+
+### Feature flags
+
+The API adapter provides an interface into feature flag values - just pass
```
sadafie: 👍

src/apiProxy/api-proxy.js
```diff
@@ -5,6 +5,16 @@ const externalRequest$ = Rx.Observable.bindNodeCallback(externalRequest);
 import * as apiConfigCreators from './apiConfigCreators';
 import { duotoneRef } from '../util/duotone';
 
+const parseResponseFlags = ({ headers }) =>
+	(headers['x-meetup-flags'] || '')
+		.split(',')
+		.filter(pair => pair)
```
sadafie: what is this filter for? do you need it?

src/apiProxy/api-proxy.js
```diff
@@ -27,20 +37,28 @@ import { duotoneRef } from '../util/duotone';
...
-	let error;
+export const parseApiResponse = ([response, body]) => {
+	let value;
+	const flags = parseResponseFlags(response);
 
-	try {
-		responseObj = JSON.parse(response);
-	} catch(e) {
-		error = `API response was not JSON: "${response}"`;
+	if (response.statusCode > 299) {
```
sadafie: nitpick, but i'd almost just rather check if the statusCode doesn't start with '2' instead, unless you're handling `1xx` informational codes as a success.

src/apiProxy/api-proxy.js
```diff
@@ -27,20 +37,28 @@ import { duotoneRef } from '../util/duotone';
...
-		error = `API problem: ${responseObj.problem}: ${responseObj.details}`;
+	try {
+		value = JSON.parse(body);
+		if (value && value.problem) {
+			value = { error: `API problem: ${value.problem}: ${value.details}` };
+		}
+		return { value, flags };
+	} catch(err) {
+		return {
+			value: { error: err.message },
```
sadafie: Another nitpick, might be a bit DRY-er to do `return formatErrorResponse(...)` for the three places that we're returning errors in this function. i.e., in case we ever want to update what the error response object looks like

src/apiProxy/api-proxy.test.js
```diff
@@ -42,18 +42,39 @@ describe('parseRequest', () => {
 });
 
 describe('parseApiResponse', () => {
+	const MOCK_RESPONSE = {
```
sadafie: move to mocks.js?

src/apiProxy/api-proxy.js
mmcgahan: Yeah, since an empty string will yield `['']` when using `split` and if there was a dangling comma you would get an empty string from `split`, this just 'validates' the input to only pass through populated values. I can leave a comment so that maintainers don't accidentally remove it

src/apiProxy/api-proxy.js
mmcgahan: I think it's more accurate to say I'm handling `1xx` as 'not errors', but yeah that's not really an edge case I'm explicitly handling. However, I don't think it's actually possible to get a `1xx` code from an ajax call? I'm curious if you know if it's possible.

src/apiProxy/api-proxy.js
mmcgahan: works for me. TODO

src/apiProxy/api-proxy.test.js
mmcgahan: I have a mild preference to keep this mock in this test file because it's very specific to the response generated by the npm `request` library - not a Hapi `response` or a fetch `response` or an `apiResponse`, which are used elsewhere.

src/apiProxy/api-proxy.js
mmcgahan: Updated to return an error for `< 200` as well

mmcgahan:
All set for final review

PR #42, opened 2016-11-09
WP-23 More unit tests

mmcgahan:
**`Blocked` until #57 merges**

**`Prelim` while I flesh out the tests - just want to get Travis running on it.**


src/util/serverUtils.js
```diff
@@ -0,0 +1,72 @@
+import https from 'https';
```
mmcgahan: moved these functions from `src/server.js` to this utils module to make it easier to mock them when testing the `server.js` code

sadafie: this is such a nitpick that's not related to this PR, but i think we talked about this once before, so I wanted to bring this up. Should we rename this directory from `util` to `utils` for consistency?

sadafie:
@mmcgahan these changed look good, though the branch has conflicts. is this still `Prelim`?

mmcgahan:
@sadafie yeah there are quite a few more tests I want to add, but a fair amount of the missing coverage is related to other tickets in my current sprint, so I wanted to build from that work in this branch - probably will stay Prelim until near the end of the sprint.

src/util/serverUtils.js
mmcgahan: yes - that change has been made in the starter repo but hasn't been carried over yet

mmcgahan:
There's a lot more testing I want to add, but I'm going to merge this for now so that it un-blocks some other stuff

More to come

PR #43, opened 2016-11-10
WP 63 Separate refresh_token from 'anonymous' token logic

mmcgahan:
**Breaking change: removes `anonymous` cookie and `auth.anonymous` Redux state**

The name of `anonAuthPlugin` is no longer accurate, along with a number of other references to 'anonymous auth' in the codebase, because many parts of the anonymous auth flow are also used for refreshing the oauth token for logged-in members as well - the `https://secure.dev.meetup.com/oauth2/authorize` and `https://secure.dev.meetup.com/oauth2/access` are not restricted to generating tokens for anonymous members, even thought that's what we were using them for originally.

The naming and shared responsibility of the code was producing an error, where the platform would set an `anonymous=true` cookie even when _refreshing_ a logged-in member's oauth token, which would put the app in a weird state.

This PR changes names and removes the `anonymous` cookie from all auth flows. Instead, applications should check the `self` property of the application state to determine whether the member is logged in or not (e.g. `const isLoggedOut = state.self.status === 'prereg'`).


src/util/config.js
```diff
@@ -12,14 +12,24 @@ export default function getConfig(overrideConfig) {
 	/**
 	 * Read all config from environment variables here once on startup
 	 */
+	if (!process.env.OAUTH_ACCESS_URL && process.env.ANONYMOUS_ACCESS_URL) {
```
sadafie: +1 for these messages!

... should we add a TODO to remove at some point?

mmcgahan:
Ideally we can have some sort of automated process to announce releases of breaking versions, but to some extent each app will have to manage its dependency versions like they do any npm package - if they need a new feature that's in a breaking version of the package, they'll need to refactor their app to get up to date.

This requirement will be a little annoying at times, e.g. when an app wants to add a new `apiConfigCreator` to support a new API endpoint, but they'll need to update unrelated app code in order to be able to update to the platform version that supports the endpoint, so I'd like to talk about options.

It may very well be that we split off the `apiConfigCreator` section into its own dependency since it will likely change more quickly than other platform code and is less likely to include truly breaking changes.

sadafie:
>This requirement will be a little annoying at times, e.g. when an app wants to add a new apiConfigCreator to support a new API endpoint, but they'll need to update unrelated app code in order to be able to update to the platform version that supports the endpoint, so I'd like to talk about options.

Let's chat about it at today's meeting. When I was working with @atrout she had some apiConfigCreator questions/concerns/suggestions, and I'd like to throw those out there as well.

PR #44, opened 2016-11-11
WP-72 Fix UTF-8 encoded characters when passing URLs through to the api.

willh-meetup:
This fixes the problem with the API not recognizing group URLs when calling through the api. We were encode the params, but not the URL itself.

### DIFF
--- a/src/apiProxy/api-proxy.js
+++ b/src/apiProxy/api-proxy.js
@@ -99,7 +99,7 @@ function urlFormatParams(params, doEncode) {
  */
 export const buildRequestArgs = externalRequestOpts => ({ endpoint, params }) => {
 	const externalRequestOptsQuery = { ...externalRequestOpts };
-	externalRequestOptsQuery.url = `/${endpoint}`;
+	externalRequestOptsQuery.url = encodeURI(`/${endpoint}`);
 
 	const dataParams = urlFormatParams(params, externalRequestOptsQuery.method === 'get');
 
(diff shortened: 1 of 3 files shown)


src/apiProxy/api-proxy.test.js
```diff
@@ -94,6 +95,17 @@ describe('buildRequestArgs', () => {
...
+	});
+
+	const testQueryResults_utf8 = mockQuery(MOCK_RENDERPROPS_UTF8);
+	const apiConfig_utf8 = queryToApiConfig(testQueryResults_utf8);
+	const encoded_url = '/%E3%83%90-%E4%BA%AC';
+
+	it('Properly encodes the URL', () => {
+		const method = 'get';
+		const getArgs = buildRequestArgs({ ...options, method })(apiConfig_utf8);
+		expect(getArgs.url.split('?')[0]).toEqual(encoded_url);
```
willh-meetup: not 100% sure how I feel about this test. 

I think it's a good idea to make sure the URL is fully encoded, but maybe there's a better way to do this.

src/apiProxy/api-proxy.js
```diff
@@ -99,7 +99,7 @@ function urlFormatParams(params, doEncode) {
  */
 export const buildRequestArgs = externalRequestOpts => ({ endpoint, params }) => {
 	const externalRequestOptsQuery = { ...externalRequestOpts };
-	externalRequestOptsQuery.url = `/${endpoint}`;
+	externalRequestOptsQuery.url = encodeURI(`/${endpoint}`);
```
willh-meetup: This seemed like the best place to make sure this is always encoded, and I can't think of a reason we wouldn't want to do this before sending along to the API.

src/apiProxy/api-proxy.test.js
mmcgahan: I think a slightly more general test would parse the URL for the `pathname` and then test that against an [isASCII regex](http://stackoverflow.com/a/14313213/579685)

```js
const { pathname } = require('url').parse(getArgs.url);
expect(/^[\x00-\xFF]*$/.test(pathname)).toBe(true);
```

willh-meetup: 👍

PR #45, opened 2016-11-11
WP 74 refactor auth plugin to use server.auth

mmcgahan:
See https://meetup.atlassian.net/browse/WP-74 for details - essentially this improves security for all routes by containing all the auth behavior in the auth plugin and performing authentication before the request handler is invoked, during Hapi's designated 'auth' part of the [request lifecycle](http://hapijs.com/api#request-lifecycle)

Note: all routes will be authenticated by default. Routes can opt out of authentication with `config: { auth: false }`, which is a documented Hapi route feature.


sadafie:
#43 has merged, so removing prelim.

src/plugins/requestAuthPlugin.js
```diff
@@ -26,20 +21,34 @@ function verifyAuth([request, auth]) {
...
+			value: auth.refresh_token,
+			opts: { path, ttl: YEAR_IN_MS * 2 },
+		}
+	};
+	Object.keys(authState).forEach(name => {
+		const cookieVal = authState[name];
+		// apply to request
+		request.state[name] = cookieVal.value;
+		// apply to response - note this special `reply` prop assigned onPreAuth
+		request.authorize.reply.state(name, cookieVal.value, cookieVal.opts);
```
sadafie: Clarification question -- Authentication is happening before the route is called, so regardless of whether that succeeds or fails, the router is still rendered and there's an attempt at making the associated queries still, yes? That part hasn't changed?

mmcgahan: This is bring up an interesting issue. In the current setup, we are assuming that authentication _never_ fails, because if the incoming request doesn't have any auth tokens, then new (anonymous) tokens are provided - that's the main function of the plugin currently.  

You are correct that the route handler will then attempt to render the app, but since there will be no API auth information in the request, that will fail with a `No auth info provided` Red Screen of Death.

It might actually be helpful to have a hard "Server Error" kind of response if/when auth fails _before_ attempting to render the app, just to contain the process a little more - I'll check out where that would go.

A related idea that I had while working on this is that we probably want to build some sort of 'retry' logic into this flow in case there's a temporary network problem when requesting auth tokens... it's pretty straightforward to implement that with RxJS, but it might be good to have some more discussion about the whole auth flow.

mmcgahan:
> it'd be good to do one of the group "code-review" of the auth code and request plugin to make sure we all understand the flows?

Sounds good. I've learned a lot about the Hapi auth tools that we hadn't been using at all, so it would be good to share that info.

src/plugins/requestAuthPlugin.js
sadafie: Ah, awesome that makes sense. thanks for clarifying! i was forgetting how authentication / anonymous auth works.

PR #46, opened 2016-11-15
WP 62 pure cookie auth

mmcgahan:
Consolidating more auth behavior to the server, making it entirely cookie based rather than having a hybrid Redux/cookie system. Adjusting the React/Redux app to extract most of the auth behavior makes it a lot simpler for applications.

One of the main implications for consumers is that "Logout" is no longer a standalone redux action, it's just a navigation to a page with `?logout` in the querystring. This makes login management much simpler, but I think it keeps all necessary functionality intact.

- **before**:
    - LOGOUT_REQUEST
    - clear cookies locally
    - make a request for a new anonymous oauth token
    - refresh page with LOCATION_SYNC
- **after**:
   - navigate anywhere with `?logout`


sadafie:
> One of the main implications for consumers is that "Logout" is no longer a standalone redux action, it's just a navigation to a page with ?logout in the querystring. This makes login management much simpler, but I think it keeps all necessary functionality intact.

Hmmm... what's to stop someone from sending a link to `meetup.com/?logout` that a person accidentally clicks and logs them out? i wonder if this needs signing some how.

src/apiProxy/api-proxy.js
```diff
@@ -304,6 +311,9 @@ export const makeApiRequest$ = (request, API_TIMEOUT, duotoneUrls) => {
 const apiProxy$ = ({ API_TIMEOUT=5000, baseUrl='', duotoneUrls={} }) => {
 
 	return request => {
+		if (!request.state.oauth_token) {
```
sadafie: Does the user get a red error screen in this case or what happens?

mmcgahan: Actually I think I can remove this block since it's taken care of by the auth plugin in the `verifyAuth` function.

BUT, yes, the user will get the red error screen (in dev) from the `onPreResponse` handler applied to the server [here](https://github.com/meetup/meetup-web-platform/blob/WP-62_cookie-auth/src/server.js#L58).

Essentially this is an operational error so the request needs to crash.

mmcgahan:
> i wonder if this needs signing some how.

Interesting thought. There's definitely nothing preventing someone from 'accidentally' signing out when clicking on a link containing the query param.

I'm not exactly sure how to sign this link in a way that doesn't make the consumer app more complicated, but maybe we need a special 'LogoutLink' component that automatically sends a hash of the user's session token so that the server can verify the logout link applies to the correct user.

Do you think this is a blocker or should I open up a follow-on?

sadafie:
>Do you think this is a blocker or should I open up a follow-on?

Have @bgruber or @LenaAnneKrug seen this? I'd like to get their eyes on this approach (from a security standpoint) before doing a follow on. I don't think we necessarily shouldn't be doing this, but it's also an approach we haven't used in the past.

LenaAnneKrug:
Humbly, I don't think the risk of someone sending a ?logout link to someone has enough risk as a significant attack vector or in terms of appetite.

Basically I am saying if resolving it adds significant complexity I am comfortable with not doing it. A more complex solution = higher risk.

But I may be missing something :)

mmcgahan:
That makes sense - my impression is that there's not so much a security risk as a trolling/annoyance risk with using a query param to sign people out, but the more I think about it, the more it makes sense to me to write a simple `<LogoutLink>` component that can contain more 'sophisticated' behavior (e.g. we probably need to _append_ the `logout=true` query param rather than making it the _only_ query param for the navigation link)

mattkime:
lgtm

mmcgahan:
`<LogoutLink>` followup: https://meetup.atlassian.net/browse/WP-82

PR #47, opened 2016-11-16
WP 47 CSRF plugin

mmcgahan:
This update adds CSRF protection to all POSTs. The server sends a CSRF cookie and a header on each app/API request, and the application is responsible for storing the _header_ in app state.

The platform will then read the CSRF from app state and send it back to the server on every POST call through `fetchQueries` - the server confirms that this CSRF matches the Cookie CSRF and the POST is authenticated.

This is a breaking change because it relies on the application to store the CSRF in `config.csrf` whenever it is returned as part of the `API_SUCCESS` action.

In the near future, [MW-406](https://meetup.atlassian.net/browse/MW-406) will allow this kind of change to be non-breaking, because the appropriate reducer will be supplied by the platform directly.


src/plugins.js
```diff
@@ -7,6 +8,16 @@ import requestAuthPlugin from './plugins/requestAuthPlugin';
  * @module ServerPlugins
  */
 
+export function getCsrfPlugin() {
+	return {
+		register: CsrfPlugin.register,
+		options: {
+			secret: 'shhhhh',
```
mmcgahan: TODO: read this from env variable

mattkime:
lgtm

mmcgahan:
Prelim until i update the `secret` to come from an env variable

mmcgahan:
@mattkime I'd like to get some help creating the new secret env variable in GCP - I've made the updates that I think need to happen in the `infrastructure/` directory, but don't remember the `kubectl` CLI sequence to actually get it set up in prod.

sadafie:
This looks good to me as well. @mmcgahan do you still have more work to do on this, re: GCP, or do you want to do that as a follow up?

mmcgahan:
@sadafie This branch is good to go - I just finished the GCP changes in the corresponding mup-web  PR https://github.com/meetup/mup-web/pull/363/

PR #52, opened 2016-11-17
WP 83 login sync

mmcgahan:
Currently, the platform depends on the consumer applications to update the application state on `LOGIN_SUCCESS`. In general, however, a successful login (like a successful logout) should _always_ refresh the current application state _completely_ to show the user-targeted data. The most direct way of doing this is through a navigation event, but since logins are POSTs, we need to simulate a navigation event to the current route when the POST succeeds.

This update will allow mup-web to remove the special-case handling of `LOGIN_SUCCESS` and `LOGIN_ERROR` in its app reducer by letting the platform automatically push a `LOCATION_SYNC` action on `LOGIN_SUCCESS`

### DIFF
--- a/src/epics/sync.js
+++ b/src/epics/sync.js
@@ -6,6 +6,7 @@ import {
 	apiSuccess,
 	apiError,
 	apiComplete,
+	locationSync,
 } from '../actions/syncActionCreators';
 import { activeRouteQueries$ } from '../util/routeUtils';
 import { fetchQueries } from '../util/fetchUtils';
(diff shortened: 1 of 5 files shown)


src/epics/sync.js
```diff
@@ -29,6 +29,18 @@ export const getNavEpic = routes => {
 };
 
 /**
+ * Any action that should reload the API data should be handled here, e.g.
+ * LOGIN_SUCCESS, which should force the app to reload in an 'authorized'
+ * state
+ */
+export const locationSyncEpic = (action$, store) =>
```
sadafie: 👍

PR #53, opened 2016-11-17
WP-84 Apply oauth from login API call, clear auth on logout

mmcgahan:
In order to set oauth cookies with values received from a login API call, we need to hijack the api-proxy response to set response cookies.

In order to reduce the auth info leaked to the browser, this PR also removes the oauth info from the login response before it is returned.

:hooray: security


mmcgahan:
I just realized this should have some tests.

mmcgahan:
... and we need a complementary feature for logout, which will also need a special query

PR #54, opened 2016-11-18
Platform reducer (WP part of MW 406)

mmcgahan:
After #52, there is no longer any need for app-specific code in the current mup-web reducer, which means that it might as well live in the platform.

This update creates a new `src/reducers/platform.js` module that exports a `makeRootReducer` function that allows applications to inject app-specific reducers alongside the full suite of platform-standard reducers (`app`, `config`, `preRenderChecklist`, `routing`).


sadafie:
@mmcgahan this one also needs an update w/ master

PR #57, opened 2016-11-21
WP 67 navigation logs

mmcgahan:
This PR adds tracking/logging data for every navigation action. The tracking data includes the following props in addition to the standard ids/metadata:

```js
{
  url,  // the _from_ pathname (not including `protocol://host:port` info)
  referrer,  // the _to_ pathname
}
```
 
More importantly, it sets up a `meta` property on the `API_REQUEST` action that is an object containing arbitrary request-related data. This data is then sent to the app server in a querystring parameter so that the server can read it and use it for tracking purposes. Currently, it just contains the 'referrer' url, but it can be expanded over time to include more info, e.g. the names of rendered React components, and possible click tracking info

### DIFF
--- a/src/actions/syncActionCreators.js
+++ b/src/actions/syncActionCreators.js
@@ -1,8 +1,8 @@
-export function apiRequest(queries, referrer) {
+export function apiRequest(queries, meta) {
 	return {
 		type: 'API_REQUEST',
 		payload: queries,
-		meta: referrer,
+		meta,
 	};
 }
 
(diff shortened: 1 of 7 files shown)


mmcgahan:
hmm looks like the logout action returns logged-in data in the first instance, but then returns logged-out data afterward. I think this is mergeable, but I'll open a Defect in WP to fix

mmcgahan:
Yeah, the issue is addressed in https://github.com/meetup/meetup-web-platform/pull/58

mattkime:
lgtm

PR #58, opened 2016-11-22
Clean up/simplify auth

mmcgahan:
While writing up the documentation for the `requestAuthPlugin`, I came across a few elements of the auth flow that could be simplified. These are those changes.


mmcgahan:
hm looks like this branch includes #42 as well

mmcgahan:
there we go. much better

mattkime:
lgtm

src/epics/sync.test.js
```diff
@@ -43,6 +43,7 @@ describe('Sync epic', () => {
 
 		const pathname = '/noQuery';
 		const noMatchLocation = { ...MOCK_RENDERPROPS.location, pathname };
+		console.log(noMatchLocation.query, 'asdfasdf');
```
mattkime: mistaken commit?

mmcgahan: ooh, yup good catch

sadafie:
LGTM

PR #62, opened 2016-11-23
Use meetup-web-mocks testUtils

sadafie:
remove methods that were moved to meetup-web-mocks and update pkg

Fixes https://meetup.atlassian.net/browse/WP-98

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+## [0.9]
+
+- **Removed** - Many of the shared methods in `src/util/testUtils`
+now are in the `meetup-web-mocks` repo
+- **Change** - Use the `meetup-web-mocks` repo for testUtil methods,
+e.g.,
+
+```js
+import {
+  createFakeStore,
+} from 'meetup-web-mocks/lib/testUtils';
+```
+
 ## [0.8]
 
 - **Change** - `API_SUCCESS` now returns a `meta` prop that contains the `csrf`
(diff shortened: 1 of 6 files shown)


sadafie:
I'm pretty sure we would consider this a breaking change since `mup-web` would break if the pkg is updated. 

i'm going to make a corresponding update in the `mup-web` repo.

mmcgahan:
Any reason not to move the rest of `testUtils` over to the mocks repo? Sort of a fine line between what is an isn't included, and the stuff that has been left in the platform repo isn't really platform-specific.

mmcgahan:
Breaking changes need a bump in the minor version number in `Makefile`, and an entry in `CHANGELOG` to indicate what a consumer app needs to do to upgrade

sadafie:
> Any reason not to move the rest of testUtils over to the mocks repo? Sort of a fine line between what is an isn't included, and the stuff that has been left in the platform repo isn't really platform-specific.

The rest seemed app server specific utilities that consumer apps wouldn't/don't use? It would also introduce Hapi, Cookie, rxjs, etc. dependencies. If we think it's better to move the whole thing I can, but if it does make sense to keep them separate, maybe it's just a naming conflict/confusion and I can rename one of the files.

mmcgahan:
That makes sense - let's see how it goes with keeping them separate. Definitely don't want to overload the dependencies for something as 'simple' as a mocks lib

PR #64, opened 2016-11-23
WP 92 Log API request+response data

mmcgahan:
**Blocked by #63 - this branch contains those changes**

This should make it easier to debug unexpected API responses

From the new `docs/api-proxy.md`:

---

### Debugging API requests/responses

The server log will show data about the Meetup API requests it generates in
response to incoming queries, although it will truncate the `body` of the
response to 256 characters, which should be sufficient to see the cause of
API errors:

```
2016-11-23 22:56:41.901, [request,api,info] data: {
  "request": {
    "query": {
      "id": "self"
    },
    "pathname": "/2/member/self",
    "method": "get"
  },
  "response": {
    "elapsedTime": 722,
    "body": "{\"service_status\":{\"status\":\"ok\"},\"org_badge\":false,\"categories\":[{\"id\":242,\"shortname\":\"outdoors-adventure\",\"name\":\"???chapter_meta_cat.outdoors-adventure.name???\",\"sort_name\":\"???chapter_meta_cat.outdoors-adventure.name???\",\"photo\":{\"id\":450131943,\"highr..."
  }
}
```

To see the full body of the response, you might be able to inspect network
traffic using something like Charles proxy.


sadafie:
@mmcgahan is there a dependency between this and #63 ? The diffs look similar so I'm not sure what I should be looking at?

mmcgahan:
> @mmcgahan is there a dependency between this and #63 ?

Yup, sorry. `Prelim` added.

src/apiProxy/api-proxy.js
```diff
@@ -344,13 +368,14 @@ export const makeApiRequest$ = (request, API_TIMEOUT, duotoneUrls) => {
...
 			makeExternalApiRequest(request, API_TIMEOUT);
 
-
-		request.log(['api', 'info'], `REST API request: ${requestOpts.url}`);
-		return request$(requestOpts)
-			.map(parseApiResponse(requestOpts.url))                   // parse into plain object
-			.map(parseLoginAuth(request, query))     // login has oauth secrets - special case
-			.map(apiResponseToQueryResponse(query))  // convert apiResponse to app-ready queryResponse
-			.map(setApiResponseDuotones);            // special duotone prop
+		return Rx.Observable.defer(() => {
```
sadafie: ah gotchya makes sense

PR #65, opened 2016-11-24
WP 97 track logout

mmcgahan:
This adds a `trackLogout` call to `trackApi` when `logout` is in the request querystring.

There's also some related cleanup to the tracking code to remove the unnecessary return values that prevent making multiple tracking calls from `trackApi`, and refactoring the tests to account for the lack of return values.

Also some consolidation of the redundant `request` definitions in the tracking tests to make it more explicit what request conditions are being tested.


mattkime:
lgtm

PR #67, opened 2016-11-28
WP 75 encrypt auth cookies

mmcgahan:
This was sort of an experimental security improvement, but it works well and provides a solid base for applying encryption to our current auth-related cookies and any secret-related cookies that we might want to pass around in the future.

The 'breaking' part of the update is that it requires a new random string environment variable: `COOKIE_ENCRYPT_SECRET`. When we update mup-web, we'll need to add this secret through kubernetes.

### DIFF
--- a/src/util/authUtils.test.js
+++ b/src/util/authUtils.test.js
@@ -1,0 +1,36 @@
+import {
+	// applyAuthState,
+	applyServerState,
+	// configureAuthState,
+	// removeAuthState,
+} from './authUtils';
+
+describe('applyServerState', () => {
+	const serverWithState = {
+		state: () => {},
+	};
+	it('calls server.state for each auth cookie name', () => {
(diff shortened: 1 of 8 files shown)


src/util/authUtils.test.js
```diff
@@ -0,0 +1,36 @@
+import {
+	// applyAuthState,
```
sadafie: commented out imports

mmcgahan: yeah, these need tests, so it's sort of a TODO but I didn't want to junk up this branch with unrelated tests - would you prefer the commented lines be removed or is it alright to leave them here as an indication that there are untested functions in authUtils. I'll go ahead and make a follow-up task

sadafie: ah okay. totally fine, just wasn't sure!

PR #68, opened 2016-11-29
WP 81 validate /api requests

mmcgahan:
In order to make the `/api` endpoint a bit more secure, this PR adds payload validation, which requires separating the `POST` route handler from the `GET/DELETE/PATCH` route handler.

Validation happens at two levels:

1. Only known querystring/post-body keys are allowed (`queries`, `metadata`, `logout`) - only `queries` is required.
2. The JSON-parsed `queries` value must be an array of query objects.

There's also a new `validation` module where we can start building up a set of validation schemas for use throughout the server code.

### DIFF
--- a/src/apiProxy/api-proxy.js
+++ b/src/apiProxy/api-proxy.js
@@ -9,6 +9,9 @@ import { duotoneRef } from '../util/duotone';
 import {
 	applyAuthState,
 } from '../util/authUtils';
+import {
+	querySchema
+} from '../util/validation';
 
 const parseResponseFlags = ({ headers }) =>
 	(headers['x-meetup-flags'] || '')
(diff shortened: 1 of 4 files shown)


mattkime:
i'm a bit confused about what kind of vulnerability this is addressing. could you describe the security concern?

mmcgahan:
The main principle is that this validation check prevents mal-formed requests from touching the `/api` router `handler` function at all because validation happens before Hapi does any deeper parsing of the incoming request, so it prevents some types of attacks from ever running code that could touch a database or an authorization endpoint, etc. It's a "best practice" update.

It's mainly a prevention of hypothetical attacks - e.g. if a Hapi security vulnerability was discovered that crashed the server when you requested `/api?foo=bar`, this would (likely) prevent our server from crashing. In a general sense, this setup is sort of like a type declaration, where the endpoint declaratively specifies what requests it will respond to, providing additional assurance that the requests are coming from a source that is configured to work with this endpoint.

There might even be some DDoS benefits for the same reason, but I'm not sure.

PR #70, opened 2016-11-29
WP 95 allow queries to specify API endpoint - deprecate apiConfigCreators

mmcgahan:
The API-agnostic benefit of the `apiConfigCreators` module introduces more dev burden than it alleviates for many/most API calls. Instead, the PR allows queries to specify the `endpoint` value directly, and passes through the query's `params` to the API unmodified. This requires engineers to have a bit more familiarity with the spec of the Meetup API endpoints they are using and may introduce some redundancy for different queries accessing the same endpoint, but it does help encapsulation and allows product development to move faster within the consumer app, without having to worry about whether the platform supports a particular endpoint through `apiConfigCreators`

### DIFF
--- a/src/apiProxy/api-proxy.js
+++ b/src/apiProxy/api-proxy.js
@@ -91,15 +91,12 @@ export const parseApiResponse = ([response, body]) => {
  */
 export function queryToApiConfig({ endpoint, type, params, flags }) {
 	if (!endpoint) {
-		const configCreator = apiConfigCreators[type];
-		if (type in configCreator) {
+		if (!(type in apiConfigCreators)) {
 			throw new ReferenceError(`No API specified for query type ${type} and no endpoint provided`);
 		}
-
-		return {
-			...configCreator(params),  // endpoint, params
-			flags,
-		};
+		const baseConfig = apiConfigCreators[type](params);
+		endpoint = baseConfig.endpoint;
+		params = baseConfig.params;
 	}
 	return {
 		endpoint,
--- a/src/apiProxy/api-proxy.js
+++ b/src/apiProxy/api-proxy.js
@@ -91,6 +91,10 @@ export const parseApiResponse = ([response, body]) => {
  */
 export function queryToApiConfig({ endpoint, type, params, flags }) {
 	if (!endpoint) {
+		console.warn(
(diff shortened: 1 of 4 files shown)


eilinora:
LGTM!

mmcgahan:
@akdetrick @sadafie I'd like to get your eyes on this when you get the chance just to make sure it jives with how you think the query system can/should work

sadafie:
Sorry, I don't work Wednesdays, so just getting around to this. I'll leave any comments/questions even though it's merged already, and we can have an offline conversation if needed. thanks!

src/apiProxy/api-proxy.js
```diff
@@ -89,13 +89,22 @@ export const parseApiResponse = ([response, body]) => {
  * @param {Object} query a query object from the application
  * @return {Object} the arguments for api request, including endpoint
  */
-export function queryToApiConfig({ type, params, flags }) {
-	const configCreator = apiConfigCreators[type];
-	if (!configCreator) {
-		throw new ReferenceError(`No API specified for query type ${type}`);
+export function queryToApiConfig({ endpoint, type, params, flags }) {
+	if (!endpoint) {
```
sadafie: can we log `type` or `ref` so we know which api needs updating? Also, I wouldn't mind a ToDo to remove this code in future minor versions?

src/apiProxy/apiConfigCreators.js
```diff
@@ -4,6 +4,7 @@
  * API
  *
  * @module apiConfigCreators
+ * @deprecated
```
sadafie: is this entire module deprecated now?

mmcgahan: The entire module is deprecated, but I wanted to leave it around until our current apps (mup-web, admin-web, starter) have refactored their queries to not need it. Since a few people are using the `npm link` method of making updates to the platform alongside app changes, I didn't want them to get in a confusing state where this had disappeared from `master` before they were ready - I'll do followups in the next sprint to change over those apps

src/apiProxy/api-proxy.js
mmcgahan: good call - will probably push that update straight to `master` since it's simple

PR #73, opened 2016-12-05
Adding 204:NoContent return

eilinora:
Adding a nocontent check as when the API returns a NoContent success message the state saves an error as their is no JSON to parse


src/util/apiUtils.js
```diff
@@ -82,14 +82,24 @@ export const parseApiResponse = requestUrl => ([response, body]) => {
 		};
 	}
 	try {
-		value = JSON.parse(body);
-		if (value && value.problem) {
-			value = formatApiError(new Error(`API problem: ${value.problem}: ${value.details}`));
+		if (response.statusCode === 204) { //NoContent response type
+			value = {
+				success: true
+			};
```
mmcgahan: will `NoContent` _always_ mean that something was successful? I'm assuming this is expected for certain `POST` responses? I'm wondering if it might make more sense to put this `success: true` key on `meta` rather than `value`.

I haven't used this kind of response before (obviously), so we should make a long-term design decision now

mmcgahan: in fact, thinking about this a bit more, I'm wondering if we should always put the statusCode in the `meta` object for all responses...

src/util/apiUtils.js
```diff
@@ -82,14 +82,27 @@ export const parseApiResponse = requestUrl => ([response, body]) => {
...
+		} else {
+			value = JSON.parse(body);
+			if (value && value.problem) {
+				value = formatApiError(new Error(`API problem: ${value.problem}: ${value.details}`));
+			}
+			return {
+				value,
+				meta,
+			};
```
mmcgahan: This is getting more nested than I'd like - there are a lot of ways of exiting the function, which might make it hard to debug. I think we can simplify by taking a step back.

1. do we need a `meta.success` prop, or can we assume a "`value`-with-no-`error`-prop" is a success?
2. if the underlying issue is that we're trying to `JSON.parse` an empty value, why not just skip `JSON.parse` when the value is empty?

would this simple change address the problem?

```js
value = body && JSON.parse(body) || {};
```

mmcgahan:
This needs tests - it should be pretty easy to mock up a `204` response in `apiUtils.test.js` to make sure it's handled as expected

src/util/apiUtils.js
```diff
@@ -69,6 +69,16 @@ export const parseApiValue = ([response, body]) => {
 		return formatApiError(new Error(response.statusMessage));
 	}
 	try {
+		if (response.statusCode === 204) { //NoContent response type
+			return {
+				value: {},
+				meta: {
+					success: true,
```
mmcgahan: is this useful? unless there's a `value.error`, we should be able to assume it's a success response

eilinora: true

PR #76, opened 2016-12-07
WP 137 - integration tests to start server and serve routes

mmcgahan:
So far, none of our unit tests actually start the platform server - this PR adds some 'component'/'integration' tests to the server start function to make sure a dummy app can start and serve routes.

Big change: new `tests_integration/` root-level directory. Since integration/component tests aren't tied to a single module, it would be awkward to try to keep them alongside the modules they integrate. Instead, I've created a new directory and a new naming pattern to distinguish unit tests (which are run by `npm run test:unit` and include code coverage) and integration tests (which are run by `npm run test:integration` and do not generate a coverage report).

`npm test` now runs _both_ types of tests in sequence - unit testing first, then integration tests.


jest.int.config.json
```diff
@@ -0,0 +1,11 @@
+{
+  "setupFiles": [
+    "<rootDir>/src/util/globals.js"
+  ],
+  "moduleFileExtensions": [
+    "js",
+    "jsx"
```
mattkime: will integration tests ever be `.jsx` files?

mattkime:
lgtm

jest.int.config.json
```diff
@@ -0,0 +1,11 @@
...
+  ],
+  "testRegex": "tests_integration\/.*\\.int\\.jsx?$"
```
sadafie: if we're ever going to have other types of tests in a test directory, i'd opt for `tests/integration/` here instead.

mmcgahan: i don't _think_ we would ever have other types of tests, but `tests/integration/` works for me! Will update and merge unless you wanted to look at anything else

PR #77, opened 2016-12-09
WP 138 confine CSRF handling to /api endpoint

mmcgahan:
This PR disables the CSRF plugin on all routes by default, and updates the API routes to opt-in to CSRF handling so that only API responses will set new CSRFs. It also updates the `config` reducer to read new CSRFs whenever they are supplied in the `meta` property of an action - this provides a little more assurance that a new CSRF will be correctly stored in app state whether or not it was supplied by a specific `API_SUCCESS` action or not.

Over time, I suspect we'll need to lock down the ways in which `/api` is called even more to avoid bugs, but this is a good non-breaking step

### DIFF
--- a/src/reducers/platform.js
+++ b/src/reducers/platform.js
@@ -40,10 +40,13 @@ export function config(state={}, action) {
 		apiUrl,
 		trackId;
 
-	switch(action.type) {
-	case 'API_SUCCESS':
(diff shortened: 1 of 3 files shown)


eilinora:
Just tested this with AvatarMember component and everything is working as expected now :)

PR #80, opened 2016-12-14
LogoutLink Component

sadafie:
append `logout` query param to any path.


src/components/logoutlink.test.jsx
```diff
@@ -0,0 +1,24 @@
...
+
+	it('exists', function() {
+		expect(this.linkEl).not.toBeNull();
+	});
+
+	it('creates an HTML a element', function() {
+		expect(this.linkEl.nodeName).toBe('A');
+	});
+
+	xit('appends the logout query param', function() {
```
sadafie: i've disable this test for now because it fails. however, this seems straight forward / should work. but `this.linkEl.getAttribute('href')` is returning null.

sadafie:
Trying out this "request a review" feature. If it's annoying or unhelpful, lmk.

src/components/logoutlink.test.jsx
mmcgahan: without a fully-initialized `<Router>`, the `<Link>` component just renders `<a data-reactroot=""></a>`, although I think you can pass an `href` prop to force it to render a particular href in tests. ([source](https://github.com/ReactTraining/react-router/blob/master/modules/Link.js#L103,L129))

But that's not really useful in this case, so I think it would be better to inspect `this.logoutLink` to see if it renders a `<Link>` with a `to` prop that includes `{query: {logout: true}}` - you technically should avoid testing what `<Link>` actually renders anyway and just treat it as a black box.

src/components/LogoutLink.jsx
```diff
@@ -0,0 +1,31 @@
...
+	render() {
+		const {
+			to,
+			children,
+			...other
+		} = this.props;
+
+		return (
+			<Link
+				to={{ pathname: to, query: {logout: true} }}
```
sadafie: i could imagine product-wise wanting a logout link that redirects you to a completely different page or a specific flow. how about making the `to` prop optional and the default being the current URL?

src/components/logoutlink.test.jsx
sadafie: > you technically should avoid testing what <Link> actually renders anyway and just treat it as a black box.

Good point. Let me modify this test case then.

src/components/LogoutLink.jsx
mmcgahan: That would work. We haven't set up redirects in our React Router yet, but we'll probably want to build _some_ of the logout-and-redirect functionality into the regular routing system, e.g. redirecting to the group homepage if someone tries to access a group admin page that they don't have authorization to access.

sadafie: Also, we probably don't want to retain the `logout` query param in the location bar / browser history. Thoughts on this? Have we done anything with history state yet?

mmcgahan: it's pretty straightforward to manipulate history state - [docs](https://github.com/ReactTraining/react-router/blob/master/docs/guides/NavigatingOutsideOfComponents.md#navigating-outside-of-components) - but my initial preference would be to keep the `logout` query param as a step in the navigation history because it's a pretty big change to application state, and distinct from the view/URL that comes before it, so it might be a little strong to `replaceState`

sadafie: does the login action currently drop the logout param? thinking about the logout -> log back in flow.

mmcgahan: good question. maybe not? worth checking out and filing a defect if it doesn't

sadafie: Currently, any sort of routing navigation drops all query params. i.e., `localhost:8000/?logout` -> click `Login` -> `localhost:8000/login`. However, if you got to `localhost:8000/login?logout` -> submit login -> you get an invalid cookie error.

```
Error: Invalid cookie value
Items.serial
/Users/sadaf/git_repo/mup-web/node_modules/hapi/node_modules/statehood/lib/index.js:244:43
done
/Users/sadaf/git_repo/mup-web/node_modules/hapi/node_modules/items/lib/index.js:31:25
internals.decode
/Users/sadaf/git_repo/mup-web/node_modules/hapi/node_modules/statehood/lib/index.js:201:32
Iron.unseal
/Users/sadaf/git_repo/mup-web/node_modules/hapi/node_modules/statehood/lib/index.js:311:24
process.nextTick
/Users/sadaf/git_repo/mup-web/node_modules/hapi/node_modules/hoek/lib/index.js:854:22
_combinedTickCallback
internal/process/next_tick.js:67:7
process._tickDomainCallback
internal/process/next_tick.js:122:9
```

i'll file a defect

mmcgahan: Hm I don't get the `Invalid cookie value` error, but it does log me out right after a successful login because it's refreshing the `?logout` URL. Send the defect over to me and I can fix it next sprint

sadafie: @mmcgahan https://meetup.atlassian.net/browse/WP-148 thx!

sadafie:
RE: not requiring a `to` property, and defaulting to the current route.

Unfortunately router context information and router properties are only accessible to `connected` components that have a `route` associated with them in `router.js`. Child components need to have the router location information passed in. So, in this branch, I've made the default of `<LogoutLink>` `to` prop be the root route `/`, and any other path is passed in via the `to` prop.

mmcgahan:
@sadafie that update to default `to` makes a lot of sense, and the more I think about it, the more I realize that we really won't have that many `LogoutLink`s outside of PageWrap, so we shouldn't have to fight against that limitation very often - there's typically just one logout link on a page, and it's in the site chrome.

sadafie:
@mmcgahan cool. good learning experience for me anyways. this is ready to merge i think. unless you have anything else to add.

PR #81, opened 2016-12-14
WP-144 read forwarded request protocol if available

mmcgahan:
in order to correctly set the `apiUrl` with the correct protocol, we need to get the forwarded protocol from the SSL proxy if it's available.


swozniak:
LGTM

PR #82, opened 2016-12-14
MW-494 don't use Secure cookies in dev

mmcgahan:
A recent Hapi upgrade made all cookies 'Secure' by default, which means they could only be sent over HTTPS.

Since our dev environment is currently not served over HTTPS, no cookies were active, and things broke.

Badly.

This PR sets `isSecure` to `false` in dev for every affected cookie

### DIFF
--- a/src/plugins.js
+++ b/src/plugins.js
@@ -9,8 +9,14 @@ import requestAuthPlugin from './plugins/requestAuthPlugin';
  */
 
 export function getCsrfPlugin(secret) {
+	const register = (server, options, next) => {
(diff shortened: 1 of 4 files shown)


eilinora:
LGTM, but do we/can we have tests for these plugins?

mmcgahan:
@eilinora the CSRF plugin is an [external dependency](https://github.com/electrode-io/electrode-csrf-jwt) with its own tests, and the other modules in this PR also have tests, although I'm reluctant to check for the `isSecure` setting because it's sort of a hack for dev rather than part of the desired functionality spec

PR #85, opened 2016-12-15
WP-93 adding DELETE method support

eilinora:
Adding support for DELETE methods for api requests 

--- 
Prerequest to tests passing:
https://github.com/meetup/meetup-web-mocks/pull/7


src/apiProxy/apiProxyRoutes.js
```diff
@@ -41,7 +41,7 @@ const getApiProxyRoutes = (path, env, apiProxyFn$) => {
 	const apiPostRoute = {
 		path,
 		handler,
-		method: 'POST',
+		method: ['POST', 'DELETE'],
```
mmcgahan: to keep things simple, I think it's better to keep `DELETE` as part of the `GET` route - it's fine to send the payload as part of the querystring and then you don't have to worry about `delete` as an additional special case in `fetchQueries`

src/util/apiUtils.js
```diff
@@ -172,6 +172,7 @@ export const buildRequestArgs = externalRequestOpts =>
 			externalRequestOptsQuery.url += `?${dataParams}`;
 			externalRequestOptsQuery.headers['X-Meta-Photo-Host'] = 'secure';
 			break;
+		case 'delete':
```
mmcgahan: unnecessary - no params need to be sent in a DELETE request to the REST API - it wouldn't know what to do with them anyway.

src/util/fetchUtils.js
```diff
@@ -32,7 +32,7 @@ export const fetchQueries = (apiUrl, options) => (queries, meta) => {
 		}
 	}
 	const searchString = `?${params}`;
-	const fetchUrl = `${apiUrl}${isPost ? '' : searchString}`;
+	const fetchUrl = `${apiUrl}${(isPost) ? '' : searchString}`;
```
mmcgahan: redundant parens

PR #86, opened 2016-12-16
Added an admin api endpoint

atrout:
Simple admin api endpoint to pass queries through from admin-web


src/apiProxy/apiConfigCreators.js
```diff
@@ -71,6 +71,16 @@ export function event(params) {
 }
 
 /**
+ * admin endpoints
+ */
+export function admin(params) {
+	return {
+		endpoint,
```
mmcgahan: `endpoint` is undefined here.

The current version of the platform allows you specify the `endpoint` directly in your query - no more need for changes to `apiConfigCreators`, which is now deprecated

atrout: Ok, this is a little confusing. I tried to do away with the apiConfigCreator and just use the endpoint in the photosQuery, but I got an error about the type being undefined.

mmcgahan: yeah, probably not using the latest meetup-web-platform - `v0.10.436`.

When you upgrade, you might get a `action$... ofType is not a function` or something like that. You'll need to `import rxjs;` at the top of your `browser-app-entry` and `server-app-entry` files to fix that, but I think the rest of the upgrade should be straightforward [more info here](https://github.com/meetup/meetup-web-platform/blob/master/CHANGELOG.md)

atrout:
Ok, don't need this.

PR #87, opened 2016-12-16
WP 119 language path prefix for routes

mmcgahan:
This PR isolates the process of determining the language of the rendered app, and adds logic to check for a language code url path prefix (e.g. `meetup.com/fr-FR/`).

The main effect of requesting the app in the non-default language is just that the app will load all static assets from the associated language, including the correct TRNs. However, it currently doesn't forward the language information to the REST API, so any strings in the API response are still un-translated. This will be fixed in a followup bug related to setting/sending language cookies (WP-124)

Technical details:

In order for the React Router routes to automatically account for the root-level URL change, we need to inject the `basename` of the URL path into the `history` consumed by the Router. The router can then read routes that are language-path-free, and `history` will automatically append the correct base path. There are a couple extra gymnastics needed to replicate this `history` behavior on the server, which doesn't have a native History API to manipulate.

**BREAKING CHANGE**: new `baseUrl` argument for the renderers and a few signature changes to consume the baseUrl.


src/util/languageUtils.js
```diff
@@ -0,0 +1,47 @@
...
+import Accepts from 'accepts';
+
+export const LANG_DEFAULT = 'en-US';
+
+export const getLanguage = (request, supportedLangs, defaultLang=LANG_DEFAULT) => {
+	const firstPathComponent = request.url.path.split('/')[1];
+	const urlLang = supportedLangs.includes(firstPathComponent) ? firstPathComponent : null;
+	const browserLang = Accepts(request).language(supportedLangs);
+
+	const langSourcePreference = [
```
eilinora: So if i'm reading this structure right, it checks the url being request first for a language request, then the browser for a language cookie set and if nothing uses the default lang... This is probably premature question, but what about language preferences from the member's profile?

mmcgahan: yup, correct - there's a sequence of about 3-4 tasks to get this fully up to spec - member profile language support will be done in [WP-122](https://meetup.atlassian.net/browse/WP-122)

src/components/dom.jsx
```diff
@@ -30,6 +30,7 @@ function getInnerHTML(__html) {
  */
 const DOM = (props) => {
 	const {
+		baseUrl,
```
eilinora: should we also be updating the PropTypes list?

eilinora:
Why are u considering this a breaking change? It _seems_ to have fallbacks for if baseUrl isn't set on the Dom in the logic. Curious what I'm missing :D

mmcgahan:
@eilinora yeah I was a little unsure whether this should be considered a breaking change, but there's more info in the changelog - since a few of the `export`ed functions have changed their signatures, it's technically a breaking change even if the main entry points into the platform have default args that shouldn't generally require updates.

That said, mup-web should definitely be updated to provide the `baseUrl`, so bumping the version number should be a good signal to other devs that they should check the MWP changelog for upgrade instructions when they change the MWP dependency... I figured the cost of bumping the version was worth avoiding the risk of some random edge case issue.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## [0.11]
+
+- **Refactored** Renderers - the `makeRenderer` functions in `server-render`
```
eilinora: more questions :)

So a for instance use of this might be, that it could be used if we want the application to server all url's under `/admin`. Would this mean the routes would _not_ need to be prefixed with `/admin` in the route path setup? (e.g. const `route = { path: 'dashboard' }` vs `const route = { path: 'admin/dashboard' }`

i'm not 100% sure what is meant by "Corresponding changes in the signatures to a few application-related modules mean that if you're relying on direct imports of those modules, your code will need to be updated."

mmcgahan: re: serve all routes with `/admin` prefix: yes

re: changelog wording: now that I look back at the changes, none of the functions whose signatures changed (e.g. `getHtml`) are actually `export`ed, so no consumers could be using them anyway... I'll roll back the `breaking change` label + code - one of the things I was protecting against was a third party `import`ing something like `getHtml` and then needing to add `baseUrl` to the function call. But that's not possible.

eilinora:
LGTM! I'll leave open for sadaf and others to review

sadafie:
>In order for the React Router routes to automatically account for the root-level URL change, we need to inject the basename of the URL path into the history consumed by the Router. The router can then read routes that are language-path-free, and history will automatically append the correct base path.

Conceptually not sure if I'm understanding this correctly. By default (i.e., w/o a language cookie) we render the app paths as they are now, but when the language changes, we render `<baseurl>/<localeCode>/<path>`. Why does the router need additional information appended in the history? I think thats the part I'm not understanding. 

Otherwise, code changes seem fine to me as well.

mmcgahan:
@sadafie for language paths, `localeCode` === `baseUrl`, although in theory you could serve the app from a different baseUrl like `/admin/`.

The `baseUrl`/`basename` config allows the router to treat `path: '/'` as `path: '/<baseUrl>/'`, so you can use the same Route `path`s from different root paths without having to rewrite them all.

There is a little more description in the HistoryJS docs: https://github.com/mjackson/history#using-a-base-url

There's a similar concept in Express: http://expressjs.com/en/api.html#req.baseUrl

PR #88, opened 2016-12-21
Filtering out cf-ray header aaand logging all other headers for review

swozniak:
We're seeing forbidden errors from Cloudflare when making requests to api.meetup.com from within the node app. We suspect the uniquely generated "cf-ray" header is being sent through, which Cloudflare is flagging and therefore rejecting the request.

### DIFF
--- a/src/util/apiUtils.js
+++ b/src/util/apiUtils.js
@@ -165,8 +165,6 @@ export const buildRequestArgs = externalRequestOpts =>
 		const externalRequestOptsQuery = JSON.parse(JSON.stringify(externalRequestOpts));
 		externalRequestOptsQuery.url = encodeURI(`/${endpoint}`);
 
-		delete externalRequestOptsQuery.headers['cf-ray'];
-
 		if (flags) {
 			externalRequestOptsQuery.headers['X-Meetup-Request-Flags'] = flags.join(',');
 		}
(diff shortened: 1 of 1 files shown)


src/util/apiUtils.js
```diff
@@ -165,6 +165,8 @@ export const buildRequestArgs = externalRequestOpts =>
 		const externalRequestOptsQuery = JSON.parse(JSON.stringify(externalRequestOpts));
 		externalRequestOptsQuery.url = encodeURI(`/${endpoint}`);
 
+		delete externalRequestOptsQuery.headers['cf-ray'];
```
mmcgahan: this works, but most of the 'dirty headers' from the incoming request are cleaned out in `parseRequest` around line 227 - ideally this would happen there as well so that we have a 'clean' set of headers from input to output. happy to merge so you can test it out, though

src/util/apiUtils.js
```diff
@@ -227,6 +223,9 @@ export function parseRequest(request, baseUrl) {
 	delete externalRequestHeaders['host'];
 	delete externalRequestHeaders['accept-encoding'];
 	delete externalRequestHeaders['content-length'];  // original request content-length is irrelevant
+	delete externalRequestHeaders['cf-ray']; // unique cloudflare id which will trip up api requests if present
+	
+	console.log(`External request headers: ${JSON.stringify(externalRequestHeaders)}`);
```
mmcgahan: sorry to confuse - `externalRequestHeaders` is set up here, but `buildRequestArgs` adds a few more, so this log won't catch `'X-Meta-Photo-Host'` and `'X-Meetup-Request-Flags'`, for example - that won't affect what you're inspecting, though, so i'll merge

PR #89, opened 2016-12-21
uncovered more cloudflare headers to filter out

swozniak:
Continued from https://github.com/meetup/meetup-web-platform/pull/88

### DIFF
--- a/src/util/apiUtils.js
+++ b/src/util/apiUtils.js
@@ -183,6 +183,8 @@ export const buildRequestArgs = externalRequestOpts =>
 			break;
 		}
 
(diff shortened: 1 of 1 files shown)


mattkime:
Perhaps we should be whitelisting headers to pass on instead of blacklisting. Seems odd to have a cdn specific concern in this location.

mmcgahan:
@mattkime yeah, we might want to go that direction.

> **Mike McGahan** do you know much about `cf-ray` and whether we should just make a header whitelist or something?
> **Steve Wozniak** my concern with a whitelist is that there are a bunch of headers we might want?

the whitelist might be a little awkward to maintain because it might be long and we would probably discover headers that we need to add over time, but would certainly make issues like this go away.

PR #90, opened 2016-12-22
Gracefully handle /api errors

mmcgahan:
CSRF errors, oauth errors, etc

The main culprit is the `onPreResponse` handler that will generate an HTML response when there's an error in the request/response to `/api`, which the app will choke on because it's not JSON.

### DIFF
--- a/src/apiProxy/apiProxyRoutes.js
+++ b/src/apiProxy/apiProxyRoutes.js
@@ -32,13 +32,16 @@ const getApiProxyRoutes = (path, env, apiProxyFn$) => {
 			enabled: true,
 		}
 	};
-
+	const app = {
+		jsonError: true,
+	};
 
 	const apiGetRoute = {
 		path,
 		handler,
 		method: ['GET', 'DELETE', 'PATCH'],
 		config: {
+			app,
 			plugins,
 			validate: {
 				query: validApiPayloadSchema
@@ -50,6 +53,7 @@ const getApiProxyRoutes = (path, env, apiProxyFn$) => {
 		handler,
 		method: 'POST',
 		config: {
+			app,
 			plugins,
 			validate: {
 				payload: validApiPayloadSchema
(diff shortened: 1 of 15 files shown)


src/apiProxy/apiProxyRoutes.js
```diff
@@ -27,12 +27,22 @@ const getApiProxyRoutes = (path, env, apiProxyFn$) => {
 	 *   by `apiAdapter.apiResponseToQueryResponse`
 	 */
 	const handler = getApiProxyRouteHandler(proxyApiRequest$);
+	const plugins = {
+		'electrode-csrf-jwt': {
+			enabled: true,
+		}
+	};
```
mmcgahan: this was accidentally removed in a previous merge

mmcgahan:
I want to check the route config `ext` to see if i can add a custom `onPreResponse` that will override the HTML-generating one. That might be cleaner than the `jsonError` config

mmcgahan:
okay all set - API_ERROR actions now add a `state.app.error` value that the consumer app can use to do something in the UI

src/routes/appRoute.js
```diff
@@ -1,12 +1,39 @@
+import React from 'react';
+import ReactDOMServer from 'react-dom/server';
 import { getAppRouteHandler } from './appRouteHandler';
 
 /**
+ * This function provides global error handling when there is a 500 error
+ */
+export const onPreResponse = {
```
mmcgahan: this was moved from being a global `onPreResponse` handler to being app-route-specific

src/util/tracking.js
```diff
@@ -123,7 +123,7 @@ export const trackApi = log => (response, queryResponses, metadata={}) => {
 	}
 	// special case - login requests need to be tracked
 	const loginResponse = queryResponses.find(r => r.login);
-	if (loginResponse) {
+	if ((loginResponse && loginResponse.login.value || {}).member) {
```
sadafie: is `loginResponse.login` ever undefined here?

mmcgahan: no - the `.find(r => r.login)` will only return something that has a truthy `.login` property

sadafie: cool

PR #92, opened 2016-12-24
WP-111 Single sign-on

mmcgahan:
This branch has forward and backward compatibility with the MEETUP_MEMBER cookie auth from Meetup Classic - it allows you to use the login flow on the platform to get a 'real' MEETUP_MEMBER cookie that will work on Classic, and will use a MEETUP_MEMBER cookie provided by Classic as your logged-in auth on the platform.

One effect of this change is that the logged-out/anonymous auth is now separate from the logged-in auth.

Logged-out cookies provided by the `/authorize` and `/access` oauth endpoints:
- `oauth_token`
- `refresh_token` 

Logged-in cookies provided by the `/sessions` API endpoint:
- `MEETUP_MEMBER`
- `MEETUP_MEMBER_DEV`

Logging in will clear `oauth_token` and `refresh_token`
Logging out will clear `MEETUP_MEMBER` and provide _new_ logged-out tokens


package.json
```diff
@@ -65,5 +65,8 @@
     "tough-cookie": "2.3.2",
     "url-search-params": "0.6.1",
     "uuid": "3.0.1"
+  },
+  "devDependencies": {
+    "iron": "4.0.4"
```
mmcgahan: this is used for mocking an encrypted cookie (which is done in Hapi using Iron, but Iron isn't an exposed package)

src/util/apiUtils.test.js
```diff
@@ -20,9 +22,9 @@ import {
 	apiResponseDuotoneSetter,
 	buildRequestArgs,
 	errorResponse$,
+	getAuthHeaders,
```
sadafie: this could probably just be a `import * as` at this point?

src/util/authUtils.js
```diff
@@ -83,3 +83,21 @@ export const applyServerState = (server, options) => {
 	server.state('refresh_token', authCookieOptions);
 };
 
+export const assignMemberState = options => (request, reply) => {
+	const memberValue = options.API_HOST.includes('.dev.') ?
+		request.state.MEETUP_MEMBER_DEV :
+		request.state.MEETUP_MEMBER;
+	request.state.MEETUP_MEMBER = memberValue;
+	request.state.MEETUP_MEMBER_DEV = memberValue;
```
sadafie: why is there different state for the dev vs non-dev member?

src/plugins/requestAuthPlugin.js
```diff
@@ -24,7 +30,7 @@ function verifyAuth(auth) {
 
 const handleLogout = request => {
 	request.log(['info', 'auth'], 'Logout received, clearing cookies to re-authenticate');
-	return removeAuthState(['oauth_token', 'refresh_token'], request, request.plugins.requestAuth.reply);
+	return removeAuthState(['MEETUP_MEMBER', 'oauth_token', 'refresh_token'], request, request.plugins.requestAuth.reply);
```
sadafie: does `MEETUP_MEMBER_DEV` also need to be removed?

src/plugins/requestAuthPlugin.js
```diff
@@ -47,22 +53,20 @@ export const requestAuthorizer = auth$ => request => {
...
-			.do(() => request.log(['info', 'auth'], 'Request does not contain auth token'))
-			.flatMap(request =>
-				auth$(request)
-					.do(applyAuthState(request, request.plugins.requestAuth.reply))
-					.map(() => request)
-			)
-	);
+
+	const authType = request.state.oauth_token && 'oauth_token' ||
+		request.state.MEETUP_MEMBER && 'MEETUP_MEMBER';
```
sadafie: does the dev meetup member need to be handled here too?

src/util/apiUtils.js
```diff
@@ -26,6 +27,13 @@ const MOCK_RESPONSE_OK = {  // minimal representation of http.IncomingMessage
 	},
 };
 
+export const createCookieJar = apiUrl => {
+	if (url.parse(apiUrl).pathname === '/sessions') {
```
sadafie: nitpick, but can this check was moved to where the method is being invoked in `buildRequestArgs`. Seems more a part of that business logic.

src/util/authUtils.js
mmcgahan: this is adapting to the setup we have in Meetup Classic, where you get MEETUP_MEMBER_DEV from www.dev.meetup.com, and MEETUP_MEMBER from www.meetup.com - I've tried to 'normalize' that setup here so that you can just reference `MEETUP_MEMBER` throughout the rest of the code, instead of having something like `if (C.IS_DEV) { MEETUP_MEMBER_DEV...` all over the show.

One thing that we might want to talk about is whether we should restrict dev to pointing to `www.api.dev.meetup.com`, the way that Meetup Classic dev works, rather than letting devs set the `API_HOST` env variable themselves - there's some complexity/constraints either way, but one of the costs of keeping it flexible is this 'normalizing' code to apply the right MEETUP_MEMBER cookie value.

src/plugins/requestAuthPlugin.js
mmcgahan: yes - I've added `isDev` to the plugin state so that it can choose the right cookie to remove. don't want to force-logout from both prod and dev if it's not necessary!

src/plugins/requestAuthPlugin.js
mmcgahan: at this point in the request cycle, `request.state.MEETUP_MEMBER` will contain the right value. Do you think I should just add an `isDev` check here instead of manipulating `request.state.MEETUP_MEMBER`?

src/util/apiUtils.test.js
mmcgahan: My preference is to leave them as explicit imports because it makes it easier to see what `export`ed functions actually have tests, although the coverage report should also be doing that.

It's also partially just a [habit from Python](https://docs.python.org/2/howto/doanddont.html#at-module-level)

src/util/authUtils.js
```diff
@@ -83,3 +86,27 @@ export const applyServerState = (server, options) => {
...
+	request.state.MEETUP_MEMBER = memberValue;
+
+	return reply.continue();
+};
+
+export const setPluginState = options => (request, reply) => {
+	// Used for setting and unsetting state, not for replying to request
+	request.plugins.requestAuth = {
+		reply,
+		isDev: options.API_HOST.includes('.dev.'),
```
mmcgahan: this will be moved to `server.app.isDevConfig` when #93 merges

src/plugins/requestAuthPlugin.js
```diff
@@ -200,8 +205,8 @@ export const getAuthenticate = authorizeRequest$ => (request, reply) => {
 			request.log(['info', 'auth'], 'Request authenticated');
 		})
 		.subscribe(
-			({ state: { oauth_token } }) => {
-				const credentials = oauth_token;
+			({ state: { MEETUP_MEMBER, oauth_token } }) => {
```
mmcgahan: this could be refactored to read from MEETUP_MEMBER_DEV when using dev endpoints...

src/util/authUtils.js
sadafie: Gotchya

> One thing that we might want to talk about is whether we should restrict dev to pointing to www.api.dev.meetup.com, the way that Meetup Classic dev works, rather than letting devs set the API_HOST env variable themselves

I think I agree with this notion for two reasons: 1) consistency with Meetup Classic will probably mitigate some confusion otherwise since the two systems would work "the same". 2) fewer possible sources of errors/issues on dev... i.e., if you're on dev you always get the dev endpoints then you don't have to double check separate configurations.

src/plugins/requestAuthPlugin.js
sadafie: Personally, I'm in favor of the explicit isDev checks, since that tends to be more clear though potentially more (distributed) code and less concise. If you haven't seen the code that does the request.state.MEETUP_MEMBER manip, it's not clear that state is updated with the correct member at this point.

src/plugins/requestAuthPlugin.js
```diff
@@ -24,7 +30,8 @@ function verifyAuth(auth) {
 
 const handleLogout = request => {
 	request.log(['info', 'auth'], 'Logout received, clearing cookies to re-authenticate');
-	return removeAuthState(['oauth_token', 'refresh_token'], request, request.plugins.requestAuth.reply);
+	const memberCookie = request.server.app.isDevConfig ? 'MEETUP_MEMBER_DEV' : 'MEETUP_MEMBER';
```
sadafie: :100:

src/plugins/requestAuthPlugin.js
```diff
@@ -47,22 +54,20 @@ export const requestAuthorizer = auth$ => request => {
...
-		request$
-			.do(() => request.log(['info', 'auth'], 'Request does not contain auth token'))
-			.flatMap(request =>
-				auth$(request)
-					.do(applyAuthState(request, request.plugins.requestAuth.reply))
-					.map(() => request)
-			)
-	);
+
+	const authType = request.state.MEETUP_MEMBER && 'MEETUP_MEMBER' ||
```
sadafie: You know, I think it may just be a naming confusion for me. Since the cookies are `MEETUP_MEMBER` and `MEETUP_MEMBER_DEV`,  I assume that other usages of `MEETUP_MEMBER` would reference a prod member. Nitpick, but could we just change the state key to be `request.state.MEMBER`?

src/plugins/requestAuthPlugin.js
sadafie: I think we should be doing that, no?

src/util/apiUtils.js
```diff
@@ -201,56 +223,79 @@ export const apiResponseToQueryResponse = query => ({ value, meta }) => ({
...
+export function getAuthHeaders({ state }) {
+	if (!state.MEETUP_MEMBER && state.oauth_token) {
+		return {
+			authorization: `Bearer ${state.oauth_token}`,
+		};
+	}
+	const cookies = { ...state };
+	const csrf = uuid.v4();
+	cookies.MEETUP_CSRF = csrf;
+	cookies.MEETUP_CSRF_DEV = csrf;
```
sadafie: Can we document how the CSRF token / api exchange works? I've already forgotten the conversation we had with Doug awhile back.

src/util/apiUtils.js
```diff
@@ -201,56 +223,79 @@ export const apiResponseToQueryResponse = query => ({ value, meta }) => ({
...
+	delete externalRequestHeaders['host'];  // let app server set 'host'
+	delete externalRequestHeaders['accept-encoding'];  // let app server set 'accept'
 	delete externalRequestHeaders['content-length'];  // original request content-length is irrelevant
-	delete externalRequestHeaders['cf-ray']; // cloudflare headers we don't want to pass on
+
+	// cloudflare headers we don't want to pass on
+	delete externalRequestHeaders['cf-ray'];
 	delete externalRequestHeaders['cf-ipcountry'];
 	delete externalRequestHeaders['cf-visitor'];
 	delete externalRequestHeaders['cf-connecting-ip'];
```
sadafie: at this point, you could just iterate over an array of headers to delete.

src/plugins/requestAuthPlugin.js
mmcgahan: technically `MEETUP_MEMBER` will contain the correct value at this point in the request lifecycle, but I'll change it to be more explicit

src/util/apiUtils.js
mmcgahan: yeah for sure - docs forthcoming

src/util/authUtils.js
mmcgahan: https://meetup.atlassian.net/browse/WP-158

src/util/apiUtils.js
mmcgahan: I was debating between putting the check in `buildRequestArgs` vs putting it here, but I decided to put it here because it keeps the logic in `buildRequestArgs` simple and declarative, and gives `createCookieJar` sole responsibility for creating a cookie jar value - if someone were to ask "where is the logic related to assigning a cookie jar to a request?", I think `createCookieJar` makes more sense than `buildRequestArgs`.

There's actually still more logic in `buildRequestArgs` than I'd like, particularly the `switch` statement.

src/util/apiUtils.js
mmcgahan: probably - this code wasn't changed in this PR, though. I just added comments to make it more clear why each item was being deleted.

Definitely good for 2017 cleanup

willh-meetup:
@mmcgahan Is this ready to merge?

PR #93, opened 2016-12-28
Server app runtime - isDevConfig

mmcgahan:
I am starting to run into more cases where it would be helpful to know if the `.dev.meetup` API urls are being used, which is determined independently of whether the app server is running in 'development' or 'production' mode. Hapi provides a convenient place to store that 'runtime' state, in `server.app`, which is also accessible from every Hapi request as `request.server.app`.

This adds a single value to the `server.app` namespace - `isDevConfig`.


willh-meetup:
👍

PR #95, opened 2016-12-30
WP-143 include all x-meetup headers in api response

mmcgahan:
Also adds `x-total-count` value, and everything is assigned to a key that is the camelCased version of the header name.

Check out the tests for examples.


src/util/apiUtils.test.js
```diff
@@ -214,6 +217,21 @@ describe('parseLoginAuth', () => {
 	});
 });
 
+describe('parseMetaHeaders', () => {
+	it('parses x-meetup headers intreturns x-meetup-flags as flags object with real booleanso camelcased obj', () => {
```
atrout: I think there are some typos in this sentence? Anyway I don't understand it.

PR #100, opened 2017-01-05
Mw 513 test accepts lang

mmcgahan:
the accepts package will use the first match from the supported language array for ambiguous language codes like en. We need to make sure that supportedLangs has the default en-US language code listed first so that it is chosen ahead of en-AU in that case.

### DIFF
--- a/src/util/languageUtils.test.js
+++ b/src/util/languageUtils.test.js
@@ -19,10 +19,11 @@ const MOCK_HAPI_REPLY = {
 	redirect() {},
 };
 const defaultLang = 'en-US';
+const similarToDefault = 'en-AU';
 const altLang = 'fr-FR';
 const altLang2 = 'de-DE';
 const altLang3 = 'es-ES';
-const supportedLangs = [defaultLang, altLang, altLang2, altLang3];
+const supportedLangs = [similarToDefault, defaultLang, altLang, altLang2, altLang3];
 describe('getCookieLang', () => {
 	it('returns undefined when no cookie in state', () => {
 		const request = { ...MOCK_HAPI_REQUEST };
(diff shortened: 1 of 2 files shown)


src/util/languageUtils.test.js
```diff
@@ -60,6 +61,13 @@ describe('getBrowserLang', () => {
 		const lang = getBrowserLang(request, supportedLangs);
 		expect(lang).toBe(false);
 	});
+	it('returns en-US instead of en-AU for "en" if en-US is preferred', () => {
```
mattkime: we should make sure that this respects the priority of languages as provided by the browser. I believe this is why my browser was working with most sites but failing with mup-web - https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.4

mmcgahan: it does ([docs](https://www.npmjs.com/package/accepts#languagelanguages)) - you were getting en-AU because `Accepts` was seeing that your first language preference was `en`, which corresponds to "any variant of English", and the first match for "any variant of English" in our app's `supportedLangs` array was `en-AU`. I've now made sure that our `supportedLangs` array is sorted so that `en-US` is first, which means that `en` will always go to `en-US`.

Technically, having `en` as your browser's top language pref makes having any other variant of English in the language preference list redundant. `en` should be specified _after_ all other specific variants have been tried.

I've now added a test to confirm that non-first-preferred languages can still be matched.

willh-meetup: Just for fun: https://tools.ietf.org/html/rfc4647#section-3.2

mmcgahan:
merge at will - it should get a greenkeeper PR for mup-web so we can get the latest

PR #101, opened 2017-01-06
WP 76 avro log

mmcgahan:
I've got a [transform stream](https://nodejs.org/api/stream.html#stream_class_stream_transform) set up to convert our tracking logs into Avro buffers that can be sent over the wire to Chapstick.

The structure of the transform stream is taken from the streams that we're already using for logging data to `stdout`, so it might appear odd at first, but [this is the reference example](https://github.com/hapijs/good-console/blob/master/lib/index.js) of what they should look like according to the HapiJS team.

The code itself is relatively simple, but we're currently manually copying the [Avro 'Activity' schema JSON](https://github.dev.meetup.com/meetup/meetup/blob/master/modules/base/src/main/versioned_avro/Activity_v3.avsc) from chapstick here. There will be followups to set things up to automatically load the latest Activity schema, but there needs to be some other work done before that's possible.

Once I test that the log info is correctly recorded by Josh's endpoint, I'll remove the `Prelim` label and we should be able to start logging all the things.


src/util/tracking.js
```diff
@@ -124,61 +129,66 @@ export const trackApi = log => (response, queryResponses, metadata={}) => {
...
-		agent: requestHeaders['user-agent'],
+	const eventDate = new Date();
+	const record = {
+		timestamp: eventDate.getTime().toString(),
+		requestId: uuid.v4(),
+		ip: requestHeaders['remote-addr'] || '',
+		agent: requestHeaders['user-agent'] || '',
 		platform: 'meetup-web-platform',
-		platform_agent,
+		platformAgent: 'WEB',  // TODO: set this more accurately, using allowed values from avro schema
```
willh-meetup: RE the comment here, I think `"WEB"` actually  makes sense (provided `platform` is a thing we can key off of).

src/plugins/good-meetup-tracking.js
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	/**
+	 * @param {Error|null} err error returned from failed request
+	 * @param {http.IncomingMessage} response the response object
+	 * @param {String} body the body of the response
+	 * @return {undefined} side effects only
+	 */
+	static postDataCallback(err, response, body) {
+		if (err) {
```
willh-meetup: Two things here:

1. Is it worth checking for a non-200 response here?

2. Not sure if we know the answer to this, but.... If the request times out, will we get an `err`?

src/plugins/good-meetup-tracking.test.js
```diff
@@ -0,0 +1,123 @@
...
+
+describe('GoodMeetupTracking', () => {
+	describe('static postDataCallback', () => {
+		it('logs an error when an error occurs', () => {
+			spyOn(global.console, 'error');
+			const expectedError = new Error('nope');
+			GoodMeetupTracking.postDataCallback(expectedError, null, null);
+			expect(global.console.error).toHaveBeenCalledWith(expectedError);
+		});
+		it('logs an error when response status is not 200', () => {
```
willh-meetup: Seems like maybe we don't need that non-200 check after all?

src/util/tracking.js
mmcgahan: yeah after talking with Tom a few weeks ago, I'm not sure if `platformAgent` should always be `WEB` and `platform` should be `mup-web|admin-web|pro-web`, or whether `platform` should always be `meetup-web-platform` and then we could use `platformAgent` to distinguish logs coming from `mup-web|admin-web|pro-web`. I kind of like the latter arrangement because `platform: 'meetup-web-platform'` looks correct and we can treat each consumer as an agent, but that doesn't seem to mesh with current usage, so ¯\_(ツ)_/¯

willh-meetup: ah, got it. Anyways, seems cool for now :).

src/plugins/good-meetup-tracking.js
mmcgahan: I believe that a timeout would populate the `err` value, but a non-200 response would _not_ populate `err` since the `response` is 'successful' even though it indicates a server error.

So I think we do need both checks - one for `err` and one for non-200 response.

willh-meetup: Yeah, that sounds right.

PR #103, opened 2017-01-07
MW-492 'NotFound' component to allow 404 status on server render

mmcgahan:
This new component uses a similar pattern to `Helmet` because it provides side effect information to the server once the main application has been rendered. If the API has not returned valid results, consumer applications should render a `NotFound` component, which will provide a signal to the server to return a 404 status code with the HTML response.

### DIFF
--- a/src/components/NotFound.jsx
+++ b/src/components/NotFound.jsx
@@ -4,10 +4,16 @@ import withSideEffect from 'react-side-effect';
 /**
  * Use this component as a wrapper for your 'not found' UI when the API returns
  * no data
+ *
+ * Note that only one child is allowed for this component since it does not
+ * provide any rendered output itself
  */
 class NotFound extends React.Component {
 	render() {
-		return React.Children.only(this.props.children);
+		if (this.props.children) {
+			return React.Children.only(this.props.children);
+		}
+		return null;
 	}
 }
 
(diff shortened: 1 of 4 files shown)


src/components/NotFound.jsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+function reducePropsToState(propsList) {
+	if (!propsList.length) {
+		return;
+	}
+	return 404;
+}
+
+function handleStateChangeOnClient(code) {
```
sadafie: is this empty function needed?

mmcgahan: it is. `withSideEffect` needs to know what to do when the app gets rendered on the server and then what happens when it gets re-rendered on the client, but in this case we want to no-op on the client. For better or worse, the package [forces this to be a function](https://github.com/gaearon/react-side-effect/blob/master/src/index.js#L13)

mmcgahan:
We do have a [404 component](https://github.com/meetup/mup-web/blob/10d1a6ec923e6ee2f47c09289af65e63379b1c70/src/features/shared/errors/404.jsx) in mup-web, but it's a dummy debug thing that needs some product/design attention.

mmcgahan:
Any other changes needed, @sadafie ?

sadafie:
Nope, merge away!

PR #105, opened 2017-01-11
WP 148 login from logout

mmcgahan:
This PR fixes an issue where if you logged out from the login page, and then tried to log in again, you couldn't. On successful login, the `?logout=true` query string will now be cleared.


sadafie:
LGTM

PR #106, opened 2017-01-11
WP 157: Clear cache and app state on logout API_REQUEST

mmcgahan:
we no longer have a dedicated 'LOGOUT_REQUEST', so this PR re-applies the state-clearing rules from LOGOUT_REQUEST to API_REQUEST when `action.meta.logout` is truthy.


mmcgahan:
Logging out is now purely a navigation action (using querystring param) rather than a standalone Redux action. That change was made to simplify reloading the app in the logged-out state and tracking and such.

PR #107, opened 2017-01-12
Allow query functions to return null when no query needed

mmcgahan:
@jmhammons has a setup with a param-matching route that will return different query objects based on the URL params, but _some_ of the matching params don't need to trigger API queries at all, so the query function returns `null`. These `null`s need to be filtered out before being sent to the app server's `/api` endpoint


src/epics/sync.test.js
```diff
@@ -25,23 +24,44 @@ import getSyncEpic from '../epics/sync';
 import * as syncActionCreators from '../actions/syncActionCreators';
 import * as authActionCreators from '../actions/authActionCreators';
 
+const MOCK_ROUTES = [
```
mmcgahan: yes - I added it there and copied it here so I didn't have to wait for the package to publish, but it looks like greenkeeper has made the update so I can clean up this branch

mmcgahan:
good to go?

PR #109, opened 2017-01-12
remove inert and js-cookie

mattkime:
depcheck says they're not used. i could test and build without them.

still would be good to get a confirmation on this.


sadafie:
`js-cookie` is being removed in #102 and replaced by `cookie` cc @mmcgahan

mmcgahan:
@sadafie actually `js-cookie` was removed in #102 because it wasn't being used (it's a client-side cookie helper), and `cookie` was added because it was being used for some server-side cookie parsing, so this cleanup is fine.

PR #111, opened 2017-01-13
Port linkify function from meetup classic

nlaz:
#### Related Issues

Implements https://meetup.atlassian.net/browse/WP-166

#### Description
This PR creates a util function that parses text and wraps `<a>` elements around urls that match the specified regex pattern.

#### Notes
There are some security concerns around turning user input into HTML. I want to do some investigation into best practices here. Suggestions are welcome. There's also some standards on what makes up a valid url. I would like to doublecheck (and test) our regex pattern matches those standards.

#### Prelim 
Adding the `Prelim` tag until tests are added, function is verified on `mup-web`, and those above concerns 👆 are answered. Just trying to get this out here so people can weigh in.


src/util/linkify.js
```diff
@@ -0,0 +1,28 @@
...
+		createLink = (href) => {
+			const target = options.target || '',
+				targetAttr = `target="${target}"`;
+
+			if (target === '_blank') {
+				targetAttr += ' rel="noopener noreferrer"';
+			}
+
+			return `<a href="${href}" title="${href}" ${targetAttr}>${href}</a>`;
+		};
```
mmcgahan: Just to be Better Than Before, let's define this `createLink` function outside of the `linkify` function, and curry the `options` into it, i.e.

```js
const urlRegex = const urlRegex = /((((ht|f){1}(tps?:[/][/]){1}))[-a-zA-Z0-9@:;%_\+.~#?&/=]+)/gm;

 // this is a function that takes `options` object and returns a function that takes `href` string
const createLink = options => href => { ... } 

function linkify(text, options) {
  if (!text) {
    return '';
  }

  return text.replace(URL_REGEX, createLink(options || {}));
}
```

src/util/linkify.js
```diff
@@ -0,0 +1,28 @@
...
+	options = options || {};
+
+	const urlRegex = /((((ht|f){1}(tps?:[/][/]){1}))[-a-zA-Z0-9@:;%_\+.~#?&/=]+)/gm,
+		createLink = (href) => {
+			const target = options.target || '',
+				targetAttr = `target="${target}"`;
+
+			if (target === '_blank') {
+				targetAttr += ' rel="noopener noreferrer"';
+			}
```
mmcgahan: in order to avoid the Travis error and be more accurate with the var name, this block could be

```js
const relAttr = target === '_blank' ? 'rel="noopener noreferrer"' : '';

return `<a href="${href}" title="${href}" ${targetAttr} ${relAttr}>${href}</a>`;
```

src/util/linkify.test.js
```diff
@@ -0,0 +1,42 @@
...
+		const expectedParagraphLink = `Did you know ${expectedLink} is a cool site?`;
+		expect(linkify(paragraphTextBase)).toBe(expectedParagraphLink);
+	});
+});
+
+describe('Link Regex', () => {
+	it('should match basic url', () => {
+		const httpBase = 'http://www.meetup.com';
+		expect(URL_REGEX.test(httpBase)).toBe(true);
+	});
```
mmcgahan: url regexes should probably have a pretty comprehensive test suite - check out https://github.com/kevva/url-regex/blob/master/test.js for an example - TBH, it might be worth just importing that `url-regex` package and not worrying about tests. It looks pretty solid.

src/util/linkify.js
```diff
@@ -1,28 +1,34 @@
...
+ *
+ * @method createLink
+ * @param {Object} options optional modifiers
+ * @param {String} href string of link passed through options
+ * @return {String} The HTML link tag
+ */
+const createLink = options => href => {
+	const target = options.target || '';
+	const targetAttr = `target="${target}"`;
+	const relAttr = target === '_blank' ? ' rel="noopener noreferrer"' : '';
```
mmcgahan: nitpick, but for consistency, don't put the extra space before `rel` here - put it in between `targetAttr` and `relAttr` in the the string template in the `return` statement. Even if it's an empty string HTML won't care that there's useless whitespace

src/util/linkify.test.js
nlaz: I think you are right about using a dependency here. It will also cut out some concerns around unsafe url characters https://perishablepress.com/stop-using-unsafe-characters-in-urls/.

src/util/linkify.js
nlaz: you got it 👍

nlaz:
Removed the url regex ported from chapstick and instead imported a library for this. Not sure how we feel about introducing a dependency here but I think it serves more advantages (security concerns) than disadvantages (dependency management).

nlaz:
@mmcgahan do you think this might need more test cases since we aren't testing regex anymore? Maybe some common use cases *just to be safe*? Happy to add more.

package.json
```diff
@@ -64,6 +64,7 @@
     "rxjs": "5.0.3",
     "source-map-support": "0.4.10",
     "tough-cookie": "2.3.2",
+    "url-regex": "^4.0.0",
```
mmcgahan: better to set an exact version for now.

nlaz:
Verified this PR on `mup-web`. Removed the `Prelim` tag.

PR #112, opened 2017-01-13
Refactor util/createStore to use store enhancer pattern

mmcgahan:
Our current `createStore` module uses custom functions to apply platform-specific functionality Redux stores in consumer apps. However, it is much easier to compose custom `createStore` function by using [Store enhancers](https://github.com/reactjs/redux/blob/master/docs/Glossary.md#store-enhancer), which can be layered on top of each other in a predictable way. Using this standard/documented pattern will make it easier to maintain and modify the platform store in the future, e.g. to inject click tracking functionality.


src/util/createStore.js
```diff
@@ -59,24 +51,40 @@ const serverFetchQueries = request => (api, options) => {
...
+export function getBrowserCreateStore(
+	routes,
+	middleware
+) {
+	const middlewareEnhancer = getPlatformMiddlewareEnhancer(
+		routes,
+		middleware,
+		fetchQueries
+	);
+	const enhancer = compose(
```
willh-meetup: compose 🙌

PR #113, opened 2017-01-14
MW 585 Redux actions for click tracking

mmcgahan:
This PR dispatches CLICK_TRACK actions for every trackable DOM event (mostly clicks, some 'change' events like in Meetup Classic), and brings in all the code that is related to registering the click event handlers and parsing the event data to construct useful log data.

Consuming this data into application state will be done separately, in WP-167

screen shot 2017-01-18 at 2 04 00 pm


mmcgahan:
hmm this needs more tests

mmcgahan:
Testing is awkward for this one since it is so DOM-dependent and the 'fake dom' used in Jest test seems to fail silently with some of the dom traversal methods that are used in the utility. I think testing will just have to be limited for this one for now

src/util/clickTracking.js
```diff
@@ -0,0 +1,154 @@
+const DATA_ATTR = 'clicktrack';
+
+function cleanTrackingUrl() {
+	// TODO: investigate and refactor to support this feature
+
+	/*
+	// remove tracking information from the url to prevent bad link copying
```
sadafie: Just curious - does this happen in meetup classic at the moment, or how is it handled there?

src/util/clickTracking.js
```diff
@@ -0,0 +1,154 @@
...
+const originalStopPropagation = typeof Event !== 'undefined' && Event.prototype.stopPropagation;
+
+/**
+ * Overriding stopPropagation for maximum click tracking
+ * @param {Function} trackClick function to track clicks
+ * @return {undefined} side effects only
+ */
+export function trackStopPropagation(trackClick) {
+	/**
+	 * A special jQuery events method that hooks the click tracking into events
```
sadafie: jquery!? :)

src/util/clickTracking.js
```diff
@@ -0,0 +1,154 @@
...
+ * Get a descriptive string for the HTML element passed in, e.g.
+ *
+ * `tagName#id.class1.class2`
+ *
+ * @param {HTMLElement} element the element to describe
+ * @param {Boolean} idOnly only use id in description
+ */
+function elementShorthand(element, idOnly) {
+	const id = element.id ? `#${element.id}` : '';
+	if (id && idOnly) {
```
sadafie: Do we actually ever need the classnames if we have an id?

src/util/clickTracking.test.js
```diff
@@ -0,0 +1,66 @@
+import * as clickTracking from './clickTracking';
+
+/*
+ * **** DISABLED ******
+ *
+ * This is disabled because tests don't have `global.documnet` defined in
```
sadafie: What about travis is different is different than the local node env that would make `global.document` unavailabe?

Also, typo in "document"

src/util/createStore.js
```diff
@@ -5,9 +5,19 @@
...
 import { fetchQueries, mergeCookies } from '../util/fetchUtils';
+import getClickTracker from './clickTracking';
 
 const noopMiddleware = store => next => action => next(action);
 
+export const clickTrackEnhancer = createStore => (reducer, initialState, enhancer) => {
+	const store = createStore(reducer, initialState, enhancer);
+	const clickTracker = getClickTracker(store);
+	document.body.addEventListener('click', clickTracker);
+	document.body.addEventListener('change', clickTracker);
```
sadafie: Doesn't `click` also get fired for `change` events on input, select, and text areas? Would this send clickTracking data twice?

sadafie: Do we care about `dblclick`, `focus`? Same concerns with sending the data multiple times across events?

src/util/clickTracking.js
mmcgahan: yeah, this is part of the clickTracking script in Meetup Classic, but obviously we don't have particular tracking-URL support in the new platform, and I imagine that might be a little ways off. I didn't want to delete the code but also didn't want to try to get it working without any purpose/application. I suspect it's something that we would need to make an Investigation plan for, so maybe that's the best next step.

src/util/clickTracking.js
mmcgahan: ewww. will fix.

src/util/clickTracking.js
mmcgahan: Since the classnames might be updated dynamically in response to user activity, there are probably cases where you would like to know if it's the `#magicButton.active` instead of the `#magicButton.disabled`, for example

src/util/clickTracking.test.js
mmcgahan: It is strange, and I'm not sure why it behaves so differently in Travis, but it seemed like a bigger issue than what I could tackle in this branch. Even locally, it was very awkward to try to get the `window.Event` object to reset correctly after each test. This script would ideally get tested in a real browser environment but I can file a followup ticket to see whether it can be refactored to be a little more node-friendly unless you feel that it's a blocker.

src/util/createStore.js
mmcgahan: This is copied from what is happening in Meetup Classic, and it looks like the `change` listener was added only in order to pick up change events on `<select>` elements with a `data-clicktracking` attribute - there's a check at the top of the `clickTracker` function that bails for `change` events on elements that don't have the attribute.

I'm not sure what product motivations went into that, but I think you're right that those elements would get some redundant `click` data in addition to the `change` data when the user was using a mouse. I'm reluctant to 'fix' that here, though, without a better defined specification of what is actually needed.

src/util/clickTracking.js
sadafie: yeah let's do that

src/util/clickTracking.js
sadafie: mm good point

src/util/createStore.js
sadafie: Cool, let's maybe do a follow up investigation for this. It makes sense to replicated Meetup Classic but at the same time I feel like we could better understand / implement some of this.

src/util/clickTracking.js
mmcgahan: https://meetup.atlassian.net/browse/WP-195

src/util/createStore.js
mmcgahan: My initial preference is to let things go as they are currently implemented and let product teams decide whether the data is suited to their needs - I'm not sure how much we can really investigate in the abstract without product applications and user data to analyze in the near future, but feel free to add an Investigation task if you have a sense of how to approach the issue

PR #129, opened 2017-01-20
read correct auth from request when new auth cookies are set

mmcgahan:
https://meetup.atlassian.net/browse/WP-173


mattkime:
lgtm. should there be tests as well?

mmcgahan:
if it's okay, i'll add tests in a followup? I can do them now but it will take another 15-20min

mmcgahan:
tests here https://github.com/meetup/meetup-web-platform/pull/130

PR #133, opened 2017-01-23
change api proxy URL.

willh-meetup:
I think this is the only change in the platform code.

@mmcgahan But, I think we need to change `src/sw.js` in `mup-web`, right?


mmcgahan:
and [here](https://github.com/meetup/meetup-web-platform/blob/master/docs/Auth.md#logout)

willh-meetup:
Should be all replaced now.

PR #134, opened 2017-01-24
WP 168 Send clickTracking cookie to app server

mmcgahan:
This PR uses SyncMiddleware + the fetchQueries function to pass along click tracking data from Redux state into a cookie that is sent to the app server on _every_ app server request (navigation, form submission, whatever).

This cookie is never actually stored on the client - it is only a 'cookie' because the cookie header is easy to parse on the app server and has enough 'storage' for the amount of data we want to send.

When the cookie is sent to the server, SyncMiddleware then clears the clickTracking data from state.


mattkime:
I wonder if you considered using a vanilla http header. a cookie is just a http header with some conventions on top. might be more efficient/direct to set aside some of those conventions.

mmcgahan:
Yeah, a custom header is definitely an option - one thing I wanted to do in the first pass was keep the data as similar to what is happening on meetup classic as is practical, so I'm sending a similar cookie header to what is sent there, but you're right that we're not really leveraging the primary benefit of cookies to store state in the browser over time.

One particular advantage of sending a cookie instead of a custom header is that Hapi will automatically parse cookies in the cookie header, so there's one less step to get to the data, but that's pretty trivial.

mmcgahan:
Let me know if the question about custom headers is a blocker or if there are particular benefits of a custom header that you have in mind @mattkime

mattkime:
I was thinking about debating the pros and cons of header vs cookie....which is most appropriate / cause the least confusion in the future, etc...

but this should never affect product work so i don't care! :-)

mmcgahan:
Don't merge yet - the cookie will probably need to be URL encoded in order to be officially valid and avoid throwing errors in `node-fetch`, but I'll need to circle back to check next week

PR #139, opened 2017-01-26
WP-188 ensure JSON-stringified tracking POST request

mmcgahan:
The url-encoded data was not playing nicely with the tracking endpoint, but I have confirmed that JSON-encoded data works just fine. At least, it gives a `{ "message": "success" }` response from the endpoint.

### DIFF
--- a/src/plugins/good-meetup-tracking.test.js
+++ b/src/plugins/good-meetup-tracking.test.js
@@ -1,4 +1,3 @@
-import querystring from 'querystring';
 import avro from 'avsc';
 import GoodMeetupTracking from './good-meetup-tracking';
 import Stream from 'stream';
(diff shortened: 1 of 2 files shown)


src/plugins/good-meetup-tracking.test.js
```diff
@@ -79,9 +77,7 @@ describe('Integration with tracking logs', () => {
 	});
 
 	it('encodes standard output from logTrack', () => {
-		const tracker = new GoodMeetupTracking({
-			postData() {},
```
mmcgahan: this is no longer needed since by default `postData` will no-op unless `NODE_ENV=production`. Leaving the default behavior in place allows us to force the test to hit the real endpoint with `NODE_ENV=production yarn test`

mmcgahan:
I was originally sending url-encoded data, and the endpoint either supports an encoding that is slightly different, or it doesn't support url-encoded data at all and it needs JSON. Either way, JSON is good.

willh-meetup:
👍

PR #144, opened 2017-01-30
This fixes the broken cookie parsing, in possibly a roundabout way...

willh-meetup:
… but it seems to work.

### DIFF
--- a/src/util/stringUtils.js
+++ b/src/util/stringUtils.js
@@ -17,10 +17,22 @@ export function toCamelCase(s) {
 	return s.replace(/-(\w)/g, g => g[1].toUpperCase());
 }
 
-/*
- * Remove surrounding quotes from a string (' and ").
+/**
+ *
+ * It appears that sometimes we add some data that shouldn't be in cookies.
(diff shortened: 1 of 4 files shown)


src/util/stringUtils.js
```diff
@@ -17,10 +17,22 @@ export function toCamelCase(s) {
...
  */
-export function removeSurroundingQuotes(str) {
-	return str.replace(/^["|']|["|']$/g, '')
+
+export function cleanRawCookies(rawCookies) {
+	return encodeURI(
+		rawCookies
+	)
+	.replace(/^[']|[']$/g, '') //replace surrounding single quotes
+	.replace(/;%20/g, '; ') //Make sure we unencode the actual cookie separator
```
willh-meetup: This sucks, but the weirdness is... URI encoding seems to fix the issue, but I'm not entirely sure at this point what the cookie parser is getting tripped up on.

willh-meetup:
@mmcgahan Not scriptable, but I'll send you my cookie headers from my non-incognito browser window.

PR #149, opened 2017-02-01
WP 196 Convert URLSearchParams to URL API

mmcgahan:
https://meetuphq.slack.com/archives/web-platform/p1485980819004304

IE 11 is not playing nice with our `URLSearchParams` polyfill

- Fortunately, polyfill.io has a `URL` polyfill that can serve the need in the browser
- BUT we then need to polyfill `URL` on Node v6 as well
- BUT polyfill.io uses a browser-only polyfill implementation
- BUT the best `URL` polyfill for Node does _not_ support `URLSearchParams`, which is the only thing we need it for
- BUT Node v7 has native support for `URL`

SO, this PR refactors all the `URLSearchParams` usage to `URL`, which has a `.searchParams` property that is a `URLSearchParams` instance, and bumps the Node requirement for consumer apps to version 7.x.

The major Node upgrade requirement warrants a breaking change, and since we're in production now, might as well bump to 1.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [1.0]
 
 - **Require Node v7** for require('url').URL
+- **Require window.URL** polyfill for browsers without native support
 
 ## [0.11]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [1.0]
+
+- **Require Node v7** for require('url').URL
+
 ## [0.11]
 
 - **Removed** `apiConfigCreators` no longer exists - use the query object
--- a/jest.config.json
+++ b/jest.config.json
@@ -11,6 +11,7 @@
     "js",
     "jsx"
   ],
+  "testEnvironment": "node",
   "testRegex": "src\/.*\\.test\\.jsx?$"
 }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: required
 language: node_js
 node_js:
-  - "6.1"
+  - "7"
 
 branches:  # pushes only get built for master
   only:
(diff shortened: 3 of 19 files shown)


jest.config.json
```diff
@@ -11,6 +11,7 @@
     "js",
     "jsx"
   ],
+  "testEnvironment": "node",
```
mmcgahan: this is required so that JSDom doesn't try to interfere with the global `URL` supplied by node - there are some related test refactors that had to happen alongside this in order to fix tests that touch browser-specific functions

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## [1.0]
```
mattkime: MAJOR VERSION CHANGE

mattkime:
might be worthwhile looking into - http://www.marcusoft.net/2015/03/packagejson-and-engines-and-enginestrict.html#enginestrict

mmcgahan:
thanks for the link @mattkime - I've added those rules into package.json so hopefully that will help ensure consumers are in sync

mattkime:
lgtm

mmcgahan:
will hold off merging until @willh-meetup has a chance to weigh in on the Node v7 upgrade implications

willh-meetup:
Only question from me is... why 7.0.0 instead of 7.5.0?

The current node version should still be stable, it's just that it's going to change a little faster than the LTS version.

I think it's cool for us to run with the latest stable.

mmcgahan:
it's `^7.0.0` just to allow minor-release updates, and the Travis `"7"` version will update to the latest 7.x release automatically - If you think it would be better to lock that down a little more, that's cool with me.

I'll go ahead and merge for the bugfix, but we can tweak the Node version spec as desired

PR #150, opened 2017-02-01
MW 622 hanging integration

mmcgahan:
Complete discussion is in the Jira ticket: https://meetup.atlassian.net/browse/MW-622

tl;dr: the RxJS `.timeout` operator doesn't cancel the internal `setTimeout` when the associated observable completes, so Jest won't exit until the `API_TIMEOUT` is reached for each test that touches `.timeout`

This is an RxJS issue, but we can speed things up by setting a shorter API_TIMEOUT interval until this PR merges: https://github.com/ReactiveX/rxjs/pull/2135

### DIFF
--- a/tests/integration/render.int.js
+++ b/tests/integration/render.int.js
@@ -11,7 +11,7 @@ jest.mock('request', () =>
 				cb(null, {
 					headers: {},
 					statusCode: 200,
-					elapsedTime: 234,
+					elapsedTime: 2,
 					request: {
 						uri: {
 							query: 'foo=bar',
@@ -19,7 +19,7 @@ jest.mock('request', () =>
 						},
 						method: 'get',
 					},
-				}, '{}'), 234)
+				}, '{}'), 2)
 	)
 );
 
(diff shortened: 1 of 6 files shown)


tests/integration/render.int.js
```diff
@@ -67,6 +67,7 @@ describe('Full dummy app render', () => {
 	const random32 = 'asdfasdfasdfasdfasdfasdfasdfasdf';
 	const mockConfig = () => Promise.resolve({
 		API_HOST: 'www.api.meetup.com',
+		API_TIMEOUT: 10,
```
akdetrick: would it be useful to define some of this in a `constants` module in `tests/`?

mmcgahan: probably - the `mockConfig` is now used in enough places that it could be broken out into a `platformMocks` module or something - I'll do that and fix up merge conflicts

trxcllnt:
Sorry, we should have had this merged a long time ago. Angling to get it in this week.

PR #155, opened 2017-02-02
WP 215 incorrect member cookie

mmcgahan:
This started with a simple fix to the member cookie, but that exposed the fact that we were setting a URL-encoded version of the `MEETUP_MEMBER` cookie returned by the API, so I went ahead and had another look at the `Invalid cookie value` issue and found some ways to prevent Hapi from choking on cookies that aren't _just so_. It turns out we can just tell it to ignore the validation step and it will pass along the 'real' MEETUP_MEMBER cookie without an issue.

I've rolled back the `cleanRawCookies` utility @willh-meetup

### DIFF
--- a/src/apiProxy/apiProxyRoutes.js
+++ b/src/apiProxy/apiProxyRoutes.js
@@ -49,6 +49,9 @@ const getApiProxyRoutes = (path, env, apiProxyFn$) => {
 					},
 				},
 			},
+			state: {
+				failAction: 'ignore',  // ignore cookie validation, just accept
+			},
 		},
 	};
 	const apiGetRoute = {
--- a/src/util/apiUtils.js
+++ b/src/util/apiUtils.js
@@ -426,9 +426,6 @@ export const injectResponseCookies = request => ([response, _, jar]) => {
 	}
 	const requestUrl = response.toJSON().request.uri.href;
 	jar.getCookies(requestUrl).forEach(cookie => {
-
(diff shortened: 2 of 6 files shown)


mmcgahan:
My guess is that the 'Invalid cookie value' error [reported in Slack](https://meetuphq.slack.com/archives/engineering/p1486189212001473) is from this code change. Maybe we do need to clean cookies in some way, although I'm skeptical that urlencoding is doing what we need it to do... will need to investigate

willh-meetup:
http://stackoverflow.com/questions/6869866/do-i-need-to-escape-cookie-values-when-setting-from-servlet-api

mmcgahan:
yeah, url encoding definitely makes sure that the cookie doesn't contain any invalid values, but from what I saw before, it also creates a different value from what Meetup classic sends through, at least for the MEETUP_MEMBER cookie, which appeared as `MEETUP_MEMBER="id=123415..."` in the browser when sent from Meetup classic, but `MEETUP_MEMBER=%22id=123415..."` when cleaned-and-set from mup-web. So I think it would be 'correct' if the whole ecosystem was urlencoding and decoding all the cookies, but I don't think that's happening in some places so I think we might create some issues while solving others with urlencoding.

My best guess right now is that there is a meetup classic cookie that sometimes contains a character that is 'invalid' from Hapi's point of view on page render, and we need to apply the same `state.failAction = 'ignore'` setting that was added to the `/mu_api` route in this branch in order for that 'invalid' cookie value to also be ignored by the app-serving route.

PR #161, opened 2017-02-07
log request headers

mmcgahan:


mmcgahan:
didn't add request.id - adding now

PR #162, opened 2017-02-07
Add a /ping route that we can use for new relic.

willh-meetup:


mattkime:
lgtm. this will be very useful in perf testing.

mmcgahan:
I talked briefly with Will on Slack about this, but the pingRoute should ideally be defined in `src/routes/index.js` instead of `src/server.js`, although it sounds like he might have had some trouble getting that to work.

I made a commit https://github.com/meetup/meetup-web-platform/commit/db79a41291f944fcacf7cb8bdd7cdee2861e667c that I think does the right thing, but I haven't tested it beyond the integration test that I added

mmcgahan:
@mattkime re: perf testing - this endpoint would primarily provide a baseline `requests/sec` value from Hapi, but since it doesn't actually touch any of the platform code that we're responsible for, we wouldn't be able to use it as a reference for how our code changes affect performance.

src/server.js
```diff
@@ -36,11 +36,17 @@ export default function start(
 	// (process.env.NODE_ENV === 'production')
 	require('source-map-support').install();
 
+	const pingRoute = {
+		path: '/ping',
+		method: 'GET',
+		handler: (request, reply) => reply('pong!'),
```
mmcgahan: just realized this will also need a `config: { auth: false }` property so that it doesn't try to generate an anonymous oauth token for every request :/

willh-meetup:
I think this is all set now.

PR #166, opened 2017-02-08
clean bad cookies before fetch

mmcgahan:
the `click-track` header from Meetup Classic contains characters that are not allowed by `node-fetch` and we don't need that header for that call, so we can strip it.

### DIFF
--- a/src/util/fetchUtils.js
+++ b/src/util/fetchUtils.js
@@ -82,6 +82,10 @@ export const tryJSON = reqUrl => response => {
 	return response.text().then(text => JSON.parse(text));
 };
 
+export const stringifyCookies = cookies =>
+	Object.keys(cookies)
+			.map(name => `${name}=${cookies[name]}`)
+			.join('; ');
 /**
  * @param {String} rawCookieHeader a 'cookie' header string
  * @param {Object} newCookies an object of name-value cookies to inject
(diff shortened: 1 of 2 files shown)


src/util/fetchUtils.js
```diff
@@ -94,8 +98,17 @@ export const mergeCookies = (rawCookieHeader, newCookies) => {
 		...oldCookies,
 		...newCookies,
 	};
-	return Object.keys(mergedCookies)
-		.map(name => `${name}=${mergedCookies[name]}`)
-		.join('; ');
+	return stringifyCookies(mergedCookies);
+};
+
+export const BAD_COOKIES = [
```
mattkime: this might be the simplest way to deal with this but it would be nice if we didn't need to blacklist.

mmcgahan: I agree. In this case, we need to pass along most of the cookies provided by Meetup Classic, but not this one. Without refactoring the clickTracking setup in Meetup Classic to ensure that it's sending 'officially valid' cookies, however, I think a blacklist is the best compromise solution.

If Meetup Classic adds more cookies that aren't valid in the future, then they will throw errors, although we now have better error logging to help us diagnose and fix the problem.

A whitelist might also work short-term, but it's hard to know when Meetup Classic might start requiring different cookies than what it currently sends.

mmcgahan:
Just FYI, I think actually we're not losing the click tracking data, since the cookie remains in the browser and will eventually be sent to meetup classic, where it will be consumed, recorded, and reset. This just makes sure the cookie isn't sent to the node server. I suspect this will all be re-figured when I finish up the new platform clicktracking tasks

PR #167, opened 2017-02-10
WP-118 React Router v4 - platform update

mmcgahan:
Complete description + docs to come - I'm just using the beta builds in dev for now


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## [2.0]
```
mattkime: 👍

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "meetup-web-platform",
   "version": "0.0.0",
```
mattkime: update the version here too?

mmcgahan: that actually gets set by Travis each time it pushes a release to npm, e.g. https://github.com/meetup/meetup-web-platform/commit/bfac5ab68d1839d428e4e364ec2df0c11b344a9a

it remains `0.0.0` in `master`

mmcgahan:
I need to sort out the nested routes to keep the same sort of 'assume parent path is part of the nested path' structure used by v3

package.json
sadafie: Does travis actually update the package.json version? i.e., will i see it in `node_modules/meetup-web-platform/package.json`?

src/components/RouteLayout.jsx
```diff
@@ -0,0 +1,46 @@
...
+				</route.component>
+			)}
+		/>
+	);
+};
+
+const SwitchRoutes = ({ routes }) => (
+	<Switch>
+		{routes.map((route, i) =>
+			<RouteWithSubRoutes key={i} {...route} />
```
sadafie: Do we envision a case ever where there's a route that needs a different component or different props than what `RouteWithSubRoutes` provides?

src/components/SyncContainer.jsx
```diff
@@ -0,0 +1,44 @@
...
+		dispatchLocationChange: bindActionCreators(locationChange, dispatch),
+	};
+}
+
+/**
+ * @module SyncContainer
+ */
+export class SyncContainer extends React.Component {
+	/**
+	 * This container only cares about routing changes. When the route changes,
```
sadafie: I know you're describing it here, but I'm still unclear as to what this container does, and possibly also why it's a separate component as opposed to functionality with `RouteLayout`

src/renderers/server-render.jsx
```diff
@@ -187,24 +191,13 @@ const makeRenderer = (
...
-		)
-		.flatMap(args => storeIsReady$.map(() => args))  // `sample` appears not to work - this is equivalent
-		.map(getRouterRenderer(store, baseUrl, clientFilename, assetPublicPath));
+	request.log(['app', 'info'], `Finding route for path: '${url}'`);
+	store.dispatch({
+		type: '@@server/RENDER',
+		payload: request.url,
+	});
+	return storeIsReady$
+		.map(() => getRouterRenderer(routes, store, url, baseUrl, clientFilename, assetPublicPath));
```
sadafie: Could you explain how this works now? When we were pairing you were saying that the consumer apps would need to pass in their `AppContainer` components to makeRenderer, but I'm not seeing that now.

package.json
```diff
@@ -60,7 +59,7 @@
     "react-helmet": "^3.3.2",
     "react-intl": "^2.2.3",
     "react-redux": "^5.0.2",
-    "react-router": "^3.0.2",
+    "react-router-dom": "4.0.0-beta.5",
```
mmcgahan: yeah, and that includes a good fix that will slightly speed up our integration tests

package.json
```diff
@@ -80,11 +79,11 @@
...
     "react-addons-test-utils": "15.4.2",
     "react-helmet": "^3.3.2",
-    "react-intl": "^2.2.3",
-    "react-redux": "^5.0.2",
-    "react-router": "^3.0.2",
-    "redux": "^3.5.2",
-    "rxjs": "^5.0.3",
+    "react-intl": "2.2.3",
+    "react-redux": "5.0.2",
+    "react-router-dom": "4.0.0-beta.5",
```
mmcgahan: Yeah I don't love the fact that we need the same package listed twice in order to correctly run tests, but in a 'pure' sense, those peer dependencies are necessary for the tests to run in dev, so they technically should be in both.

I'd be interested in figuring out a better way, but I'm hesitant to add another dependency to deal with dependency challenges... feel free to open a ticket

src/components/RouteLayout.jsx
mmcgahan: There are definitely ways that `RouteWithSubRoutes` could be extended to support different route styles (e.g. using [`render`](https://reacttraining.com/react-router/#route.render) or [`children`](https://reacttraining.com/react-router/#route.children) instead of `component`) - but the current setup essentially supports a `routes` definition that closely parallels the routes definition for RRv3, which keeps it simpler for apps to upgrade to v4 in the short term.

Over time, we should be able to add flexibility, but I'm reluctant to do that now before we know how we're going to manage async routes and how the query system will adapt to different route compositions.

src/components/SyncContainer.jsx
mmcgahan: Okay, I'll reword. This container essentially serves the role of `react-router-redux`, which would dispatch an action each time the route changed. Since `react-router-redux` [doesn't support v4](https://github.com/reactjs/react-router-redux/issues/520#issuecomment-277528454), this component will directly dispatch the `LOCATION_CHANGE` action when the route changes.

src/renderers/server-render.jsx
mmcgahan: yeah, as soon as I started writing `<AppContainer />` for the starter kit, I realized that it would essentially be identical to `RouteLayout` for every application, so we can just keep the `getRenderer` API stable and use a platform-hosted `RouteLayout`.

In the future, I suspect we might want allow apps to pass in a 'custom' `<App />` that implements its own routing, but since the routing config is so closely tied to API requests, I don't think we're ready to support that functionality.

package.json
mmcgahan: technically npm updates the package.json version, but yes

mmcgahan:
@sadafie I believe this is ready for final review - nested route `path` definitions are supported/required now.

sadafie:
This looks good to me. The latest changes broke something in the starter branch, so I'm just trying to track down an error over there. Doubtful it needs to be fixed here, but I'll let you know:

```
TypeError: Cannot read property 'path' of undefined
 29     at Switch.render (/Users/sadaf/git_repo/web-platform-starter/node_modules/react-router-dom/node_mod    ules/react-router/Switch.js:73:70)
 30     at /Users/sadaf/git_repo/web-platform-starter/node_modules/react-dom/lib/ReactCompositeComponent.js    :796:21
 31     at measureLifeCyclePerf (/Users/sadaf/git_repo/web-platform-starter/node_modules/react-dom/lib/Reac    tCompositeComponent.js:75:12)
 32     at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (/Users/sadaf/git_    repo/web-platform-starter/node_modules/react-dom/lib/ReactCompositeComponent.js:795:25)
 33     at ReactCompositeComponentWrapper._renderValidatedComponent (/Users/sadaf/git_repo/web-platform-sta    rter/node_modules/react-dom/lib/ReactCompositeComponent.js:822:32)
```

package.json
```diff
@@ -80,11 +79,11 @@
...
     "react-addons-test-utils": "15.4.2",
     "react-helmet": "^3.3.2",
-    "react-intl": "^2.2.3",
-    "react-redux": "^5.0.2",
-    "react-router": "^3.0.2",
-    "redux": "^3.5.2",
-    "rxjs": "^5.0.3",
+    "react-intl": "2.2.3",
+    "react-redux": "5.0.2",
+    "react-router-dom": "^4.0.0-beta.6",
```
sadafie: i've noticed `yarn add <package>` adds the `^` notation to versions, here and on a few other lines. Just double checking that we're intentionally adding it, and not just missing it.

package.json
```diff
@@ -80,11 +79,11 @@
     "react": "^15.4.2",
     "react-addons-test-utils": "15.4.2",
     "react-helmet": "^3.3.2",
```
sadafie: `^`

src/components/RouteLayout.jsx
```diff
@@ -0,0 +1,49 @@
...
+const RouteWithSubRoutes = route => {
+	if (!route.component) {
+		throw new Error(`route for ${route.path} must have a 'component' property`);
+	}
+	if (route.render || route.children) {
+		console.warn('route.render and route.children function not supported');
+	}
+	return (
+		<Route
+			path={route.path}
```
sadafie: path should be optional here, right?

package.json
mmcgahan: #ItWasntMe

will fix.

package.json
mmcgahan: We'll need to add `.yarnrc` like [the one in the starter kit](https://github.com/meetup/web-platform-starter/blob/master/.yarnrc) so that it doesn't do this. Will update

src/components/RouteLayout.jsx
mmcgahan: Every `route` needs a `path`, I think?

sadafie: Nope: https://reacttraining.com/react-router/#route.path. Any route without a path will always match. I think this might be why I'm getting an [error](https://github.com/meetup/meetup-web-platform/pull/167#issuecomment-282520535) in [web-platform-starter#173](https://github.com/meetup/web-platform-starter/pull/173).

mmcgahan: ah, right. so an empty `path` will remain an empty path - it is optional.

The error you're seeing is from a `route` object being undefined, not the path.

package.json
sadafie: this peer dependency needs to be updated to `-beta.6` as well. This devDependency has been updated already.

src/components/RouteLayout.jsx
sadafie: I redid my `yarn link` and updated some package json version numbers on this branch, and the error has gone away. I think it was just pointing to some wrong version of a module somewhere.

mmcgahan:
In order to avoid making this a major blocker for bugfixes, we may end up merging the beta release into mup-web `master` before this gets merged - we'll just need to update this PR with each non-beta release of MWP to make sure mup-web can have bugfixes in addition to RRv4

src/components/PlatformApp.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { Provider } from 'react-redux';
+import SyncContainer from '../components/SyncContainer';
+import RouteLayout from '../components/RouteLayout';
```
nlaz:
```js
import SyncContainer from './SyncContainer';
import RouteLayout from './RouteLayout';
```
Maybe?

src/components/SyncContainer.jsx
```diff
@@ -0,0 +1,45 @@
...
+function mapDispatchToProps(dispatch) {
+	return {
+		dispatchLocationChange: bindActionCreators(locationChange, dispatch),
+	};
+}
+
+/**
+ * @module SyncContainer
+ */
+export class SyncContainer extends React.Component {
```
nlaz: Curious: Why separate this functionality from the `PlatformApp` component?

src/components/RouteLayout.jsx
```diff
@@ -0,0 +1,49 @@
...
+				</route.component>
+			)}
+		/>
+	);
+};
+
+const SwitchRoutes = ({ routes, currentPath }) => (
+	<Switch>
+		{routes.map((route, i) => {
+			route.path = `${currentPath || ''}${route.path}`;  // modify path to reflect full match
```
nlaz: Set default value in the parameters? `({ routes, currentPath = '' })`?

src/components/RouteLayout.jsx
```diff
@@ -0,0 +1,49 @@
...
+			return <RouteWithSubRoutes key={i} {...route} />;
+		})}
+	</Switch>
+);
+
+
+/**
+ * @module RouteLayout
+ */
+class RouteLayout extends React.Component {
```
nlaz: Isn't this class redundant with `SwitchRoutes`? Could you combine their use?

mmcgahan: yeah true - previous iterations included some differences but it's combinable, thanks 👍

src/components/SyncContainer.jsx
mmcgahan: 2 main reasons

1. `SyncContainer` needs to be a class with a `mapStateToProps` function, which means that it has to live _inside_ `<Provider>` - I can't add a `mapStateToProps` function directly to `PlatformApp` because it is the thing that provides the `<Provider>`. There are ways to get the same functionality without using `connect`, but they're not pretty 🙂.
2. syncing Redux state with the current route is well-defined, distinct feature of the platform app, so encapsulating it here is helpful from a modular architecture perspective - this container is essentially what the `react-router-redux` package used to do in previous versions of React Router, so it's kind of nice to still make it a standalone module.

nlaz: Oh interesting thanks!

mmcgahan:
Because the new setup doesn't make a redundant `/mu_api` request when the app first loads, it doesn't set the `x-csrf-jwt` and `x-csrf-header` cookies that are needed for a user to successfully POST.

This branch will need to be updated to set those cookies in the initial response from the application route.

mmcgahan:
Cookie-setting routine added in #193 - setting `don't merge` label until that PR merges since I've pulled the changes into this branch.

mmcgahan:
Found a route-matching issue - the nested route `path` values are not being built up correctly - they are building on top of the matched url (e.g. `/hq-faff`) instead of the matched path (e.g. `/:urlname`), so params are being lost

I'll add tests and make the fix

sadafie:
The latest commits look good to me

sadafie:
Sorry, didn't mean to close! Pressed the wrong comment button

PR #174, opened 2017-02-17
Linkify - Add protocol to links w/o one

nlaz:
#### Description
This PR fixes an issue with the linkify method that creates links for urls without protocols. This is because the library being used matches urls that don't have protocols. The solution was to check for `www.` at the beginning of the matched url and add the `http` protocol to the beginning. If we want  to support more complex urls, it might be better to use our own regex again.

#### Related Issues
Fixes https://meetup.atlassian.net/browse/MUP-8457

### DIFF
--- a/src/util/linkify.js
+++ b/src/util/linkify.js
@@ -12,8 +12,9 @@ const createLink = options => href => {
 	const target = options.target || '';
 	const targetAttr = `target="${target}"`;
 	const relAttr = target === '_blank' ? 'rel="noopener noreferrer"' : '';
+	const link = (href.search('www.') === 0) ? `http://${href}` : href;
 
-	return `<a class="link" href="${href}" title="${href}" ${targetAttr} ${relAttr}>${href}</a>`;
+	return `<a class="link" href="${link}" title="${href}" ${targetAttr} ${relAttr}>${href}</a>`;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/util/linkify.js
```diff
@@ -12,8 +12,9 @@ const createLink = options => href => {
 	const target = options.target || '';
 	const targetAttr = `target="${target}"`;
 	const relAttr = target === '_blank' ? 'rel="noopener noreferrer"' : '';
+	const link = (href.search('www.') === 0) ? `http://${href}` : href;
```
mmcgahan: probably better to check for the _absence_ of the protocol rather than the _presence_ of `www.`. I guess the URL matcher won't match on other subdomain strings, but this is an indirect test.

nlaz: You are right but it also matches [`https`, `ws` `ftp` and `//`](https://github.com/kevva/url-regex/blob/master/test.js#L49-L61) at the beginning of the string. I'm worried that we might miss a case if we test for protocol. it might be better to go back to our own regex expression.

mmcgahan: yeah, I was thinking that checking for the absence of any supported protocol is pretty trivial

```js
const hasProtocolRE = /^(?:https?:|ws:|ftp:)?\/\//;
const link = hasProtocolRE.test(href) ? href : `http://${href}`
```

nlaz:
@mmcgahan happy with this PR?

PR #178, opened 2017-02-22
set configuration on server.app instead of passing around as arg

mmcgahan:
_This is a precursor to [WP-241](https://meetup.atlassian.net/browse/WP-241) - it makes that refactor more straightforward_

This is a change to how configuration info is passed around the application, and is intended to make it significantly simpler.

TL;DR: instead of passing config properties down through the server application manually, we can just assign the config as a property of `server.app` and read from it directly wherever we have access to `request.server` or `server`.

### DIFF
--- a/src/apiProxy/api-proxy.js
+++ b/src/apiProxy/api-proxy.js
@@ -33,30 +33,26 @@ import {
  * query-specific options.
  *
  * @param {Request} request Hapi request object
- * @param {Object} baseUrl API server base URL for all API requests
  * @return Array$ contains all API responses corresponding to the provided queries
  */
-const apiProxy$ = ({ API_TIMEOUT=8000, baseUrl='', duotoneUrls={} }) => {
+const apiProxy$ = request => {
+	request.log(['api', 'info'], 'Parsing api endpoint request');
+	// 1. get the queries and the 'universal' `externalRequestOpts` from the request
+	const parsedRequest = parseRequest(request);
 
-	return request => {
-		request.log(['api', 'info'], 'Parsing api endpoint request');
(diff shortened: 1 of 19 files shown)


src/apiProxy/api-proxy.js
```diff
@@ -33,30 +33,26 @@ import {
...
-		const apiRequests$ = queries
-			.map(queryToRequestOpts)
-			.map((opts, i) => ([opts, queries[i]]))  // zip the query back into the opts
-			.map(makeApiRequest$(request, API_TIMEOUT, duotoneUrls));
-
-		// 4. zip them together to send them parallel and return responses in order
-		return Rx.Observable.zip(...apiRequests$);
-	};
+const apiProxy$ = request => {
+	request.log(['api', 'info'], 'Parsing api endpoint request');
```
mmcgahan: this is primarily an indentation change because the config values no longer need to be threaded through this function into `parseRequest` and `makeApiRequest$`

src/apiProxy/apiProxyRoutes.js
```diff
@@ -11,13 +7,7 @@ const validApiPayloadSchema = Joi.object({
 	logout: Joi.any(),
 });
 
-const getApiProxyRoutes = (path, env, apiProxyFn$) => {
-	const proxyApiRequest$ = apiProxyFn$({
-		API_TIMEOUT: env.API_TIMEOUT,
-		baseUrl: env.API_SERVER_ROOT_URL,
-		duotoneUrls: getDuotoneUrls(duotones, env.PHOTO_SCALER_SALT),
-	});
```
mmcgahan: these config values are now read directly from the server config in `server.app`

PR #179, opened 2017-02-22
WP 241 call API proxy internally without HTTP request for server render

mmcgahan:
This PR allows the app endpoint on the app server to call into the API proxy endpoint's handler directly rather than using a purely-isomorphic HTTP `fetch` to itself. This allows the app server to skip quite a bit of custom re-authentication and HTTP parsing, which should provide a strong performance benefit.

I will try to get solid performance comparison numbers before this merges.


mmcgahan:
This probably could also use some more updated tests

mmcgahan:
@sadafie the `requestAuthPlugin` changes mainly relate to rolling back the `__internal_oauth_token` cookie that was being set in order to authorize the 'internal' API requests that didn't have auth info sent from the browser.

Now, the `requestAuthPlugin` sets the auth info in `request.plugins.requestAuth.oauth_token` (no need for cookie munging!), which can then be applied to the `Authorization` header when the browser didn't send its own auth info. https://github.com/meetup/meetup-web-platform/pull/179/files#diff-59c6644481955fdc13789b5fa2c12c09R219

sadafie:
@mmcgahan feel free to merge!

PR #182, opened 2017-02-23
Log cleanup 1

mmcgahan:
First pass at implementing logging spec

https://docs.google.com/document/d/1OBskbgbI4TyPKAmdC5aU6gLQ7LPXyCXCJb61yAktikk/edit#

main effect is converting existing `request.log` and `console.log` calls to `console.log(JSON.stringify({ message, ... })`, which will be much easier to find and analyze in Stackdriver.


src/renderers/server-render.jsx
```diff
@@ -94,7 +93,10 @@ const getRouterRenderer = (store, baseUrl, clientFilename, assetPublicPath) =>
 				200;
 		} catch(e) {
 			// log the error stack here because Observable logs not great
-			console.error(e.stack);
+			console.error(JSON.stringify({
+				message: `Internal rendering error ${e.message}`,
```
willh-meetup: I'm not entirely sure what's in `e.message` and `e.stack`, but I assume that it doesn't have the URL that the user was hitting. That might be nice info to pass along?

mmcgahan: this is just a vanilla `try/catch` JS Error https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/prototype

I'll have another look at refactoring `getRouterRenderer` to consume the original Hapi `request` if possible

willh-meetup: Cool. Otherwise this looks fine, would be fine pushing something like that to a follow up.

mmcgahan: Looking into this, I realized I could just leave the logging to the server-level `onPreResponse` function, which has a reference to the `request`, rather than doing anything special here.

PR #185, opened 2017-02-24
send transformed data to next transform - stdout

mmcgahan:
The `console.log` calls from the good-meetup-tracking Stream.Transform were not appearing in production logs - it appears that `console.log` from inside a `good` Stream Transform works differently than `console.log` in other parts of the code (perhaps it's considered some sort of child process? I don't know much about Node Streams).

However, the normal way for `good` console streams to work is for them to pipe to `stdout` in the plugin config, and this output _does_ appear correctly in production logs, so this PR changes the behavior of the tracking stream to send the encoded data to `stdout`

### DIFF
--- a/src/plugins.js
+++ b/src/plugins.js
@@ -62,7 +62,8 @@ export function getConsoleLogPlugin() {
 						}],
 					}, {
 						module: GoodMeetupTracking,
-					}
+					},
+					'stdout'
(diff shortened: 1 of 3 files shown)


mmcgahan:
this is a high priority PR because currently there is no active tracking in production

PR #186, opened 2017-02-25
Wp 231 api success refactor

mmcgahan:
This will require more discussion and probably a presentation in a web eng meeting to get feedback and make sure consumer apps are ready to make the updates required to handle this breaking change.

I've been talking primarily with @nlaz about how to clarify and improve the way that API requests are handled in web platform apps. The main concern about the current query response implementation is that bundling _all_ REST API responses into a single array that gets treated as 'SUCCESS' or 'ERROR' is too crude, and parsing 'SUCCESS' data for errors is non-intuitive and cumbersome, so it would be helpful to separate successful responses from error/invalid responses, and deliver those responses individually rather than munged together alongside their corresponding queries.

This PR also makes a couple other tweaks to make it easier to work with API responses:

1. responses now have an explicit `ref` property rather than having the `ref` as a single top-level key.
2. the HTTP status code of the REST API call is returned as `response.meta.statusCode`
3. setting the CSRF token is done by a standalone action rather than as part of the API_SUCCESS action - it will be easier to see when it is and is not set


eilinora:
@mmcgahan random thoughts, but maybe there is a way to craft this change so it doesn't cause a mega PR. Could we maybe add a deprecation warning for the old implementation and have this available for those who want to use it? Not sure if that is possible with the nature of this change, but just a thought. :)

mmcgahan:
@eilinora I had briefly entertained a similar thought, but didn't get very far with it... the core mechanism is all-or-nothing, but now that I think of it, I think we could have 'parallel actions', with new action `type`s that will eventually replace the old. `API_SUCCESS` and `API_ERROR` would continue to operate like they currently do, but we would have a parallel set of actions that were dispatched according to the new rules.

Proposal:

1. `API_SUCCESS` and `API_ERROR` continue to work like they currently do - they each contain an array of responses from the API
  - deprecation warnings will have to be added manually to places in the code that currently use `API_SUCCESS` and `API_ERROR`, since the platform can't know whether they are being used. A global warning could be added to server startup.
2. a new `API_RESPONSE_OK` action will be dispatched for _each_ successful query response from an individual REST API request
3. a new `API_RESPONSE_ERROR` action will be dispatched for each HTTP 4xx/5xx response from the API
4. a new `API_RESPONSE_FAILURE` action will be dispatched for a HTTP 5xx response from the app server API proxy endpoint

Does that make sense? I agree that the cost of managing mega-PR in mup-web probably outweighs the cost of having parallel API request-response patterns in the short term.

eilinora:
@mmcgahan I really like this proposal. This way everyone gets to opt in over time as they update their code and it's clear to find where the changes are needed as their distinct response types. 

It also gives everyone a chance to make their changes before we cut them off, while helping to offload the burden of managing the code changes who own that domain area vs putting the ownness on the originator of the mega-PR.

Would be great if @nlaz weighs in too, since he is also eager to use this setup.

nlaz:
I think you both are right. It would be another large hoop to jump through to make everyone adapt to this change, and have to manage all the merges. However, I think `API_RESPONSE_SUCCESS` is unnecessarily verbose so I would like to propose another idea. Using constants for `API_SUCCESS`, `API_ERROR` and `API_FAILURE`. We could do a little switch-a-roo and solve two problems in one. Something like:
```js
// meetup-web-platform
export const API_SUCCESS = 'API_RESPONSE_SUCCESS'; // or whatever

// mup-web/blah/blah/eventReducer.js
import { API_SUCCESS} from 'meetup-web-platfom';

switch(action.type){
   case API_SUCCESS:
   case API_ERROR:
```
FYI no one is using constants for `API_SUCCESS` at the moment. All references are strings.

mmcgahan:
Yeah, the suggested 'type' strings are a bit verbose, but I'd argue that's why we have auto completion and gzip :). I'd be reluctant to have `API_SUCCESS` be an alias for something other than `'API_SUCCESS'` as long as `'API_SUCCESS'` exists - I can see that leading to confusion and bugs when people are expecting one thing and getting another.

Since we're abstracting a lot of the "API-ness" out of the system with the Query pattern, what do you think of `Q_REQUEST, `Q_SUCCESS`, `Q_ERROR`, and `Q_FAILURE`? I think that wouldn't be too ambiguous, given that there aren't a lot `Q___` things in the system, but it could also be the same thing with `QUERY_` typed out.

nlaz:
> that's why we have auto completion
:| I guess I need to start using that yikes.

eilinora:
I'd prefer the more verbose `API_RESPONSE_*`. Would definitely prefer  `API_RESPONSE_SUCCESS` over OK. Q_ feels to shorthand to provide context. Honestly people will be copying examples from other people, parts of code so I thinks ok for is to be more explicit.

mmcgahan:
Following up: in order to make a clean cut between the old system and the new system, I'm creating new modules that are modifications of the current sync-related modules rather than trying to splice the new code directly into the existing code. This will make deprecation (and ultimate removal) of the current system _much_ easier at the slight cost of bulkier code in the meantime. The one exception is the current `platform.js` reducer (which populates `state.app`), which will need to process the payloads from existing actions as well as the new actions - this should mean that 

New modules:

1. epics/api
2. actions/apiActionCreators

new actions, keeping the `API_` naming as consistent as possible:

1. `API_REQ` is the new `API_REQUEST` ('req' and 'resp' are common abbreviations in Node, and should satisfy some of our interest in having terse action `type`s).
2. `API_RESP_SUCCESS`
3. `API_RESP_ERROR`
4. `API_RESP_FAIL`
5. `API_RESP_COMPLETE`

I'll export string constants for all of these so that there is less manual string management

mmcgahan:
I've pushed changes related to my previous comment, but I now need to back-port support for the current API_REQUEST setup. I think that I'll need to have 'hybrid' support in

- sync middleware
- cache middleware
- platform reducer

although I think the dual-wielding code changes won't be too dramatic.

mmcgahan:
Updates are nearly complete, so I'm going to start testing in mup-web to make sure this doesn't break anything

1. Sync middleware actually doesn't need to do much to support the old system - API_REQUEST is now treated the same as API_REQ, and both actions will result in API_RESP____ actions. All navigation is channeled through the new query system.
2. Caching is disabled completely for the old system. The only practical cost for this is that any existing async GETs in consumer applications will have to wait for the API to respond. As soon as they convert `apiRequest` to `api.get`, caching will work again.
3. Reducer: the reducer script has a clearly demarcated section to support the deprecated sync action creators, as does the corresponding test script.

mmcgahan:
need to refactor post and delete middleware in order to correctly handle the new output format of fetchQueries

mmcgahan:
Two options for making the PR smaller:

- update cache middleware in followup - keep caching old pattern for now
- update query response object to have `ref` key as prerequisite for this branch - it's a relatively simple refactor but is a little confusing in addition to the other changes.

mmcgahan:
This is feeling pretty good to me - once the blocker PRs get cleared, I think it's ready for final review. I'll kick the tires in mup-web to make sure nothing changes there

mmcgahan:
@nlaz @sadafie @mattkime @eilinora @aquillano @willh-meetup if any/all of you could have a look through this, I think it's good to go. I loaded it up in mup-web and everything appears to be working with no changes there, which hopefully means the deprecation is working as expected. Big changes will be needed in mup-web to fully upgrade, but with this in place it should be easier for product teams to make those changes over the next couple sprints.

mmcgahan:
@sadafie I've added a line item to our weekly meeting to discuss, but yeah, version 3.0 will remove all the deprecated code introduced in 2.x, so we'll need to make a plan for that upgrade.

PR #188, opened 2017-02-27
WP 245 Fix INVALID_JWT when using multiple tabs

mmcgahan:
This PR updates the CSRF header handling to remove it from app state, where it was confined to a particular tab, to a new `x-csrf-header` cookie, which will be updated for all tabs at the same time that the corresponding `x-csrf-jwt` cookie is updated from an API request.

Security-wise, there should be no change, because the CSRF header cookie value is still sent back to the server as a header, which means that we can still be certain that the request was not a cross-site request (which are not allowed to send headers).

As a bonus, this refactor consolidates CSRF management into `fetchUtils` where it belongs, which makes our redux reducer and middleware setup cleaner as well.


src/util/fetchUtils.js
```diff
@@ -27,6 +41,12 @@ export const parseQueryResponse = queries => ({ responses, error, message }) =>
  * @return {Promise} resolves with a `{queries, responses}` object
  */
 export const fetchQueries = (apiUrl, options) => (queries, meta) => {
+	if (
+		typeof window === 'undefined' &&  // not in browser
+		typeof test === 'undefined'  // not in testing env
+	) {
+		throw new Error('fetchQueries was called on server - cannot continue');
```
willh-meetup: 👍

src/util/fetchUtils.js
```diff
@@ -27,6 +41,12 @@ export const parseQueryResponse = queries => ({ responses, error, message }) =>
  * @return {Promise} resolves with a `{queries, responses}` object
  */
 export const fetchQueries = (apiUrl, options) => (queries, meta) => {
+	if (
+		typeof window === 'undefined' &&  // not in browser
+		typeof test === 'undefined'  // not in testing env
```
willh-meetup: Is this a global var that the testing env sets?

src/util/fetchUtils.js
```diff
@@ -1,24 +1,38 @@
...
+	if (csrfHeader) {
+		// set cookie with header value - this ensures all open tabs share the CSRF
+		BrowserCookies.set(CSRF_HEADER_COOKIE, csrfHeader);
+	}
+}
+
 /**
  * Wrapper around `fetch` to send an array of queries to the server. It ensures
- * that the request will have the required Oauth access token and constructs
+ * that the request will have the required OAuth and CSRF credentials and constructs
```
willh-meetup: Might be worth mentioning CSRF is set in a cookie now?

src/util/fetchUtils.js
mmcgahan: yes. Not very happy about having to reference test-specific vars in production code, but I've been starting to think that we should make a clear separation between node code and browser code in the platform, and if we do that, then we can make sure to run tests on this in a simulated browser environment and this can be stripped out again.

I'll make a note to bring it up in our next weekly.

.yarnrc
```diff
@@ -0,0 +1 @@
+save-prefix false
```
mmcgahan: when i added the `js-cookie` dependency, I saw that it set a caret-version, so I went ahead and added this config file so that yarn will set explicit dependency versions in the future.

mmcgahan:
> We're getting the csrf token in the same way, but now instead of storing in the app state, we store it in a cookie.
>
> This means that when a new tab is opened, the cookie updates and when the request happens that is read and the correct token is sent along to the server?

correct. The header-based value will now have the same lifecycle as the cookie-based value, so they can never* get out of sync

*: hopefully never

.yarnrc
mattkime: good call. we should add this to the web-platform-starter

mmcgahan: you mean like this https://github.com/meetup/web-platform-starter/blob/master/.yarnrc 😉

PR #193, opened 2017-03-03
WP-256 Set both CSRF cookies from server

mmcgahan:
There's not really a good reason to require the client to set the `x-csrf-header` cookie - the server can use its 'set-cookie' header to take care of that.

The benefit of letting the server take care of it is

1. Easier to unit test because it doesn't rely on browser-dependent cookie-setting calls
2. the cookie can be set from the initial page load rather than waiting for an API request to return.

Blocker for #167

### DIFF
--- a/src/plugins.js
+++ b/src/plugins.js
@@ -21,11 +21,16 @@ export function getCsrfPlugin(secret) {
 	const register = (server, options, next) => {
 		const cookieOptions = {
 			path: '/',
-			isHttpOnly: true,
 			isSecure: process.env.NODE_ENV === 'production',
 		};
-		server.state('x-csrf-jwt', cookieOptions);  // set by plugin
-		server.state('x-csrf-jwt-header', cookieOptions);  // set by onPreResponse
+		server.state('x-csrf-jwt', {
+			...cookieOptions,
+			isHttpOnly: true,
+		});  // set by plugin
+		server.state('x-csrf-jwt-header', {
+			...cookieOptions,
+			isHttpOnly: false,
+		});  // set by onPreResponse
 		const registration = CsrfPlugin.register(server, options, next);
 		server.ext('onPreResponse', setCsrfCookies);  // must add this extension _after_ plugin is registered
 		return registration;
--- a/src/plugins.js
+++ b/src/plugins.js
@@ -17,22 +17,39 @@ export function setCsrfCookies(request, reply) {
 	return reply.continue();
 }
 
+/**
+ * The CSRF plugin we use - 'electrode-csrf-jwt' compares a cookie token to a
+ * header token in non-GET requests. By default, it will set the cookie token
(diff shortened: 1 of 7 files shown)


src/util/serverUtils.js
```diff
@@ -71,7 +71,7 @@ function onPreResponseExtension(request, reply) {
 		type: 'response',
 		direction: 'out',
 		info: {
-			headers,
+			headers: response.headers,
```
mmcgahan: we were logging the _request_ headers in the log for the _response_ :/

src/util/serverUtils.js
```diff
@@ -92,10 +93,8 @@ export function registerExtensionEvents(server) {
 	server.ext([{
 		type: 'onRequest',
 		method: onRequestExtension,
-	}, {
-		type: 'onPreResponse',
-		method: onPreResponseExtension,
 	}]);
+	server.on('response', logResponse);
```
mmcgahan: turns out the full _response_ headers are not available in `onPreResponse`, so we have to hook into this event in order to correctly log the full response info

mmcgahan:
need to move some stuff around and add tests

src/plugins.js
```diff
@@ -9,12 +9,31 @@ import requestAuthPlugin from './plugins/requestAuthPlugin';
...
+			path: '/',
 			isSecure: process.env.NODE_ENV === 'production',
-		});
-		return CsrfPlugin.register(server, options, next);
+		};
+		server.state('x-csrf-jwt', {
+			...cookieOptions,
+			isHttpOnly: true,
+		});  // set by plugin
+		server.state('x-csrf-jwt-header', {
```
willh-meetup: OK, so as I understand `server.state`, it essentially sets cookies.

And as I understand this, we are setting two cookies with no value, and then `x-csrf-jwt-header` gets set to the header `onPreResponse`.

What are we doing with `x-csrf-jwt`?

Possible I'm missing some other code elsewhere.

src/util/serverUtils.js
willh-meetup: 😢

src/plugins.js
mmcgahan: `server.state` just sets the default cookie options - the cookies are actually created by `reply.state()` calls for a particular request (i.e. [global cookie config](https://hapijs.com/tutorials/cookies?lang=en_US#configuring-the-server) vs. [setting the cookie](https://hapijs.com/tutorials/cookies?lang=en_US#setting-a-cookie))

- `electrode-csrf-jwt` server plugin calls `reply.state` to set `'x-csrf-jwt'`
- the `onPreResponse` handler will call `reply.state` to set `'x-csrf-jwt-header'`

These two `server.state` configuration calls just make sure that the 'cookie options' are in sync for both cookies.

I'll add docs

PR #201, opened 2017-03-07
support react-helmet v4

mattkime:


package.json
```diff
@@ -58,7 +58,7 @@
   },
   "peerDependencies": {
     "react": "^15.4.2",
-    "react-helmet": "^3.3.2",
+    "react-helmet": ">3.3.2",
```
mmcgahan: any reason not to require `^4.0.0`?

mattkime: its a little more disruptive since it forces consumers to change the version of react-helmet they supply. otherwise their build will break and likely cause a bit of confusion.

mmcgahan:
oh, this will need a `yarn` update before merge

PR #202, opened 2017-03-07
Helmet v4

mattkime:


mmcgahan:
You'll need to move the `git diff-files --quiet` command to `before_script`

From https://docs.travis-ci.com/user/customizing-the-build#Breaking-the-Build

> If any of the commands in the first four stages of the build lifecycle return a non-zero exit code, the build is broken:

> If `before_install`, `install` or `before_script` return a non-zero exit code, the build is errored and stops immediately.

> The exit code of `after_success`, `after_failure`, `after_script` and subsequent stages do not affect the build result. However, if one of these stages times out, the build is marked as a failure.

mattkime:
@mmcgahan before i could even ask the question....thanks!

PR #213, opened 2017-03-13
Yarn Mondays: 3/13/2017

sadafie:
Closes #210 
Closes #204 
Closes #206 
Closes #211
Closes #200 
Closes #205 
Closes #214


mattkime:
if `history` is a problem, lets create a google doc for the problem updates

mmcgahan:
`history` is just being used to work with `baseUrl` in react-router so there is a peer dependency version restriction - best not to update it at all before RRv4 comes through. if only `package.json` could have inline comments to document this stuff!

PR #215, opened 2017-03-14
WP 260 parse Link header

mmcgahan:
The Link header contains all the info we need for paged API requests - just need to parse and store it in the query response object.

### DIFF
--- a/src/util/apiUtils.mockRequest.test.js
+++ b/src/util/apiUtils.mockRequest.test.js
@@ -25,12 +25,15 @@ jest.mock('request', () => {
 						},
 						method: 'get',
 					},
-				}, '{}'), 200
+				}, '{}'), 2
 			)
 	);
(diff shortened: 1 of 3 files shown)


src/util/apiUtils.mockRequest.test.js
```diff
@@ -25,13 +25,16 @@ jest.mock('request', () => {
...
-				}, '{}'), 200
+				}, '{}'), 2
 			)
 	);
 	mock.jar = jest.fn(() => 'myMockJar');
 	return mock;
 });
 
+const API_TIMEOUT = 10;
+const MOCK_HAPI_REQUEST = { server: { app: { API_TIMEOUT } } };
```
mmcgahan: just some cleanup here to make these tests run faster

eilinora:
:hooray: :)

PR #217, opened 2017-03-24
Duotones - Fix values

nlaz:
#### Description 🔵🔴
This PR updates the duotone values because they didn't match the db table. For some reason, they were updated in the database and not in our documentation. The duotone table:
screen shot 2017-03-24 at 6 19 09 pm

Please doublecheck these new values are right. Hex codes are hard to read some times.

Additionally, this PR removes the `group_photo` from being duotone'd. The group photo shouldn't be shown in duotone colors, just the key photo.

### DIFF
--- a/src/util/apiUtils.js
+++ b/src/util/apiUtils.js
@@ -308,7 +308,7 @@ export function parseRequest(request) {
  * @return {Object} the mutated group object
  */
 export const groupDuotoneSetter = duotoneUrls => group => {
-	const photo = group.key_photo || group.group_photo || {};
+	const photo = group.key_photo || {};
 	const duotoneKey = group.photo_gradient && duotoneRef(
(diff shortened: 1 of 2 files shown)


mmcgahan:
excellent - any ideas on how we might be able to keep this up to date automatically?

mmcgahan:
@akdetrick was there an npm package or something that was going to make the duotone values consumable by third parties? I think there was some talk about that at one point.

nlaz:
@mmcgahan We talked about creating an admin tool that documents the colors available in the `duotones` table. Seems like a decent way to reference the current values.

akdetrick:
@mmcgahan We were looking at storing these in the swatches replacement, which we haven't started yet. Header/footer work and component support is taking priority for the next few sprints.

PR #220, opened 2017-03-31
WP 169 consume click cookie

mmcgahan:
This PR updates the server to consume the click-tracking cookie generated by fetchUtils that contains all the click-tracking data in State.


mmcgahan:
the bulk of the changes in the diff are from refactoring the activity-tracking-specific plugin to be more general - separating the logging from the serialization/encoding, which actually makes the whole thing easier to understand and test IMO

src/util/avro.js
```diff
@@ -0,0 +1,72 @@
+const avro = require('avsc');
+
+const click = {
```
willh-meetup: Might be nice to say where we copied this schema from too?

src/util/serverUtils.js
```diff
@@ -21,6 +23,8 @@ export function checkForDevUrl(value) {
 }
 
 export function onRequestExtension(request, reply) {
+	request.id = uuid.v4();
```
willh-meetup: Does this generate a new UUID for this?

src/util/tracking.js
```diff
@@ -172,7 +172,7 @@ export const logTrack = platformAgent => (response, trackInfo) => {
 	const eventDate = new Date();
 	const record = {
 		timestamp: eventDate.getTime().toString(),
-		requestId: uuid.v4(),
+		requestId: response.request.id,
```
willh-meetup: Ah, I see... so this means that both click/activity tracking share a UUID right?

willh-meetup: (I'm assuming that's what we want)

src/util/avro.js
```diff
@@ -0,0 +1,72 @@
+const avro = require('avsc');
```
willh-meetup: 👍 This is a nice refactor.

src/util/tracking.js
mmcgahan: yeah, the Hapi request always had a unique ID, but it wasn't formatted like a UUID, which I think is why we were generating a unique UUID in this step previously. Now, the `onRequestExtension` will overwrite the default `request.id` with a UUID, which can be used consistently for tracking and general logging.

I did some investigation about whether it is 'okay' to override the default `request.id` and everything I saw suggested that it shouldn't be an issue.

PR #221, opened 2017-03-31
WP 275 support query.meta.method

mmcgahan:
This update makes `apiRequest([queries])` more powerful by allowing the query itself to determine the request method. Queries now have a new optional property, `meta`, which can have a `method` key.


docs/Queries.md
```diff
@@ -63,6 +66,15 @@ GET requests) or the request body (for POST requests).
 An array of feature flag (Runtime Flag) names that should be returned
 alongside the main request.
 
+### `meta`
+
+#### `method`
+
+You can force the query to be sent with a particular HTTP method by specifying
+it here as a _lowercase_ string: `get`, `post`, `delete`, or `patch`. Note that
```
sadafie: Nitpick, but could we just make this case agnostic?

src/util/fetchUtils.js
```diff
@@ -27,35 +27,15 @@ export const parseQueryResponse = queries => ({ responses, error, message }) =>
...
- */
-export const fetchQueries = (apiUrl, options) => (queries, meta) => {
-	if (
-		typeof window === 'undefined' &&  // not in browser
-		typeof test === 'undefined'  // not in testing env
-	) {
-		throw new Error('fetchQueries was called on server - cannot continue');
-	}
-	options.method = options.method || 'GET';
+export const getFetchArgs = (apiUrl, options, queries, meta) => {
```
sadafie: Add some jsdoc back to this :)

src/util/validation.js
```diff
@@ -7,5 +7,8 @@ export const querySchema = Joi.object({
 	mockResponse: Joi.object(),
 	params: Joi.object(),
 	type: Joi.string(),
+	meta: Joi.object({
+		method: Joi.string().only('get', 'post', 'delete', 'patch'),
```
sadafie: `Joi.string().lowercase().only('get', ...)` to support any case seems nice to me. i could imagine getting an error and not realizing it's because my method is supposed to be lowercase.

docs/Queries.md
sadafie: Also, would be nice to show an example get and post request

docs/Queries.md
```diff
@@ -63,6 +66,15 @@ GET requests) or the request body (for POST requests).
...
+you should never try to make a request that contains multiple queries with
+different `method`s.
```
sadafie: I probably agree -- but what if I'm doing a `post` of some data, and then i want to retrieve a new object via `GET`? Or some similar workflow.

src/util/fetchUtils.js
```diff
@@ -27,35 +27,15 @@ export const parseQueryResponse = queries => ({ responses, error, message }) =>
...
 	const {
-		method,
 		headers={},
 	} = options;
 
+	const method = (queries[0].meta || {}).method ||  // allow query to set method
```
sadafie: There's a note that you shouldn't mix and match queries with different methods: https://github.com/meetup/meetup-web-platform/pull/221/files#diff-84ac740a7d37c0a9aefc0c3fd48e01afR75

docs/Queries.md
mmcgahan: those would have to be synchronous anyway - you wouldn't want to send the GET before knowing whether the POST was successful. Teams could talk to API engineers to optionally return the created/updated object for a successful POST.

mmcgahan: Also, you can always generate multiple `API_REQUEST` actions if you _really_ want to do simultaneous requests with different methods - they won't be batched, but at most you would be doing 1-request-per-method from the browser

src/util/fetchUtils.js
mmcgahan: I can add some validation for this so that it doesn't surprise anyone during dev.

docs/Queries.md
mmcgahan: It's annoying that the casing on HTTP methods is so inconsistent in different environments, but I guess you're right that that ship has sailed and I shouldn't try to enforce one or the other here.

src/util/validation.js
mmcgahan: Looks like I would use [`.insensitive()`](https://github.com/hapijs/joi/blob/v10.4.1/API.md#stringinsensitive) but yeah

src/util/fetchUtils.js
```diff
@@ -27,35 +27,15 @@ export const parseQueryResponse = queries => ({ responses, error, message }) =>
...
+		options.method.toLowerCase() ||  // fallback to options
+		'get';  // fallback to 'get'
```
nlaz: Would it be weird to make constants for the HTTP methods? Inside `mup-web` we could just have something like:
```js
import { POST } from 'meetup-web-platform/lib/api';
// later..
meta: {
   method: POST,
},
```

mmcgahan: I think it would be a little weird just because those terms are used in lots of different contexts and the should be well-known - using a constant to enforce string case seems awkward, so I'll just coerce-lowercase whatever is supplied and be done with it.

mmcgahan:
I believe we are good to go @sadafie

src/util/fetchUtils.js
nlaz: My argument is that whenever you are counting on strings in two places to match, it makes the code a bit less stable. If you used constants here, wouldn't it turn the potential runtime errors into compile time errors?

mmcgahan: In this case, no, because `import { FOO } from 'bar/baz'` doesn't throw a compile time error when `FOO` doesn't exist but the `'bar/baz'` module _does_ exist, meaning that

```js
import { POSTX } from 'meetup-web-platform/lib/api';

{ method: POSTX }
```

would give you an error at the same time `{ method: 'postx' }` would - it would happen when you first try to send that query to the server at runtime, and the server would tell you that `method` must be one of `'get', 'post', 'delete', 'patch'`.

I _think_ this might be a place where we could set up flow to do compile-time type checking of the arguments to `fetchQueries`, which would probably be better all around without `import` boilerplate.

nlaz: hmm alright

mmcgahan:
@sadafie I'm going to go ahead and merge this since I think all the feedback has been addressed - if there's anything outstanding that I should follow up on, just let me know and I'll get it into a new PR

PR #222, opened 2017-04-01
WP 274 multipart form data

mmcgahan:
I need to figure out the best way to test this with an actual POST to a file-upload API endpoint, but I am pretty sure it's sending the data the way it is supposed to.

Supporting multipart/form-data requires updates in two places

1. in `fetchQueries`, which needs to ensure that the _body_ of the fetch request matches exactly what should be sent to the API - the encoded `FormData` object from the browser. The current code checks whether `query.params` is a `FormData` object, and will treat the query as a multipart/form-data request. This special kind of 'POST' will contain the query information in the querystring params (excluding the `FormData`), but send the `FormData` in the body of the request. The content-type header is set automatically by `fetch`

2. in `apiUtils`, multipart requests are parse differently than normal POSTs because the body of the request has to be passed along in the `formData` field of the `requestOpts`, and the query information has to be read from the querystring. The content-type header is set automatically by the `request` library when `formData` is supplied

I'd like to make all of these elements more clear, but there's a fair amount of piece-by-piece manipulation of the data as it moves through the API proxy.

I'll look to optimize and add tests on Monday.


nlaz:
> I need to figure out the best way to test this with an actual POST to a file-upload API endpoint, but I am pretty sure it's sending the data the way it is supposed to.

Happy to help with that. I can try to verify on mup-web.

src/apiProxy/apiProxyRoutes.js
```diff
@@ -32,7 +32,7 @@ const getApiProxyRoutes = (path, apiProxyFn$) => {
 	};
 	const apiGetRoute = {
 		...routeBase,
-		method: ['GET', 'DELETE', 'PATCH'],
+		method: ['GET', 'DELETE',],
```
nlaz: Extra comma here 🔪

nlaz:
Curious: Why is this blocked by https://github.com/meetup/meetup-web-platform/pull/220?

src/util/apiUtils.js
```diff
@@ -295,17 +303,21 @@ export function parseRequestQueries(request) {
...
+		method: request.method,
+		headers: parseRequestHeaders(request),  // make a copy to be immutable
+		mode: 'no-cors',
+		time: true,  // time the request for logging
+		agentOptions: {
+			rejectUnauthorized: baseUrl.indexOf('.dev') === -1
 		},
+	};
+	if (request.mime === 'multipart/form-data') {
+		externalRequestOpts.formData = request.payload;
```
nlaz: Where does `request.payload` come from? That's not the same as the dispatch action payload right?

src/util/apiUtils.js
```diff
@@ -270,10 +275,13 @@ export function parseRequestHeaders(request) {
 export function parseRequestQueries(request) {
 	const {
 		method,
+		mime,
```
nlaz: Where is `mime` coming from? Would new actions look something like this:

```js
export const exampleAction = (params) => ({
   type: 'EXAMPLE_ACTION',
   payload: {
      mime: 'multipart/form-data',
      query: {
         type: 'EXAMPLE',
         endpoint: `/members/${params.member_id}/photos`,
         params: params,
         formData: params.formData,
         ref: EXAMPLE_REF,
     },
     onSuccess: onSuccessCallback,
     onError: onErrorCallback,
  }
});
```

mmcgahan: For better or worse, you have to forget everything you know about Redux when you're working in server-side request handling code. The `request` here is a [Hapi Request Object](https://hapijs.com/api#request-object), which has a `mime` property derived from the `Content-Type` header of the request - it's not populated manually.

src/util/apiUtils.js
mmcgahan: This is also the Hapi request object - `request.payload` is essentially the Body of the HTML request

mmcgahan:
> Why is this blocked by #220? They seem unrelated.

Conflicts in `fetchUtils`

src/util/apiUtils.js
nlaz: So I guess my real question is how are we going to set the content-type in `mup-web`?

mmcgahan: easy: don't. 🙂 

the query system is intended to make you think about REST as little as possible - from the PR description:

> The current code checks whether `query.params` is a `FormData` object, and will treat the query as a multipart/form-data request.

I'll add more docs for this, but you should never have reason to manually managing content-type.

nlaz: Ah ok. I think things will get tricky. but i think examples will definitely help.

mmcgahan: Done https://github.com/meetup/meetup-web-platform/blob/4a328a45cd0845faca87035eb6cb69bc28df1bb5/docs/Queries.md#recipes

docs/Queries.md
```diff
@@ -221,3 +221,163 @@ If the response does _not_ trigger an `API_SUCCESS` action, you will have to
 write a custom reducer to parse the response data and apply the results to a
 different part of `state`. Otherwise, they will be ignored.
 
+## Recipes
```
sadafie: This is great, thank you for adding!

src/util/apiUtils.js
```diff
@@ -185,32 +187,44 @@ export const parseApiResponse = requestUrl => ([response, body]) => {
...
-		const dataParams = querystring.stringify(params);
-
-		switch (externalRequestOptsQuery.method) {
-		case 'get':
-		case 'delete':
-			externalRequestOptsQuery.url += `?${dataParams}`;
-			externalRequestOptsQuery.headers['X-Meta-Photo-Host'] = 'secure';
-			break;
+		switch (externalRequestOpts.method) {
 		case 'post':
```
sadafie: add a `patch` case that falls through here?

src/util/apiUtils.js
```diff
@@ -185,32 +187,44 @@ export const parseApiResponse = requestUrl => ([response, body]) => {
...
-		externalRequestOptsQuery.jar = createCookieJar(externalRequestOptsQuery.url);
+		const dataParams = querystring.stringify(params);
+		const headers = { ...externalRequestOpts.headers };
+		let url = encodeURI(`/${endpoint}`);
+		let body;
+		const jar = createCookieJar(url);
 
 		if (flags) {
-			externalRequestOptsQuery.headers['X-Meetup-Request-Flags'] = flags.join(',');
+			headers['X-Meetup-Request-Flags'] = flags.join(',');
```
sadafie: What happens if we always set this header? Will it break? Maybe something like:

```
headers['X-Meetup-Request-Flags'] = (flags || []).join(',')
```

In general I'm trying to see if we can move in the direction of not branching logic for which headers get set, but see if it's possible to send a value that is ignored, or see if the API just ignores certain headers depending on the method.

src/util/apiUtils.js
```diff
@@ -185,32 +187,44 @@ export const parseApiResponse = requestUrl => ([response, body]) => {
...
+			}
+			body = dataParams;
+			headers['content-type'] = 'application/x-www-form-urlencoded';
 			break;
+		case 'delete':
+		case 'get':
+		default:
+			url += `?${dataParams}`;
+			headers['content-type'] = 'application/json';
+			headers['X-Meta-Photo-Host'] = 'secure';
```
sadafie: Same question here, could the `X-Meta-Photo-Host` header be set for all method types?

src/util/apiUtils.js
sadafie: Once again, love the examples!

src/util/fetchUtils.js
```diff
@@ -78,23 +79,31 @@ export const getFetchArgs = (apiUrl, options, queries, meta) => {
...
 		// send other metadata in searchParams
-		fetchUrl.searchParams.append('metadata', rison.encode_object(metadata));
+		if (metadata) {
+			fetchUrl.searchParams.append('metadata', rison.encode_object(metadata));
+		}
+	}
 
+	if (!isFormData) {
+		// need to manually specify content-type for any non-multipart request
+		headers['content-type'] = isPost && 'application/x-www-form-urlencoded' ||
```
sadafie: Sorry, could you explain why we're setting the content-type again here as well?

src/util/apiUtils.js
mmcgahan: that can be added in the PR that supports `PATCH`: https://github.com/meetup/meetup-web-platform/pull/223

src/util/apiUtils.js
mmcgahan: I'm not sure - I agree with the desire to avoid branching logic, but it also seems wrong to send unnecessary headers, particularly something like this that is framed as a 'request' - it would be like asking for nothing rather than not asking at all, and it essentially shifts the `if` from the client code to the server code - the REST API would have to do the null check on their end to avoid making an empty DB query.

I think I'm happier to keep the client request lean, even at the cost of an `if`

src/util/apiUtils.js
mmcgahan: technically it should only be needed for `GET` because all other methods are mutations that shouldn't return data. Do you think it should be added to all requests?

src/util/fetchUtils.js
mmcgahan: the content-type is set here in the browser for requests to the app server, but is set in `apiUtils` for requests to the API server.

It might be possible to avoid setting the `content-type` in `apiUtils`, but it can be a pretty fragile setting so I'd like to figure that out in a followup

nlaz:
🎉

PR #223, opened 2017-04-03
WP 273 support PATCH requests

mmcgahan:


nlaz:
So if MWP is supposed to abstract REST for `mup-web`, does `fetchUtils` set the request type anywhere? Where is `PATCH` set versus a `POST` request?

src/util/fetchUtils.js
```diff
@@ -28,36 +28,33 @@ export const parseQueryResponse = queries => ({ responses, error, message }) =>
...
 		headers={},
 	} = options;
 
-	const isPost = method.toLowerCase() === 'post';
-	const isDelete = method.toLowerCase() === 'delete';
+	const method = ((queries[0].meta || {}).method || '').toLowerCase() ||  // allow query to set method
+		options.method.toLowerCase() ||  // fallback to options
+		'get';  // fallback to 'get'
+
+	const hasBody = method === 'post' ||
```
nlaz: Unfortunately, we will also have to support `DELETE` requests with bodies. For example: https://www.meetup.com/meetup_api/docs/:urlname/events/:id/#delete

Maybe we can add that support in the same PR?

mmcgahan:
`fetchUtils` assigns the method based on the query [here](https://github.com/meetup/meetup-web-platform/blob/0485935b5a39ca3f589b0cdb8920fdea90776464/src/util/fetchUtils.js#L50), but falls back to an optional `method` specified in a passed-in config option (this is essentially deprecated, but is currently used by the POST and DELETE middlewares, which we'll be phasing out soon), and ultimately falls back to `'get'` if nothing else is specified

nlaz:
@mmcgahan I appreciate the clarification. Thanks!

mmcgahan:
reminder https://github.com/meetup/meetup-web-platform/pull/222/files/2d98479ad3d41b6648a23f1c473cab3afba8ff38#r110182396

mmcgahan:
Ready for final review @nlaz

src/util/fetchUtils.js
```diff
@@ -80,11 +81,11 @@ export const getFetchArgs = (apiUrl, queries, meta) => {
 
 	if (!isFormData) {
 		// need to manually specify content-type for any non-multipart request
-		headers['content-type'] = isPost && 'application/x-www-form-urlencoded' ||
+		headers['content-type'] = hasBody && 'application/x-www-form-urlencoded' ||
```
nlaz: This part has got me confused. Is this a different type of ternary?

mmcgahan: effectively, yes. since it's confusing I'll change it to a regular ternary, though.

inJS, `&&` has higher operator precedence than `||`, so the first `&&` grouping is treated as a pair, and the value will fallback to the `|| 'application/json'` value if `hasBody` is falsey

nlaz: Thanks! 👍

PR #226, opened 2017-04-05
always call apiSuccess or apiError for posts and deletes

mmcgahan:
`onSuccess` and `onError` handlers for _POST and _DELETE actions are now truly optional - redundant `apiSuccess`/`apiError` calls can be avoided


eilinora:
do we need to update doc's around this? https://github.com/meetup/meetup-web-platform#post

PR #229, opened 2017-04-07
timestamp formatting and shorter platform name

mmcgahan:
Talked with Tom and Sach about some issues that appeared when the tracking data was actually being processed - the timestamp needs to be an ISOString and the 'platform' tag needs to be shorter in order to work with the tables where it's stored


src/util/clickTracking.js
```diff
@@ -134,6 +134,7 @@ function getTrackClick(store) {
 
 		// 2. Create click action with metadata
 		const clickTrackAction = click({
+			timestamp: new Date().toISOString(),
```
willh-meetup: part of me thinks it's nice to abstract this into a `getISOString()` function since we're doing it more than once... but, it's also only one line.

Might be nice to make them consistent, though, and either set an eventDate variable or not (I also like the syntax with an extra set of parens like `(new Date()).toISOString()` )

mmcgahan: sounds like you're a closet Lisper.

I don't think I'd abstract into a function because the 'native' implementation is explicit, but I'll set a var in both places.

PR #230, opened 2017-04-07
tag without url

mattkime:
if tagging fails it dumps our creds into publicly available logs


mattkime:
need to verify that tagging is working

mattkime:
This alone did not work. Need to follow the directions here first - https://github.com/alrra/travis-scripts/blob/master/doc/github-deploy-keys.md

mmcgahan:
Let's get some comments on this code soon so that the next dev doesn't have to work as hard to figure it out.

PR #239, opened 2017-04-11
use meetup fork of electrode csrf jwt

mmcgahan:
the CSRF plugin has [this issue](https://github.com/electrode-io/electrode-csrf-jwt/issues/11), which is preventing retries of failed uploads.

I've submitted [this PR](https://github.com/electrode-io/electrode-csrf-jwt/pull/12) to fix it upstream, but their Travis test runner isn't configured correctly so it's unclear how long it will take to merge. Until then, we can use the Meetup fork of the package that contains the fixes.


mmcgahan:
I'm going to merge this because every other workaround requires a lot of code and most likely tech debt.

PR #241, opened 2017-04-12
travis - sudo false and comment out compiler

mattkime:


.travis.yml
```diff
@@ -14,17 +14,19 @@ branches:  # exclude tags
...
-      - ubuntu-toolchain-r-test
-    packages:
-      - g++-4.8
+# disabled since we're not using native modules
+# addons:
+#  apt:
+#    sources:
+#      - ubuntu-toolchain-r-test
+#    packages:
+#      - g++-4.8
```
mmcgahan: go ahead and delete instead of commenting - the compiler is only used when a repo depends on sass-tools

PR #242, opened 2017-04-12
consolidate post & delete epics

mmcgahan:
See https://meetup.atlassian.net/browse/WP-296

Basically just consolidating logic to make some upcoming deprecations & refactors simpler

There's not a lot to review here - essentially no logic changes and the same unit tests are all running


src/epics/mutate.js
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Make the mutation call to the API and send the responses to the appropriate
+ * places
+ *
+ * 1. Always send successful mutation responses to API_SUCCESS action
+ * 2. If mutation action has an 'onSuccess' action creators, send successful
+ *    responses there as well
+ * 3. Failed mutation responses will be sent to the mutation action's `onError`
+ */
+const doFetch$ = fetchQuery => ({ query, onSuccess, onError }) =>
```
mmcgahan: this function is identical to the previous `doDelete$` and `doPost$` functions.

src/epics/mutate.js
```diff
@@ -0,0 +1,91 @@
...
+		})
+		.catch(err => {
+			const actions = [apiError(err)];
+			if (onError) {
+				actions.push(onError(err));
+			}
+			return Rx.Observable.from(actions);
+		});
+
+const getMethodEpic = method => fetchQueries => (action$, store) =>
```
mmcgahan: This function is the same as the previous `getDeleteEpic` and `getPostEpic` functions, but it takes a `method` argument in order to make a method-specific epic.

src/epics/mutate.js
```diff
@@ -0,0 +1,91 @@
...
+	action$.filter(({ type }) =>
+		type.endsWith(`_${method.toUpperCase()}`) || type.startsWith(`${method.toUpperCase()}_`))
+		.map(({ payload }) => payload)
+		.flatMap(
+			doFetch$(
+				getMethodQueryFetch(method, fetchQueries, store)
+			)
+		);
+
+export const getPostEpic = getMethodEpic('post');
```
eilinora: how does PATCH fit in, i know this i just a migration of current code but i _believe_ we added patch in or do we not still support PATCH? I know we have the forced `method: 'patch'` but does that go down a different path than this?

src/epics/mutate.js
```diff
@@ -0,0 +1,91 @@
+import Rx from 'rxjs';
```
eilinora: obnoxious question alert, why mutate?

src/epics/mutate.js
eilinora: I think i might know the answer to my own question but is this the alternate code path?
https://github.com/meetup/meetup-web-platform/blob/94b1f2951a0b57dad5830a2a5686707b4adb647c/src/epics/sync.js#L1

src/epics/mutate.js
mmcgahan: mainly because the more I've learned about systems solving similar problems, the more I've come across that terminology - Relay, Falcor, and 'JSON-pure' all ignore POST/PATCH/DELETE distinctions and instead refer to them as 'mutations' - they are the HTTP requests that _change_ data, unlike GETs, which in theory should always be idempotent.

eilinora: gotcha

PR #243, opened 2017-04-13
Wp 282 deprecate POST and DELETE middleware

mmcgahan:
See https://meetup.atlassian.net/browse/WP-282

These epics can be removed now that we have a simpler alternative pattern for creating POST, DELETE, and PATCH requests from queries. To avoid a breaking change, this PR just adds deprecation notices

### DIFF
--- a/src/epics/mutate.js
+++ b/src/epics/mutate.js
@@ -80,7 +80,7 @@ const doFetch$ = fetchQuery => ({ query, onSuccess, onError }) =>
 const getMethodEpic = method => fetchQueries => (action$, store) =>
 	action$.filter(({ type }) =>
 		type.endsWith(`_${method.toUpperCase()}`) || type.startsWith(`${method.toUpperCase()}_`))
-		.do(({ type, query: { endpoint } }) => {
+		.do(({ type, payload: { query: { endpoint } } }) => {
 			if (endpoint.indexOf('.dev.') > -1) {
 				// using a dev endpoint, render a deprecation warning
 				console.warn(`This application is using Post/Delete middleware through ${type}.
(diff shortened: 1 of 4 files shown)


src/epics/mutate.js
```diff
@@ -0,0 +1,100 @@
...
+				actions.push(onError(err));
+			}
+			return Rx.Observable.from(actions);
+		});
+
+const getMethodEpic = method => fetchQueries => (action$, store) =>
+	action$.filter(({ type }) =>
+		type.endsWith(`_${method.toUpperCase()}`) || type.startsWith(`${method.toUpperCase()}_`))
+		.do(({ type }) => {
+			if (process && process.pid) {
```
mmcgahan: just realized this will never appear in the server logs because posts and deletes only happen from browser...

src/epics/mutate.js
```diff
@@ -0,0 +1,100 @@
...
+			return Rx.Observable.from(actions);
+		});
+
+const getMethodEpic = method => fetchQueries => (action$, store) =>
+	action$.filter(({ type }) =>
+		type.endsWith(`_${method.toUpperCase()}`) || type.startsWith(`${method.toUpperCase()}_`))
+		.do(({ type, payload: { query: { endpoint } } }) => {
+			if (endpoint.indexOf('.dev.') > -1) {
+				// using a dev endpoint, render a deprecation warning
+				console.warn(`This application is using Post/Delete middleware through ${type}.
```
eilinora: WOOHOO! :D

sadafie:
I like this deprecation approach! Is there a strategy to timebox this on the consumer end for complete removal?

mmcgahan:
@sadafie I'll add an item to our weekly agenda, but at the moment I think giving the product teams a 1-2 sprint heads up about removal would be good. We can/should recommend that the teams roll all the '3.0' upgrades into a single sprint - if this deprecation and the API_SUCCESS deprecation can be resolved in one sprint, we'll be in good shape to make a major version update that drops support for the old systems.

PR #247, opened 2017-04-14
master-only builds

mmcgahan:
as discussed in our weekly, we're going to limit our travis builds and see whether we miss the extra push builds


mmcgahan:
Actually I'm going to self-verify

PR #248, opened 2017-04-14
WP 301 fix member id track

mmcgahan:
The member id tracking was reliant on a separate MEMBER_ID cookie that wasn't behaving correctly - it's something that we might want to come back to when we fully develop a login system, but probably best not to hack together support for a hypothetical use case

### DIFF
--- a/src/util/cookieUtils.js
+++ b/src/util/cookieUtils.js
@@ -1,0 +1,8 @@
+import querystring from 'qs';
+
+const isProd = process.env.NODE_ENV === 'production';
+export const MEMBER_COOKIE = isProd ? 'MEETUP_MEMBER' : 'MEETUP_MEMBER_DEV';
+
+export const parseMemberCookie = state =>
+	querystring.parse(state[MEMBER_COOKIE]);
+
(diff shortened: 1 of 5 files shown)


src/util/cookieUtils.js
```diff
@@ -0,0 +1,8 @@
+import querystring from 'qs';
+
+const isProd = process.env.NODE_ENV === 'production';
+export const MEMBER_COOKIE = isProd ? 'MEETUP_MEMBER' : 'MEETUP_MEMBER_DEV';
+
+export const parseMemberCookie = state =>
```
willh-meetup: Wondering if it makes sense to handle setting `id` to 0 and parseInt here, instead of a bunch of times throughout the code.

Also, then you can write some tests for this! (particularly I'm think about the cookie doesn't exist case, which will totally be a thing at some point soon).

mmcgahan: definitely. I was wondering about the parseInt + fallback, but I was a little cautious about assuming the return value should be an `int` in all cases... but it makes sense to treat that as the default, and it can be re-stringified if need be

PR #249, opened 2017-04-14
publish src

mattkime:
testing, for now.


mmcgahan:
seems like a fine change to me!

mattkime:
mergin'...tree shaking is on its way...

PR #252, opened 2017-04-18
WP-293: Node Convict

sadafie:
## What's changed

- There is a central `./util/config.js` that exports a [node-convict configuration object](https://github.com/mozilla/node-convict#the-schema). The default values in this configuration file are production values, and empty strings for secrets.

- We no longer need `~/.mupweb.config`. There are corresponding `config.development.json`, `config.production.json`, and `config.test.json` files at the root of the repo. The first two are gitignored because they contain secrets. The last one (`config.test.json`) is used only in the context of jest unit and integration tests, and so it is checked into the repo. All the secrets and keys are dummy values. Inside `.util/config.js`, after the default values are loaded, the configs in the json file are merged in.

-  All references to `process.env` now use use `config.get('...')`

-  Introduced `isDev` and `isProd` configuration variables in the node-convict config

- `./util/config.js` used to be Promise based. However, the best practice with node-convict to to export a module, so we are no longer using a promise based system, with the config object passed around in the code. Instead, any file that needs access to config simply imports it and calls `config.get`. This change is the main reason the diff is so large.

## TODO

- Fix failing tests
- Write new unit test cases for config.js
- Update docs with node-convict info + change .mupweb.config steps to be config.development.json
- Update mup-web and other consumer repos to use the config module instead of `proccess.env`

## Questions
- @mmcgahan the configuration is loaded to state at some point, and read from state in several of the epics / middleware. (i.e., mutate.js, sync.js, etc.). Could you remind me why the configuration is being loaded through state in these cases?
- @mmcgahan Now that we have `config.get()` and centralized config management, is it necessary to also use [Hapi serverapp](https://hapijs.com/api#serverapp)? Specifically referencing https://github.com/meetup/meetup-web-platform/pull/252/files#diff-145dda275fde2198942fed5c08f89307R129 in `serverUtils.js`.
- It's not super clear to me yet how to share configurations across the platform and consumer repos. I'm assuming this is why the current `~/.mupweb.config` file lives under the home directory, so there's a chance that I will need to remove the `config.<env>.json` files from this repo and have them in the root directory as well. Unless we have access to the Hapi server in the consumer repo, in which case this probably ties back to the previous question, and we could just use `serverapp` instead of `process.env..` in mup-web.


src/util/config.js
```diff
@@ -1,83 +1,153 @@
...
-		OAUTH_ACCESS_URL: process.env.OAUTH_ACCESS_URL ||
-			process.env.ANONYMOUS_ACCESS_URL ||
-			'https://secure.dev.meetup.com/oauth2/access',
-		PHOTO_SCALER_SALT: process.env.PHOTO_SCALER_SALT,
-		oauth: {
-			secret: process.env.MUPWEB_OAUTH_SECRET,
-			key: process.env.MUPWEB_OAUTH_KEY,
+const oauthError = new Error('get oauth secrets from #web-platform team');
+
+let config = convict({
```
sadafie: I'm considering naming this `envConfig` so it doesn't get confused w/ every other configuration in the server code (there are a few). Thoughts?

config.test.json
```diff
@@ -0,0 +1,15 @@
+{
```
sadafie: hmm, weird highlighting in GH. Is this an issue with my (our) editorconfig?

src/util/config.js
swozniak: This file or this variable?

config.test.json
swozniak: If it's .json then Github probably expects all keys and string values to be surrounded in double quotes

src/util/authUtils.js
```diff
@@ -74,14 +76,14 @@ export function validateSecret(secret) {
...
-	server.app.isDevConfig ? 'MEETUP_MEMBER_DEV' : 'MEETUP_MEMBER';
+	server.app.isDev ? 'MEETUP_MEMBER_DEV' : 'MEETUP_MEMBER';
 
 /**
  * apply default cookie options for auth-related cookies
  */
 export const configureAuthCookies = server => {
-	const password = validateSecret(server.plugins.requestAuth.config.COOKIE_ENCRYPT_SECRET);
-	const isSecure = process.env.NODE_ENV === 'production';
+	const password = validateSecret(config.get('cookie_encrypt_secret'));
```
swozniak: Not sure how this works - is `server.plugins.requestAuth.config` as referenced on line 83 the same as this `config` variable (imported from the config)?

src/util/config.js
```diff
@@ -1,83 +1,146 @@
...
+);
+
+config.set(
+	'isDev',
+	config.get('env') === 'development'
+);
+
+config.validate();
 
+export default config;
```
swozniak: Opening this up for discussion.
It's much more straightforward to return the config object from node-convict than to resolve the JSON output from that object, but it does make for more to track down in the future if we stop using node-convict (though nothing insurmountable).

Also, this used to return an immediately resolved promise with the object literal of config values - wasn't sure what the impetus was behind that. @mmcgahan ?

src/plugins/requestAuthPlugin.js
```diff
@@ -244,13 +250,12 @@ const refreshToken$ = refresh_token => Rx.Observable.of({
  * For an anonymous auth, the request header information is used to determine
  * the location and language of the anonymous member
  *
- * @param {Object} config { OAUTH_AUTH_URL, OAUTH_ACCESS_URL, oauth }
  * @param {Object} request the Hapi request that needs to be authorized
  */
-export const getRequestAuthorizer$ = config => {
+export const getRequestAuthorizer$ = () => {
```
mmcgahan: I'd prefer to keep existing pure functions pure, i.e. making sure that we pass in the data they need as arguments rather than giving them having an implicit reliance on scope provided by an `import`. The downside is that we have to thread configuration down to various functions, but that's far better IMO than a `config` import that can lead to spaghetti code with implicit dependencies.

The original code wasn't as specific as it could/should be (e.g. passing the whole config object when functions only needed a subset of properties), but I'd lean toward fixing that weakness rather than doubling down :)

src/util/authUtils.js
mmcgahan: use `server.app` to access the config - skip the import

src/util/authUtils.test.js
```diff
@@ -1,4 +1,5 @@
 import Joi from 'joi';
+import config from './config';
```
mmcgahan: probably not great to have unit tests depend on environment config - we should always mock the config in tests

src/util/clickTrackingReader.js
```diff
@@ -1,11 +1,11 @@
-import { parseMemberCookie } from './cookieUtils';
+import config from './config';
 
-const isProd = process.env.NODE_ENV === 'production';
+import { parseMemberCookie } from './cookieUtils';
 
 export const clickCookieOptions = {
```
mmcgahan: let's make this a function that can be passed a `server` instance, so that the `isProd` can be read from `server.app` rather than an import. `clickCookieOptions` is only called from places where we have access to `request.server`, so it should be fine.

src/util/config.js
mmcgahan: it's fine to roll back the Promise setup - it was a vestige of a time where the server had to make some HTTP requests to get some config data when booting, so the config object had to be built async. I think we can avoid that requirement in the future, but I had originally kept the `Promise` just in case we needed it.

Treating the config object as a Plain Old Javascript Object throughout the rest of the code would definitely make it easier to understand - the getters and setters feel a little awkward to me, although I'm sure 90% of that is just not being used to node convict's API. I can't think of a good reason not to export the POJO version, though.

src/util/globals.js
```diff
@@ -1,12 +1,14 @@
...
 global.URL = require('url').URL;
 global.fetch = fetch;
 
 // runtime values needed by browser and server
 global.CONFIG = {
-	ASSET_SERVER_HOST : process.env.ASSET_SERVER_HOST,
-	ASSET_SERVER_PORT : process.env.ASSET_SERVER_PORT,
+	ASSET_SERVER_HOST : config.get('asset_server.host'),
+	ASSET_SERVER_PORT : config.get('asset_server.port'),
 };
```
mmcgahan: pretty sure this `global` object is dead now - I've been meaning to delete it for aaaages.

src/util/serverUtils.js
```diff
@@ -128,31 +119,16 @@ export function registerExtensionEvents(server) {
...
+export function server(routes, connection, plugins, platform_agent) {
 	const server = new Hapi.Server();
 
 	// store runtime state
 	// https://hapijs.com/api#serverapp
-	server.app = {
-		isDevConfig: checkForDevUrl(config),  // indicates dev API or prod API
-		...config
-	};
+	// TODO: do we still need this?
```
mmcgahan: Yes, and in general we should try to exclusively use the `server.app` object to refer to server configuration - it's a [Hapi convention](https://hapijs.com/api#serverapp) that reduces the need for importing config ad-hoc. I like the `server.app` property because it forces "server configuration" to _be part of the server_ object. Config that is _not_ related to the server can go elsewhere, but we might not have anything else.

I've noted a few places where `server.app` can/should be used in this PR instead of importing `config`, so hopefully that can reduce the need for ad-hoc config imports, which can make the dependency structure harder to see and maintain. It might take a few iterations to fully clean it up (my ideal is that we import `config` exactly once - when the server is first configured - but we'll see how close we can get)

tests/integration/requestAuthPlugin.int.js
```diff
@@ -1,5 +1,7 @@
 import Hapi from 'hapi';
 import Iron from 'iron';
+
+import config from '../../src/util/config';
```
mmcgahan: no environment dependencies in unit tests

sadafie:
@mmcgahan @swozniak -- I tried to roll back some of the commits here to limit the scope of the changes as we talked about in last week's meeting, however, between the rollback and the merge from master, I think I'v managed to get this branch into a wonky state. I'm going to close this PR and create a new branch for this. Sorry!

PR #253, opened 2017-04-19
WP 277 better logger

mmcgahan:
This is a full refactor of all console logging on the server, including tracking output. The main motivation is to make dev logs easier to read while maintaining the full JSON-encoded logs for prod.

Although the changes are relatively extensive, the functional parts of the application haven't been touched, only the stdout logging. Also, there is a net reduction in LOC because the 'good-console' plugins that were being used for logging and tracking have been removed.

The interface into logging is provided by Pino and Hapi-Pino, which inject some properties onto the `server` and `request` objects that hook into a shared 'logger' instance that is used throughout the app. Where possible, I've accessed the logger through `server.logger`, although in places where it is not available I've imported the logger instance directly from the util/logger module.

**Before**

screen shot 2017-04-19 at 2 07 18 pm

**After**

screen shot 2017-04-19 at 2 07 54 pm


src/plugins/requestAuthPlugin.js
```diff
@@ -261,28 +263,22 @@ export const getRequestAuthorizer$ = config => {
...
-	}));
+	const { server: { app: { logger } }, raw: { req } } = request;
+	logger.debug({ req }, 'Authenticating request');
 	return authorizeRequest$(request)
 		.do(request => {
-			console.log(JSON.stringify({
-				message: 'Request authenticated',
-				info: { id }
-			}));
+			logger.debug({ req }, 'Request authenticated');
```
sadafie: Do we have log levels configured somewhere?

src/renderers/server-render.jsx
```diff
@@ -81,43 +81,33 @@ const getRouterRenderer = (
...
-			200;
-
-	} catch(error) {
-		// log the error stack here in dev to make it a little more legible
-		// - prod will get stackdriver formatting
-		if (process.env.NODE_ENV !== 'production') {
-			console.error(error.stack);
-		}
-		throw error;
+	appMarkup = ReactDOMServer.renderToString(
```
sadafie: I'm assuming the try/catch here is safe to remove?

src/renderers/server-render.jsx
```diff
@@ -190,19 +182,15 @@ const makeRenderer = (
...
-			id: request.id,
-		}
-	}));
-	store.dispatch({
+	const action = {
 		type: SERVER_RENDER,
 		payload: url,
-	});
+	};
+	server.app.logger.debug(
```
sadafie: What's the advantage of using `server.app.logger` instead of just the logger module directly? Is it just one fewer import? The reason I ask is because I think it will be unclear for others working in this code the difference (none) and/or when server.app is available and when it's not.

sadafie: A follow up question / suggestion -- is there a way to make `server` available in most places without complicating things, so that 99% of the time we look for modules on `server.app`? (I'm thinking in the context of configs as well as logging)

src/util/apiUtils.js
```diff
@@ -521,30 +521,32 @@ export const logApiResponse = request => ([response, body]) => {
...
-			}, {}),
-			method,
-			id,
-			originRequestId: request.id,
-			statusCode: statusCode,
-			time: elapsedTime,
-			body: body.length > 256 ? `${body.substr(0, 256)}...`: body,
-		}
-	}));
+	const log = (statusCode >= 400 && logger.error ||
```
sadafie: Should we include `logger.fatal` and `logger.debug` in this logic?

src/util/logger.js
```diff
@@ -0,0 +1,22 @@
+import pino from 'pino';
+
+
+const isDev = process.env.NODE_ENV !== 'production';
```
sadafie: Such a nitpick, but to me `test` environment is different than `development` environment, so can we have a `isProd` flag instead of an `isDev` here?

src/util/logger.js
```diff
@@ -0,0 +1,22 @@
...
+const isTest = process.env.NODE_ENV === 'test';
+
+const reqSerializerDev = req => `${req.method.toUpperCase()} ${req.url.href}`;
+const resSerializerDev = res => res.statusCode;
+
+const logger = pino({
+	prettyPrint: isDev,
+	timestamp: isDev,
+	messageKey: 'message',
+	enabled: !isTest,
```
sadafie: I'm assuming this means that during the tests, if  you run jest with `-- verbose` you won't see any of the logging. If that's the case, sometimes I do use logging in methods and tests to debug a failing test or while writing new ones. Not sure if anyone else does, but i might be nice to keep it enabled all the time since we default run jest in `-- silent`

src/plugins/requestAuthPlugin.js
mmcgahan: I'll add some docs, but Pino supports standard [Log4j levels](https://en.wikipedia.org/wiki/Log4j#Log4j_log_levels)

src/renderers/server-render.jsx
mmcgahan: the previous `catch` was just a way to log the error in a readable format and then re-throw it, so now that's unnecessary

src/renderers/server-render.jsx
mmcgahan: yeah, this has a similar constraint to importing the `config` instance in different parts of the application, and the resulting setup is similar. We should always prefer to read from `server.app.logger` because it is by definition the server-side logging instance and there's no need for an extra import when you have access to `server`.

The need to import `logger` directly might be an indication that a module could/should be refactored to be better integrated into the `server`.

I'll add some documentation, but in practice it won't matter if people import `logger` directly, but using `server.app.logger` is cleaner

src/util/apiUtils.js
mmcgahan: I don't think so - `fatal` is for major 'system shutdown' kinds of errors, and `debug` is just for tracing code execution.

src/util/logger.js
mmcgahan: This is a pretty common idiom in Node code, and I'd argue that everything outside of `'production'` is development of some description, but I take your point. I'll change this to `prettyPrint` so that it's more clear what it's being used for

src/util/logger.js
mmcgahan: yeah I'll have another think about this - the main problem is that while `jest --silent` will silence `console.x` output, it _doesn't) silence this logger because this logger uses `process.stdout` directly. I'll see if there's an easy way to conditionally enable it during tests.

PR #254, opened 2017-04-19
remove intl from browser bundle

mattkime:
Intl is being bundled into our app js even though we're polyfilling it as well. I'm trying to figure out why its being included.


mattkime:
Intl was being included in our browser bundle even though we polyfill it. Removing it from `localizationUtils`, which is used by the browser, fixes the problem.

PR #256, opened 2017-04-20
WP 303 enable flowtype annotations

mmcgahan:
Add flow config to WP - starter implementation in server-render.js and linkify.js

More 'advanced' modules will require more time and effort, but I'm eager to get into that in separate tickets.

the `flow-typed/` directory is where any existing flow type definitions for our dependencies are kept - it helps make sure we call them in the correct way.


flow-typed/platform.js
```diff
@@ -0,0 +1,32 @@
+// @flow
+declare var Intl: Object;
+
+export type FluxStandardAction = {
```
mikespencer: I ran into some funky stuff that I *think* was due to using `export type` as opposed to `declare type`... although at this point I'm not 100% sure if that was the actual issue or not (just in case you run into issues with Flow not being able to find custom defined types that are dependencies of other custom defined types)

package.json
```diff
@@ -10,8 +10,9 @@
   "scripts": {
     "build": "babel src --out-dir lib --ignore test.js,test.jsx --source-maps",
     "coveralls": "cat coverage/lcov.info | coveralls",
+    "flow": "flow",
     "lint": "eslint . --ext .js,.jsx --fix",
-    "test": "yarn run test:unit && yarn run test:integration",
+    "test": "yarn run flow && yarn run test:unit && yarn run test:integration",
```
mikespencer: We ran into some issues in running flow Travis in pro-web when adding new packages to node_modules (something caching related I think), so we had to disable flow in Travis for the time being till we can figure out what the deal with that is. Just a heads up in case you run into a similar issue!

flow-typed/platform.js
mmcgahan: okay cool - yeah I'll change this over to `declare` and make sure it still works as expected - that should keep it aligned with the flow docs a little better as well.

package.json
mmcgahan: will keep an eye out for this, thanks. Since this package is built much differently in Travis than the platform apps (no Docker, using more built-in Travis APIs), it may end up not being an issue for us but I'll let you know.

mattkime:
might we be able to work with `flow-typed` definitions the same way we work with `node_modules`?

mmcgahan:
possibly - what do you mean?

mattkime:
.gitignore the `flow-typed` dir and run `flow-typed install` as you would `yarn install`, although i suspect you'd include them as part of npm publish.

mmcgahan:
ah, that would be worth investigating, but `flow-typed` is a little awkward to work with because it won't run when you have dependencies listed in both `peerDependencies` and `devDependencies` (which is dumb, but...). I had to temporarily disable that check in order to install the current libdefs. `flow-typed` is also not very fast. The manual install isn't great, but there's also not a major problem with the type definitions getting out of date. We'll probably just clean them out and re-update periodically - they don't change very frequently.

So, definitely worth following up with at some point, but pretty low priority IMO

PR #258, opened 2017-04-20
update mocks and fix missing rxjs operators

mmcgahan:
meetup-web-mocks was masking a lot of missing rxjs operators, which allowed tests to pass for modules with missing dependencies.

I'll leave this open for a little bit, but since this is blocking platform updates in consumer apps, I'll merge by EOD

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,13 +55,13 @@
     "uuid": "3.0.1"
   },
   "peerDependencies": {
-    "react": "^15.4.2",
-    "react-helmet": ">=3.3.2",
+    "react": "^15.5.4",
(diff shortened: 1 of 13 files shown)


mattkime:
>meetup-web-mocks was masking a lot of missing rxjs operators

so some tests were lacking data....and therefore passing??

mmcgahan:
`meetup-web-mocks/src/app.js` had `import Rx from 'rxjs'` at the top, which meant that all of the RxJS operators had been populated for any test that imported it - the operators were implicitly in-scope in tests, but not in prod

PR #259, opened 2017-04-21
WP 307 inFlight array of in-flight API request `ref`s

mmcgahan:
After talking with @nlaz about how they would like to use `isFetching`, it looks like it would be much more useful to keep track of all in-flight API requests in an array instead of a 'dumb' boolean, which will allow for much more reliable tracking of API request state.

This PR removes `state.api.isFetching` and replaces it with `state.api.inFlight` - the old `state.app.isFetching` is still in place for legacy support.

Also, flow type annotations for the new platform api reducer.

### DIFF
--- a/src/reducers/platform.js
+++ b/src/reducers/platform.js
@@ -27,7 +27,7 @@ type ApiState = {
 };
 
 export const DEFAULT_API_STATE: ApiState = { inFlight: [] };
-export const DEFAULT_APP_STATE = { isFetching: false };
+export const DEFAULT_APP_STATE = {};
 
 export const responseToState = (response: QueryResponse): { [string]: QueryResponse } =>
 	({ [response.ref]: response });
@@ -80,9 +80,8 @@ export function app(state=DEFAULT_APP_STATE, action={}) {
 		if ((action.meta || {}).logout) {
 			return DEFAULT_APP_STATE;  // clear app state during logout
 		}
-		return { ...state, isFetching: true };
+		return state;
 	case 'API_SUCCESS':
-		state.isFetching = false;  // fall through - everything else is the same as CACHE_SUCCCESS
 		// API_SUCCESS contains an array of responses, but we just need to build a single
 		// object to update state with
 		newState = action.payload.responses.reduce((s, r) => ({ ...s, ...r }), {});
@@ -92,7 +91,6 @@ export function app(state=DEFAULT_APP_STATE, action={}) {
 		return {
 			...state,
 			error: action.payload,
-			isFetching: false,
 		};
 	default:
 		return state;
--- a/src/reducers/platform.js
+++ b/src/reducers/platform.js
@@ -34,10 +34,10 @@ export const responseToState = (response: QueryResponse): { [string]: QueryRespo
 /*
  * The primary reducer for data provided by the API
  */
-export function api(state: ApiState = DEFAULT_APP_STATE, action={}) {
+export function api(state: ApiState = DEFAULT_APP_STATE, action: FluxStandardAction) {
 	switch (action.type) {
 	case API_REQ: {
-		const requestRefs = action.payload.map(({ ref }) => ref);
+		const requestRefs = (action.payload || []).map(({ ref }) => ref);
 		const inFlight = state.inFlight
 			.filter(ref => !requestRefs.includes[ref]) // clean out current duplicates
 			.concat(requestRefs); // add new requested refs
@@ -57,8 +57,8 @@ export function api(state: ApiState = DEFAULT_APP_STATE, action={}) {
 		delete state.fail;  // if there are any values, the API is not failing
 		return {
 			...state,
-			...responseToState(action.payload.response),
-			inFlight: state.inFlight.filter(ref => ref !== action.payload.response.ref),
+			...responseToState((action.payload || {}).response),
+			inFlight: state.inFlight.filter(ref => ref !== (action.payload || {}).response.ref),
 		};
 	case API_RESP_FAIL:
 		return { ...state, fail: action.payload };
--- a/src/reducers/platform.js
+++ b/src/reducers/platform.js
@@ -35,7 +35,7 @@ export const responseToState = (response: QueryResponse): { [string]: QueryRespo
 /*
  * The primary reducer for data provided by the API
  */
-export function api(state: ApiState = DEFAULT_API_STATE, action: FluxStandardAction) {
+export function api(state: ApiState = DEFAULT_API_STATE, action: FluxStandardAction): ApiState {
 	switch (action.type) {
 	case API_REQ: {
 		const requestRefs = (action.payload || []).map(({ ref }) => ref);
--- a/src/reducers/platform.js
+++ b/src/reducers/platform.js
@@ -26,7 +26,8 @@ type ApiState = {
 	[string]: QueryResponse,
 };
 
-export const DEFAULT_APP_STATE: ApiState = { inFlight: [] };
+export const DEFAULT_API_STATE: ApiState = { inFlight: [] };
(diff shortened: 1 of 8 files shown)


nlaz:
Thanks for this update! <3

nlaz:
Doesn't look like anyone is currently using `isFetching`: [mup-web](https://github.com/meetup/mup-web/search?utf8=%E2%9C%93&q=isFetching&type=), [pro-web](https://github.com/meetup/pro-web/search?utf8=%E2%9C%93&q=isFetching&type=), [admin-web](https://github.com/meetup/admin-web/search?utf8=%E2%9C%93&q=isFetching&type=).

Do we _really_ need to worry about legacy support? There's the other concern of cluttering the state and obfuscating functionality.

src/reducers/platform.js
```diff
@@ -19,45 +20,52 @@ import {
...
+		};
 	case API_RESP_FAIL:
-		state.fail = action.payload;
-		// fall through - fetch is complete
-	case API_RESP_COMPLETE:
-		return { ...state, isFetching: false };
-
-	default:
-		return state;
+		return { ...state, fail: action.payload };
```
nlaz: This should update `inFlight` state as well, right?

src/reducers/platform.js
```diff
@@ -19,45 +20,52 @@ import {
...
-		return { ...state, ...responseToState(action.payload.response) };
+		return {
+			...state,
+			...responseToState((action.payload || {}).response),
+			inFlight: state.inFlight.filter(ref => ref !== (action.payload || {}).response.ref),
+		};
 	case API_RESP_FAIL:
-		state.fail = action.payload;
-		// fall through - fetch is complete
-	case API_RESP_COMPLETE:
```
nlaz: Why get rid of this action? because `inFlight` should take care of it?

src/reducers/platform.test.js
```diff
@@ -75,55 +76,56 @@ describe('app reducer', () => {
...
+			inFlight: [],
 		});
 	});
 	it('populates an `fail` key on API_RESP_FAIL', () => {
 		const API_RESP_FAIL = apiActions.fail(new Error('this is the worst'));
-		const errorState = api({ ...DEFAULT_APP_STATE }, API_RESP_FAIL);
+		const errorState = api({ ...DEFAULT_API_STATE }, API_RESP_FAIL);
 		expect(errorState).toEqual(expect.objectContaining({ fail: API_RESP_FAIL.payload }));
 	});
-	it('sets isFetching:true on API_REQ', () => {
```
nlaz: Do we want to deprecate these tests instead of removing them?

src/reducers/platform.js
```diff
@@ -19,45 +20,52 @@ import {
...
  */
-export function api(state=DEFAULT_APP_STATE, action={}) {
+export function api(state: ApiState = DEFAULT_API_STATE, action: FluxStandardAction): ApiState {
 	switch (action.type) {
-	case API_REQ:
+	case API_REQ: {
+		const requestRefs = (action.payload || []).map(({ ref }) => ref);
+		const inFlight = state.inFlight
+			.filter(ref => !requestRefs.includes[ref]) // clean out current duplicates
+			.concat(requestRefs); // add new requested refs
```
nlaz: Can you explain again why it makes more sense to keep track of requests in an array versus an object?  I can why semantically it makes sense because its a list of objects. But it seems like it would simplify this step and every other step?

mmcgahan:
> Do we really need to worry about legacy support?

Probably not, but legacy support is 'free' and I had a weak inclination to save a few hard-breaking changes for v3. Good to know `isFetching` is not being used, though - probably should go ahead an d kill it

src/reducers/platform.js
mmcgahan: it's more structural than technical, but to my mind an array represents the 'flow' of `ref`s a little more accurately - we are essentially representing the 'queue' network requests, ordered by time. Queue/List/Array matches that analogy. The data structure is probably also a symptom of my preference for functional over OO solutions - managing mutability of state represented by an object is a little more bug prone because `state.inFlight[ref] = true` seems intuitive, but `state.inFlight = { ...state.inFlight, [ref]: true }` is actually preferred for immutability.

Objects also bring up a question of how to correctly represent a completed `ref` - do you set it to `false` or do you `delete` the key? What are the consequences of one approach vs the other? The object approach ultimately feels like a more complex one, particularly with the nice `Array.prototype` methods that are available for immutably updating arrays.

I'm open to finding out more about what problems an object-based representation would solve, but in the consumer code I don't think it would be much simpler - `inFlight.includes(ref) ?` vs `inFlight[ref] ?`

src/reducers/platform.js
mmcgahan: I'm a little unsure how to correctly handle `API_RESP_FAIL` and `API_RESP_COMPLETE`, because they currently don't report what refs they correspond to, so there's no clear way to accurately update `inFlight` when the response fails - the payload for `API_RESP_FAIL` is an `Error` object, and `API_RESP_COMPLETE` doesn't have a payload. Maybe those actions should have a `meta.queries` property so that it is easy to see what request(s) they correspond to. I'll make that update in this PR.

src/reducers/platform.test.js
mmcgahan: the deprecated tests are for the `app reducer` starting on line 13 - the changes here are just to update the tests to use a DEFAULT_STATE specific to `state.api` and change from `isFetching` to `inFlight` in the new system.

nlaz: Ah missed that!

src/reducers/platform.js
nlaz: Thanks for the clarity. That's a super reasonable explanation. I'm fine with the array approach.

nlaz:
The other big question I have is whether this is the best approach for handling loading states. Using `inFlight` would something like this right:
```js
// AttendeesContainer.jsx
const mapStateToProps = (state, ownProps) => ({
   isAttendeesListLoading: state.api.inFlight.includes(ATTENDEES_LIST_REF),
   attendees: state[EVENT_REDUCER].attendees,
});

// render
   <AttendeesList isLoading={isAttendeesListLoading} />
```
Is that preferable to managing `isLoading` inside the `attendees` object at the reducer level? Doing it inside related component seem more redux-y. We could theoretically support both. Is this going to be the preferred way to do things?

nlaz:
Fat fingers, sorry about closing!

mmcgahan:
> Is this going to be the preferred way to do things

I think so - we'll have to get a feel for it in practice, but I think the added complexity of a custom reducer should only be considered when `state.api` really doesn't provide sufficient data to represent an object. A related reason for not injecting `isLoading` data into particular objects is that the 'loading' state can/should be thought of more as 'network state' rather than 'data object state' - e.g. an _attendee_ object is intended to represent a _person_, not a network request, so `isLoading` ideally shouldn't be injected into it even though the UI needs to know about both types of data. We'll see how well that distinction holds in practice, though.

I'm hoping that over time there will be less pressure to munge data through custom reducers - I totally understand why they're used for mug comm, but I think with selectors and a cleaner/better-organized `state.api` object, they won't be necessary in the majority of cases, and different features across the site will have more similar `state.api`-based data flows.

mmcgahan:
cc @akdetrick I wanted to make sure you saw this PR - the `isFetching` setup we talked about a little while ago is being converted to an `inFlight` array that represents all the current 'open' API requests. Effectively, the global `isFetching` value can be determined with `state.api.inFlight.length > 0`

src/reducers/platform.js
mmcgahan: updated so that `API_RESP_COMPLETE` is solely responsible for clearing `inFlight`

PR #260, opened 2017-04-21
WP 308 Support variants service through API headers

mmcgahan:
This updates the Query `meta` object with a `variants` property that can be used to request 'variants' from the variants service integrated into the REST API: https://meetup.atlassian.net/wiki/display/MUP/X-Meetup-Variants

Docs included

```js
// request Query
{
  ...
  meta: {
    variants: {
      'my-experiment': memberId
    } 
  }
}

// query response
{
  ...
  meta: {
    variants: {
      'my-experiment': {
        [memberId]: 'variantname',
      }
    }
  }
}

// access from Redux state
const variant = state.api[ref].meta.variants['my-experiment'][memberId];
```

### DIFF
--- a/docs/Queries.md
+++ b/docs/Queries.md
@@ -47,7 +47,7 @@ A Query is just a plain object with the following shape:
     method: 'get',  // generally not needed
     variants: {
       'my-member-experiment': 1234,  // memberId
-      'my-group-experiment: 5678,    // chapterId
+      'my-group-experiment': 5678,    // chapterId
     },
   },
 }
--- a/docs/Queries.md
+++ b/docs/Queries.md
@@ -25,6 +25,9 @@ A Query is just a plain object with the following shape:
   meta?: {
     flags?: string[],
     method?: string
+    variants: {
+      [string]: string | string[],  // e.g. { experiment1: chapterId }
+    },
   },
 }
 ```
@@ -64,12 +67,12 @@ array or a singleton.
 - `comment`
 - feature-specific objects like `home`, `conversations`
 
-#### `flags`
+#### `meta`
+
+##### `flags`
 
 An array of feature flag (Runtime Flag) names that should be returned
 alongside the main request.
-
-#### `meta`
 
 ##### `method`
 
@@ -81,6 +84,13 @@ different `method`s.
 Alternatively, you can use method-specific action creators to automatically
 assingn the method and make the request for individual requests.
 
+##### `variants`
+
+You can request variant names for particular experiments with particular
+contexts by populating `meta.variants` with an object containing keys that are
+experiment names and values that are string IDs (member ID or chapter ID
+depending on the experiment)
+
 ### Query Response
 
 A query response is also a plain object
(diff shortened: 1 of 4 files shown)


mmcgahan:
this is a blocker for some member growth stuff, although it sounds like it's not immediately pressing

src/util/apiUtils.js
```diff
@@ -205,15 +230,25 @@ export const parseApiResponse = requestUrl => ([response, body]) => {
...
+	({ endpoint, params, flags, meta={} }) => {
 		const dataParams = querystring.stringify(params);
 		const headers = { ...externalRequestOpts.headers };
 		let url = encodeURI(`/${endpoint}`);
 		let body;
 		const jar = createCookieJar(url);
 
-		if (flags) {
-			headers['X-Meetup-Request-Flags'] = flags.join(',');
+		if (flags || meta.flags) {
```
willh-meetup: Are we planning on deprecating flags?

docs/Queries.md
```diff
@@ -90,8 +121,14 @@ A query response is also a plain object
   ref: string,
   value: any,
   type?: string,
-  flags?: array,
-  meta?: object,
+  meta?: {
+    flags?: string[],
+    variants?: {
+      [string]: {  // experiment
+        [string]: string  // context: variant
```
willh-meetup: nitpick, but context generally seems to be an integer right now? But, maybe that won't always be true?

src/util/apiUtils.test.js
```diff
@@ -254,6 +255,22 @@ describe('parseMetaHeaders', () => {
 	});
 });
 
+describe('parseVariantsHeader', () => {
+	it('parses a variants header into a nested object', () => {
```
willh-meetup: Possibly worth adding a test case for: "The server may not return a variant for a given experiment, and it may return nothing at all, depending on the availability of the variant service in GCP."

Basically, what happens when the header doesn't return a variant. Also, curious what that looks like... I assume it's `bing-pilot=123|` ?

docs/Queries.md
mmcgahan: I'd be interested in your recommendation on this - chapterId and memberId seem like they should be pretty safe as integers, but yeah I'm not sure if that is a hard rule for what 'context' can be in the variants service... maybe I can just declare `[string|number]` as the type and let consumers determine what they should use.

src/util/apiUtils.js
mmcgahan: just moving `flags` from a top-level Query property to `meta.flags`. All the documentation now advises the latter but for awhile it was defined as a top-level property so this takes care of both formats. I think there's only one flag actually being used in mup-web, so it should be easy to drop the old style completely - I'll open a ticket for that.

docs/Queries.md
willh-meetup: Yeah, I think `[string|number]` sounds right to me. I can definitely foresee at some point adding a context that isn't a number.

mmcgahan: actually, supporting a `number` in the parsed variants header would be a little awkward, e.g.

```js
const contextInt = parseInt(context, 10);
if (!isNaN(contextInt)) {
  // use`contextInt`
} else {
  // use `context`
}
```

so I'm inclined to force the _response_ state to be string-keyed, but allow the _request_ to use integer `context`s.

src/util/apiUtils.js
willh-meetup: Cool.

docs/Queries.md
willh-meetup: Ah, ok, yeah I'm cool with that

PR #262, opened 2017-04-24
Deprecate browser render

mmcgahan:
**Prelim until I flesh out docs and tests**

Instead of letting the platform call `ReactDOM.render`, it will be better to package up the browser rendering into a component that can be directly imported by consumer applications. This will allow consumer apps to further wrap the application if they so choose, e.g. for hot reloading.

cc @mikespencer I think you might appreciate this one

### DIFF
--- a/src/components/BrowserApp.jsx
+++ b/src/components/BrowserApp.jsx
@@ -1,0 +1,59 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { getBrowserCreateStore } from '../util/createStoreBrowser';
+import BrowserRouter from 'react-router-dom/BrowserRouter';
+import PlatformApp from '../components/PlatformApp';
+
+/**
+ * This function creates a 'renderer', which is just a function that, when
+ * called, will call ReactDOM.render() to render the application
+ *
+ * The routes, reducer, and app-specific middleware are provided by the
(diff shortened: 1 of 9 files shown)


src/components/BrowserApp.jsx
```diff
@@ -0,0 +1,30 @@
...
+				<PlatformApp store={store} routes={routes} />
+			</BrowserRouter>
+		);
+	}
+}
+
+BrowserApp.propTypes = {
+	routes: PropTypes.array.isRequired,
+	store: PropTypes.object.isRequired,
+	baseUrl: PropTypes.string.isRequired,
```
mikespencer: Should this be `basename`? (Also in default props)

mmcgahan: ah, thanks, I was flipping back and forth on the name of this property because some packages refer to it as baseUrl and some refer to it as basename so I'm undecided where I should make the name switch in the platform interface...

mmcgahan:
> the idea is that you are able to import <BrowserApp /> and render that instead of importing and calling makeRenderer from browser-render.jsx

Correct. I'm flip-flopping about how much logic will get packaged into `<BrowserApp>`, so I need to make some decisions on that today and make this ready for final review

PR #263, opened 2017-04-25
convert MEETUP_LANGUAGE to accept-language header

mmcgahan:
https://meetup.atlassian.net/browse/WP-310

apparently the API doesn't read the MEETUP_LANGUAGE cookie to set the language of its response - it respect `Accept-Language` only

This PR forces the `Accept-Language` header to be the cookie language.

### DIFF
--- a/src/util/apiUtils.js
+++ b/src/util/apiUtils.js
@@ -320,11 +320,22 @@ export function getAuthHeaders(request) {
 	};
 }
 
+export function getLanguageHeader(request) {
+	const cookieLang = getCookieLang(request);
+	const headerLang = request.headers['accept-language'];
+	const acceptLang = cookieLang ?
+		`${cookieLang},${headerLang}` :
+		headerLang;
+	return {
+		'accept-language': acceptLang,
+	};
+}
+
 export function parseRequestHeaders(request) {
 	const externalRequestHeaders = {
 		...request.headers,
 		...getAuthHeaders(request),
-		'accept-language': getCookieLang(request) || request.headers['accept-language'],
+		'accept-language': getCookieLang(request),
 	};
 
 	delete externalRequestHeaders['host'];  // let app server set 'host'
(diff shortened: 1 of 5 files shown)


mmcgahan:
actually this needs a few unit tests, coming shortly!

src/util/apiUtils.js
```diff
@@ -317,10 +320,22 @@ export function getAuthHeaders(request) {
 	};
 }
 
+export function getLanguageHeader(request) {
+	const cookieLang = getCookieLang(request);
+	const headerLang = request.headers['accept-language'];
+	const acceptLang = cookieLang ?
+		`${cookieLang},${headerLang}` :
```
sadafie: Curious if the cookie language would ever be different than the header Language?

sadafie: From slack:

> - MEETUP_LANGUAGE may not be set - browser will set accept-language
>- browser ‘accept-language’ can have a series of fallbacks

sadafie: Only thing I can see happening here is dupe language entries, ie. something like `en-US,en-US,en`, but I'm guessing that won't matter

src/util/apiUtils.js
```diff
@@ -317,10 +320,22 @@ export function getAuthHeaders(request) {
...
+		headerLang;
+	return {
+		'accept-language': acceptLang,
+	};
+}
+
 export function parseRequestHeaders(request) {
 	const externalRequestHeaders = {
 		...request.headers,
```
eilinora: Hmmm I wonder why it wasn't getting passed from the original request as it is passing accept-language. 
screen shot 2017-04-25 at 2 15 56 pm

mmcgahan: I think `Accept-Language` was getting passed, it just wasn't being set to the value of the `MEETUP_LANGUAGE` cookie

eilinora: actually, i see even when i change my locale it wasn't passing the write language to accept-language

PR #264, opened 2017-04-25
Node Convict

sadafie:
This is still Prelim - I need to fix the integration tests. I also need to go through your previous questions and comments in the old PR to make sure I covered everything. But feel free to look through this PR.


swozniak:
Overall this looks good. Have you tried pairing this with a consumer repo yet?

I still probably have the same question as before re: if we're okay with the config import returning the node-convict object instead of a plain JavaScript object literal - for maintenance reasons it seems easier to just read values from a JSON object.

sadafie:
> I still probably have the same question as before re: if we're okay with the config import returning the node-convict object instead of a plain JavaScript object literal - for maintenance reasons it seems easier to just read values from a JSON object.

Sorry, I must have missed that in the previous PR! It's easy enough to return `.getProperties()` and have the config be a plain JSON object. But in that case, I guess I'm missing the big picture about why we're wanting to use `node-convict` to begin with? Per my understanding, the value of `node-convict` is to have a config module that you import and easily access values via getters and setters. If we're opting to keep our methods pure and pass the config object around as arguments, and we rather the object be a plain JSON object, what is `node-convict` getting us that our current implementation isn't?

src/plugins.js
```diff
@@ -41,7 +44,7 @@ export function getCsrfPlugin(secret) {
 	const register = (server, options, next) => {
 		const cookieOptions = {
 			path: '/',
-			isSecure: process.env.NODE_ENV === 'production',
+			isSecure: config.get('isProd'),
```
mmcgahan: prefer to read from `server.app.isProd`

mmcgahan: related - use `server.app.csrf_secret` for `options.secret` on line 67 instead of passing it in as an argument. That way, you don't have to pass `config` into `getPlugins` at all

src/plugins/requestAuthPlugin.js
```diff
@@ -107,17 +107,17 @@ export const applyRequestAuthorizer$ = requestAuthorizer$ => request => {
  * Get an anonymous code from the API that can be used to generate an oauth
  * access token
  *
- * @param {Object} config { OAUTH_AUTH_URL, oauth }
+ * @param {Object} server the Hapi app server instance
```
mmcgahan: I think I'd favor passing just `server.app` instead of the whole `server` instance - the function doesn't need to know about anything other than a few config vars, and the original argument destructuring makes the dependencies explicit.

Not sure if you'd want to do it here, but the `redirect_uri` value is hardcoded in this module and could instead be part of `config` as `oauth_redirect_uri` or something - let me know if you _don't_ want to do that here so that I can make a ticket for doing it later.

src/plugins/requestAuthPlugin.js
```diff
@@ -252,13 +252,12 @@ const refreshToken$ = refresh_token => Observable.of({
  * For an anonymous auth, the request header information is used to determine
  * the location and language of the anonymous member
  *
- * @param {Object} config { OAUTH_AUTH_URL, OAUTH_ACCESS_URL, oauth }
- * @param {Object} request the Hapi request that needs to be authorized
+ * @param {Object} server the Hapi app server instance
  */
-export const getRequestAuthorizer$ = config => {
+export const getRequestAuthorizer$ = (server) => {
```
mmcgahan: why consume the whole server instance? just need `server.app`

src/plugins/requestAuthPlugin.test.js
```diff
@@ -12,10 +14,51 @@ import register, {
...
+};
+
+const MOCK_HEADERS = {};
+
+const MOCK_OAUTH = {
+	auth_url: 'http://example.com/auth_fakeout',
+	access_url: 'http://example.com/access_fakeout',
 	key: '1234',
-	secret: 'asdf',
+	secret: 'asdf'
```
mmcgahan: nitpick - trailing commas are useful

src/util/config.js
```diff
@@ -1,83 +1,157 @@
...
+		env: 'PHOTO_SCALER_SALT',
+	}
+});
 
-	// currently all config is available syncronously, so resolve immediately
-	return Promise.resolve({ ...config, ...overrideConfig })
-		.then(validateConfig);
+// Load environment dependent configuration
+const env = config.get('env');
+const configFile = `../config.${env}.json`;
```
mmcgahan: what is this path relative to? probably more explicit to use `path.resolve(__dirname, ...)`

src/util/serverUtils.js
```diff
@@ -130,15 +113,12 @@ export function registerExtensionEvents(server) {
...
+ * @param {Object} config the node-convict config object
+ * @return null
  */
 export function configureEnv(config) {
 	// When using .dev.meetup endpoints, ignore self-signed SSL cert
-	const USING_DEV_ENDPOINTS = checkForDevUrl(config);
-	https.globalAgent.options.rejectUnauthorized = !USING_DEV_ENDPOINTS;
-
-	return config;
+	https.globalAgent.options.rejectUnauthorized = config.get('isProd');
```
mmcgahan: at the moment, NODE_ENV and api urls are independent, so `checkForDevUrl` could return a different value than `!config.get('isProd')`. Let's leave `checkForDevUrl` in place for now, and I'll clean it up in https://meetup.atlassian.net/browse/WP-158

src/util/config.js
swozniak: Would it be better if we just expected the consumer app to pass in the full path to a config file on server start?

src/plugins.js
sadafie: 👍 still missing a few `server.app` usages

src/plugins/requestAuthPlugin.js
sadafie: nope make sense, i can do it here!

src/util/config.js
sadafie: I'll use `path.resolve`. 

> Would it be better if we just expected the consumer app to pass in the full path to a config file on server start?

My logic was we want environment specific config files, so either this logic goes here or in the consumer app, and it seems more clear to have the platform dictate how to configure itself. i'm open to changing this though

swozniak:
> Sorry, I must have missed that in the previous PR! It's easy enough to return .getProperties() and have the config be a plain JSON object. But in that case, I guess I'm missing the big picture about why we're wanting to use node-convict to begin with? Per my understanding, the value of node-convict is to have a config module that you import and easily access values via getters and setters. If we're opting to keep our methods pure and pass the config object around as arguments, and we rather the object be a plain JSON object, what is node-convict getting us that our current implementation isn't?

For me, I saw a few specific benefits to us:
- Dedicated tooling for environment configuration management with node apps from a reputable team (Mozilla)
- Built-in support for loading and merging config files. We can now just read a `.mupweb.config`-like file directly instead of having to unset/set those variables whenever we change them. Also, devs can run mup-web, pro-web, admin-web locally each with its own set of ports, configuration, etc.
- Out-of-the-box (well, with `node-validator`) validation with presets for common env config variables like ports, urls, ips, etc., instead of writing our own code to pair it with Joi validation.

src/util/config.js
mmcgahan: Having a hard-coded 'default' config file location is probably best to start - I like the idea of passing in a config file path as a command line arg, but let's aim to do interface optimization in the future.

_Who configs the config?_

src/plugins/requestAuthPlugin.js
sadafie: Unrelated thought --- should we be distinguishing between `server.settings.app` and `server.app`? From the [hapi documentation](https://hapijs.com/api):

```
app - application-specific configuration which can later be accessed via server.settings.app. Note the difference between server.settings.app which is used to store static configuration values and server.app which is meant for storing run-time state. Defaults to {}.
```

Perhaps a follow on task if we decide that we should.

src/plugins/requestAuthPlugin.test.js
```diff
@@ -56,65 +95,72 @@ const MOCK_AUTHED_REQUEST = {
...
+
+		const getToken$ = getAccessToken$({ ...MOCK_SERVER_APP }, null);
 		getToken$(MOCK_HEADERS)(MOCK_CODE).subscribe(done);
 	});
-	it('throws an error when no oauth.key is supplied', function() {
-		const oauthNoKey = { ...oauth };
-		delete oauthNoKey.key;
-		expect(() => getAccessToken$({ oauth: oauthNoKey, OAUTH_ACCESS_URL }, null))
+
+	xit('throws an error when no oauth.key is supplied', function() {
```
sadafie: This test (and the one after) is causing `MOCK_SERVER_APP` to be mutated, but I'm not sure why this is happening. 

Because it's deleting `MOCK_SERVER_APP.oauth.key`, the [subsequent test](https://github.com/meetup/meetup-web-platform/pull/264/files#diff-b061c55b4fb48f7b0f38751009857f4cR150) is failing. 

Anyone have thoughts? cc @mmcgahan

mmcgahan: object destructuring makes a shallow copy, so you'll need to make a deeper copy of `...MOCK_SERVER_APP.oauth`

```js
const oauth = {
  ...MOCK_SERVER_APP.oauth,  // shallow copy of `.oauth`
};
delete oauth.key;  // delete from shallow copy

const serverAppNoOauthKey = {
  ...MOCK_SERVER_APP,
  oauth,
};
```

src/plugins/requestAuthPlugin.js
mmcgahan: oooh, I haven't come across `server.settings.app` - that's definitely what we should be using. We'll get into long property chains when accessing it, but that's probably okay

```js
const ouathKey = response.request.server.settings.app.oauth.key;
```

src/plugins/requestAuthPlugin.test.js
sadafie: I see! Do you know if there's a way to test if something is a copy or a reference of an object? I was simply debugging with console.logs and the data structure looks correct, so I was missing the shallow copy portion.

mmcgahan: I think it's just something you have to get accustomed to with using object destructuring - it's definitely a pitfall when reducers and such that update Redux state as well. Adam D suggested using a lib called deepFreeze to help make sure that tests aren't modifying objects (https://meetup.atlassian.net/browse/MW-383), but unless we wanted to go all-in with something like Immutable.js, I'm skeptical about using a tool that protects against mutation in some places and not others.

For me, it just took a little time to get used to and now every time I see `{ ...foo }`, I immediately check to make sure that no deeper mutations are happening. There's still a risk but the solution involves a pretty heavy refactor.

src/plugins/requestAuthPlugin.js
mmcgahan: Interestingly, `server.settings.app` appears to be the key to fixing the integration tests.

When the plugin `register` function is called, it doesn't appear to receive `server.app`, but it does receive `server.settings.app`. If I squint, I can sort of see the reasoning behind this since `server.app` is intended to be for runtime state, and plugins are meant to be loaded statically, so should in theory only be provided with static configuration info.

I'll start refactoring to `server.settings.app` and confirm that it is working as intended

src/plugins.js
```diff
@@ -30,62 +30,62 @@ export function setCsrfCookies(request, reply) {
...
+ * @param {Object} config the environment configuration object
+ * @return {Object} the { register } object for a `server.register` call.
  */
-export function getCsrfPlugin(secret) {
+export function getCsrfPlugin(config) {
 	const register = (server, options, next) => {
 		const cookieOptions = {
 			path: '/',
-			isSecure: process.env.NODE_ENV === 'production',
+			isSecure: options.isProd,
```
mmcgahan: `server.settings.app.isProd`

src/plugins.js
```diff
@@ -30,62 +30,62 @@ export function setCsrfCookies(request, reply) {
...
+
 	register.attributes = CsrfPlugin.register.attributes;
+
 	return {
 		register,
 		options: {
-			secret,
-		}
+			secret: config.csrf_secret,
+			isProd: config.isProd,
```
mmcgahan: don't need `options` at all any more

src/routes/appRoute.js
```diff
@@ -16,7 +17,7 @@ export const onPreResponse = {
 	 */
 	method: (request, reply) => {
 		const response = request.response;
-		if (!response.isBoom || process.env.NODE_ENV === 'production') {
+		if (!response.isBoom || config.isProd) {
```
mmcgahan: this should be `request.server.settings.app.isProd` - you can remove the `config` import

mmcgahan:
Summary of updates:

1. refactor to `server.settings.app` so that plugins can access config
2. update a few tests to use `tests/mocks.js:getServer` a bit more consistently + populate the mock config more fully (this could be simplified further by importing `config.test.json`, but that can be a followup).
3. add `config.test.json` so that config-dependent unit tests can get a valid mock config in Travis
4. fix the import path to the test config file - it should be relative to the `process.cwd()`, which is the directory from which the node process was started (typically the _consumer app_ root_)

mmcgahan:
one other thing: this error appears when you run tests with `--verbose`:

```
      Warning: configuration param 'duotone_urls' not declared in the schema
      configuration param 'api_server_root_url' not declared in the schema
      configuration param 'isProd' not declared in the schema
      configuration param 'isDev' not declared in the schema
```

This indicates that the `config` properties that are being manually set must still be declared in the `convict` schema.

src/plugins.js
sadafie: I was trying that before --- `server.app` was not set at all, so I passed it back in as options. I can see if it's set in `server.settings.app`, but I'm now a little confused about when `server.app` is available (I'm assuming it's not because the top level `server()` call has completed yet), and when we should be using `settings.app` v. `app` within our code.

sadafie:
Thank you @mmcgahan ! 

> refactor to server.settings.app so that plugins can access config
> update a few tests to use tests/mocks.js:getServer a bit more consistently + populate the mock config more fully (this could be simplified further by importing config.test.json, but that can be a followup).

Could you point me to the commit that does those two things? I clicked through each one, but with the formatting changes, I couldn't spot these specific changes. Just want to see/understand it.

src/plugins.js
sadafie: Same method from last comment -  will fix, but would like to understand why `server.app` didn't work, but `server.settings.app` does now

src/plugins/requestAuthPlugin.js
sadafie: Ohhhh! I'm just seeing this comment ---- you can ignore my questions from the comments below :) 

I'll update that last reference to use server.settings.app and presume to use server.settings.app everywhere then?

src/plugins.js
sadafie: I just saw your explanation above! I will fix this :)

sadafie:
>This indicates that the config properties that are being manually set must still be declared in the convict schema.

Fixed

@mmcgahan @swozniak This is ready for final review! Thank you!

src/util/apiUtils.js
```diff
@@ -589,7 +593,7 @@ export const injectResponseCookies = request => ([response, _, jar]) => {
 			path: cookie.path,
 			isHttpOnly: cookie.httpOnly,
 			isSameSite: false,
-			isSecure: process.env.NODE_ENV === 'production',
+			isSecure: config.isProd,
```
sadafie: is `request.server` available here?

src/util/config.js
```diff
@@ -1,97 +1,168 @@
-import Joi from 'joi';
+import fs from 'fs';
```
sadafie: We should probably have unit tests for this. I'm wondering what we can test? That it's exports a valid JSON object? Test for specific config params?

src/util/apiUtils.js
sadafie: @mmcgahan would you happen to know why `request.server.settings` is not available here?

mmcgahan:
> 1. we have a `config.test.json` file commited now. we really don't need the `config.dev.json` yet, because all the defaults in the `util/src/config.js` default to dev values. But, do we need one for prod?

For prod, we should always rely on env variables - we definitely don't want to commit any prod secrets/config.

> 2. Do we consider this a breaking change? Should I wait until I create a corresponding branch in `mup-web` before merging this?

No, but it should definitely bump the minor version and deprecate the env-variables-in-dev that was described in the README, which brings up an important change - the README needs to document a complete `config.development.json` file that can be used when there are _no_ environment variables. Currently it just describes a stub that still relies on 'secrets' populated in env variables.

The minor version update should get an entry in the CHANGELOG.md

src/util/apiUtils.js
mmcgahan: no, that seems odd. Let me know if you'd like some help investigating

src/util/config.js
swozniak: It wouldn't hurt to test any custom validation functions we write.
Otherwise yeah, the only other thing I would test is to confirm it returns a valid JS object.

src/util/apiUtils.js
sadafie: The unit test mock that it was using didn't have a `server` prop so I added it with `getServer()` 👍

sadafie:
> No, but it should definitely bump the minor version and deprecate the env-variables-in-dev that was described in the README, which brings up an important change - the README needs to document a complete config.development.json file that can be used when there are no environment variables. Currently it just describes a stub that still relies on 'secrets' populated in env variables.

I've already updated the README. The `config.development.json` file is completely optional because node-convict will use the default values in `config.js`, and those are currently all dev values. Let me know if you think the existing README changes aren't enough.

I'll update the CHANGELOG and bump the minor version number.

src/util/config.js
```diff
@@ -1,97 +1,183 @@
...
+			format: 'int',
+			default: 8000,
+			env: 'API_TIMEOUT',
+		},
+		root_url: {
+			format: String,
+			default: '',
+		},
+	},
+	asset_server: {
```
sadafie: Just thought of this - 

`asset_server` is a `mup-web` configuration, and not related to the platform code. Should mup-web have it's own convict instance for `asset_server`, `asset_path`, `enable_hmr`, and other such consumer configurations?

sadafie:
@mmcgahan this is ready for final review. Is it good to merge?

swozniak:
Just to confirm, this optionally reads a `config.dev.json` file from the root of the consumer repo?

Do we have a strategy for moving everyone off `~/.mupweb.config`? We probably need to update the README.md in the `web-platform-starter`/consumer repos too, right?

mmcgahan:
@swozniak yes, the [`process.cwd()`](https://github.com/meetup/meetup-web-platform/pull/264/files#diff-19ae3cea58dd881db041e62b0b2ba24bR147) command will return the current working directory from which the node process was started - i.e. the consumer repo root.

We definitely need to update the starter kit README - it should actually be linking to the MWP docs rather than duplicating the environment config instructions.

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## [2.5]
+
+- **Deprecated** `~/.mupweb.config` has been deprecated - node convict now
+  manages are configurations in `util/src/config.js`. To override any of the
```
mmcgahan: `are` => `our`

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
...
+  default values in development, create a `config.development.json` file
+  with the name of the configuration and new value. This file is automatically
+  gitignored.
```
mmcgahan: Probably useful to add a line: "See `config.test.json` for an example config"

README.md
```diff
@@ -22,36 +22,24 @@ In general, application-specific code will live outside of this package.
...
-```
-
-**Note**: you _can_ use `dev.meetup.com` URLs for `API_HOST`, `OAUTH_AUTH_URL`,
-and `OAUTH_ACCESS_URL`, but you will need to ensure that your devbox is up and
-running with a recent build of Meetup classic.
-
-To automatically add these env variables into your terminal session,
-`source` the config file in your `.bashrc` or `.zshrc`:
+For example, if you want to use the production api in dev, 
+create the following `config.dev.json` file:
```
mmcgahan: Actually I should have made the instructions comment here - we need to document what a 'full' configuration file looks like, ideally with explanations of each key, but that can come later. The main thing that people will need is a reference for what a complete config looks like and I think you can just link/point them to `config.test.json` here

mmcgahan:
This will need an update to split the _build time_ config from the _runtime config_ - we have intentionally avoided passing runtime config/secrets through Travis at build time, so we need to have a version of the config that can be validated for the build without having access to

```
cookie_encrypt_secret: Cookie Encrypt Secret must be a random 32+ char string
csrf_secret: CSRF Secret must be a random 32+ char string
oauth.secret: Invalid OAUTH Secret
oauth.key: Invalid OAUTH Key
photo_scaler_salt: Invalid Photo Scaler Salt
```

These variables are passed into the container by Kubernetes at runtime.

One option is to break this config into two modules, e.g. `src/util/buildConfig.js` and `src/util/runtimeConfig.js`, each of which can be imported separately, _or_ having a single module that exports buildtime and runtime 'getters' that only validate the config for the appropriate version of the config. The former is slightly simpler to consume, the latter is DRYer and I think it provides better code organization.

The getters could be memoized so it's not _actually_ validating the config each time the config is accessed... I'll have to think about this a little more.

mattkime:
@mmcgahan sounds like it might be time to put together a new plan of attack on this. now that we have a better understanding of scope and the units of work to be done perhaps we can define simpler bits that can be committed easier.

mmcgahan:
I just checked on mup-web, and the beta version from this PR works there on `master` (no breaking changes), so I'm actually thinking it might be best to merge this PR and then start a new one to make the changes required to get it running in consumer applications - the follow-on changes are much smaller than what has already been reviewed and approved here, so that is probably the least complicated approach

PR #265, opened 2017-04-26
refactor makeRenderer - webpack dll changes

mattkime:


src/renderers/server-render.jsx
```diff
@@ -31,15 +31,9 @@ const DOCTYPE = '<!DOCTYPE html>';
  * @module ServerRender
  */
 
-function getHtml(baseUrl, assetPublicPath, clientFilename, initialState={}, appMarkup='') {
+function getHtml(props) {
```
mmcgahan: definitely an overdue change - instead of opaque `props` object however, we could eliminate another layer of abstraction by just accepting a `<Dom>` instance - it's easy enough for the caller to write `getHtml(<Dom {...props}>)` and the actual values are not used in the `getHtml` function - it also makes the `getHtml` call a bit more self-explanatory while reducing the need for redundant code documentation

src/renderers/server-render.jsx
```diff
@@ -124,6 +118,15 @@ const getRouterRenderer = (
...
 };
 
+const makeRenderer$ = (config: Object) => makeRenderer(
+		config.routes,
+		config.reducer,
+		config.clientFilename,
+		config.assetPublicPath,
+		config.middleware,
+		config.baseUrl
+	);
```
mmcgahan: it would be helpful to define a `type` for this `config` - there are lots of `config` objects floating around server code.

```js
type ServerRenderConfig = {
  routes: Array<Object>,  // we'll define a `Route` type at some point to make this `Array<Route>`
  reducer: (state: ?Object, action: FluxStandardAction) => Object,
  ...
}
```

mmcgahan: This unused constant should also be throwing lint errors - looks like we need to add `yarn lint` to the `test` script.

mattkime: I'll address this later (but before closing the pr). At the moment, flow looks like gibberish to me.

mattkime:
@mmcgahan I could use your advice on this - 

```
src/renderers/server-render.jsx:158
158: 	reducer: Reducer<?Object, FluxStandardAction>,
     	         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type application of identifier `Reducer`. Expected polymorphic type instead of
129: 		config.reducer,
     		^^^^^^^^^^^^^^ type `Reducer`
```

what is it complaining about?

mmcgahan:
@mattkime I think I might have the type declaration wrong for `Reducer` - I'll look into this tomorrow

mmcgahan:
I fixed the issue in `master` and merged the changes here

TL;DR: the type annotation for `reducer` in `makeRenderer` was incorrect

Details: the `<Object, FluxStandard>` suffix is used when you want to have 'type placeholders'/[generics](https://flow.org/en/docs/types/generics/) in the type declaration, which is called polymorphism. I thought that it corresponded to the arguments of the function, but the function argument types are already defined and are unrelated to the generics syntax, so the fix was just to remove all the generics syntax - 5a9d032

mattkime:
@mmcgahan i think this is ready to go. let me know if there's anything i need to address

src/components/dom.jsx
```diff
@@ -70,7 +75,7 @@ const DOM = (props) => {
 			<body>
 				<div id='outlet' dangerouslySetInnerHTML={getInnerHTML(appMarkup)} />
 				<script dangerouslySetInnerHTML={getInnerHTML(`window.APP_RUNTIME=${JSON.stringify(APP_RUNTIME)};`)} />
-				<script src={`${assetPublicPath}${clientFilename}`} />
+				{scripts.map(url => <script src={url} />)}
```
mmcgahan: React throws a warning about inserting arrays of elements without `key` props - you can see the error generated by this line if you run the integration tests without `--silent`.

`.map` makes sense if you have a large or arbitrary length array, but since we really want to keep the number of `<script>` tags small and explicit, I recommend continuing to write out each `<script>` individually (do you expect more than 2-3 total?)

If you want to map, use

```jsx
{scripts.map((src, key) => <script key={key} src={src} />)}
```

src/components/dom.jsx
```diff
@@ -59,6 +60,10 @@ const DOM = (props) => {
 		escapedState,
 	};
 
+	if(clientFilename){
```
mmcgahan: I'm interested in the motivation for making this optional - in what case do you want to render the app without a client script?

mattkime: backwards compatibility

src/components/dom.jsx
mattkime: >I recommend continuing to write out each <script> individually

I don't want to lock all platform consumers into using the same number of js files.

src/components/dom.jsx
mmcgahan: Can you elaborate? It's hard to evaluate this kind of change without a description of what it's moving toward - as a standalone change, it is a step backward because it allows the client to call `makeRenderer` in a way that would silently break the application (no client-side rendering).

If this change is only needed when we have a different implementation to replace it, I'd prefer to make the change at that point to see it in context.

mmcgahan: just catching up on mup-web PR - I see that you're intending `scripts` to be the single source of truth for the scripts to render. That makes sense - the thing I'd recommend here is throwing an error like

```js
if (!scripts.length) {
  throw new Error('No client scripts supplied - check the `makeRenderer` call in your server entry script');
}
```

PR #266, opened 2017-04-28
WP 269 async route support

mmcgahan:
Code splitting + server rendering is a non-trivial problem as [described in the React Router docs](https://reacttraining.com/react-router/web/guides/code-splitting/code-splitting-server-rendering), which actually recommend dropping SSR entirely. The docs lay out the challenges quite well:

> 1. You need synchronous module resolution on the server so you can get those bundles in the initial render.
> 2. You need to load all the bundles in the client that were involved in the server render before rendering so that the client render is the same as the server render. (The trickiest part, I think its possible but this is where I gave up.)
> 3. You need asynchronous resolution for the rest of the client app’s life.
>
> ...Godspeed those who attempt the server-rendered, code-split apps.

However, our pattern of using a plain JS object to define the route configuration offers a solution that is _much_ simpler than a 'pure' component-based `<Route>` system. The route configuration object is essentially a lightweight routing table that is used to generate the full component-based application structure. Think of it as a routing blueprint (I might trademark that phrase).

Here's a summary of how it works

1. Synchronous module resolution on the server:

    The server reads through the route configuration and finds the active routes for a particular URL. If any matched route has a `load` property, that function is called and the (async) result is used to inject the resolved `component` as a new `component` property on the route. The resulting _modified_ route object is then used to render the application.

2. Loading async bundles on client before render:

   The browser can use a similar pattern to the server to _delay_ rendering until all bundles required for _initial_ rendering are received. This is only possible because the route config object can be read _without_ rendering the application, unlike 'standard' RRv4 apps.

3. You need asynchronous resolution for the rest of the client app’s life.

    No problem - when a new async route is first rendered, a dummy `Loading` component is rendered until the updated bundle is received, at which point the async component is cached for later re-use when the corresponding 'load' function is next encountered. This is the primary use of the `<AsyncRoute>` component

normal route:

```js
{
  path: '/',
  component: FooContainer
}
```

async route:
```js
{
  path: '/',
  load: () => import('./FooContainer').then(c => c.default),
}
```

#### Followup

- discuss the 'loading state' - how do we make this better/configurable? The trick is that we want any loading state to be as lightweight as possible so that the core app bundle isn't weighed down with arbitrary dependencies, but we also want the loading state to be 'relevant' and not too generic. cc @akdetrick to get your thoughts.


mmcgahan:
This implementation is going to need an update - resolving component modules async is a good proof of concept, but we really need to be able to load entire route subtrees asynchronously in order to split the app into reasonably-large chunks and not worry about needing every route to define itself asynchronously.

The main downside of asynchronous routes is that the client requires two round trips in order to render a page - first to get the async route definition, and second to fetch the queries that are defined in the newly-loaded piece. I'll need to see if there's a clever way of getting all the data in one request, but at the moment I think it's a necessary compromise. However, server push and optimistic pre-fetching might obviate the timing problem.

I am hoping to get this wrapped up tomorrow

mmcgahan:
Needs a few `routeUtils` tests, but otherwise good to go.

.eslintrc.js
```diff
@@ -21,7 +21,6 @@ const config = {
 		'array-callback-return': 2,
 		'comma-dangle': 0,
 		'eol-last': 2,
-		indent: [2, 'tab', { SwitchCase: 1 }],
```
mmcgahan: managed by Prettier now

Makefile
```diff
@@ -1,5 +1,5 @@
 CI_BUILD_NUMBER ?= $(USER)-snapshot
-VERSION ?= 2.4.$(CI_BUILD_NUMBER)
+VERSION ?= 2.5.$(CI_BUILD_NUMBER)
```
willh-meetup: Pretty sure it doesn't, but this won't break existing functionality right?

README.md
```diff
@@ -17,6 +17,7 @@ In general, application-specific code will live outside of this package.
 - [Auth flow from `requestAuthPlugin`](docs/auth.md)
 - [API requests through api-proxy](docs/api-proxy.md)
 - ['Query': structuring data dependencies](docs/Queries.md)
+- [Rendering in consumer applications](docs/Rendering.md)
```
willh-meetup: Should we be linking to `Routing.md` here too?

src/util/createStoreServer.js
```diff
@@ -30,7 +30,7 @@ export const serverFetchQueries = request => () => queries =>
  */
 export function getServerCreateStore(routes, middleware, request, baseUrl) {
 	const middlewareToApply = [
-		catchMiddleware,
+		catchMiddleware(err => request.server.app.logger.error(err)),
```
willh-meetup: Are these changes related, or just making sure we are actually console logging errors? :)

src/util/routeUtils.js
```diff
@@ -1,74 +1,92 @@
...
+ * routes, called with the provided location URL
+ *
+ * The function returned from calling `_matchedRouteQueriesReducer` with a
+ * `location` should be used as the callback to an `array.reduce` call.
  */
-export const matchedRouteQueriesReducer = location => (
-	queries,
-	{ route, match }
-) => {
+export const _matchedRouteQueriesReducer = (location: URL) => (
```
mmcgahan: ah, i need to drop the `export` because it is expected to _only_ be used internally by `getMatchedQueries`

willh-meetup: 👍  There was another below this too.

Makefile
mmcgahan: correct - async routing is an opt-in feature and existing routing config _should_ continue to work normally

README.md
mmcgahan: yes

docs/Routing.md
```diff
@@ -56,12 +61,39 @@ Only render the current route when the URL matches exactly
...
 contains just the `component` that should be rendered when the location/url
 matches the root `path` _exactly_.
 
+### `getIndexRoute` (asynchronous)
+
+A function that, when invoked, will return a Promise that resolves with a
+`Route` object. This can be used for code splitting by using `import()` to
+dynamically import a route to be used as the `exact` index root.
+
+### `routes`
```
willh-meetup: (synchronous)?

src/util/createStoreServer.js
mmcgahan: just an update from the current `if (typeof window === 'undefined')` structure in order to use the new and improved server-side logger on the server - not related to routing update, so I can break it out into a separate PR if it's preferable.

willh-meetup: Nah, it's cool, just wanted to make sure I wasn't missing something.

Makefile
willh-meetup: 👍

docs/Routing.md
willh-meetup: Or, maybe not? If not, maybe move this below so all the sync/async bits are together.

PR #269, opened 2017-05-04
Route post/delete middleware functionality through apiActionCreators/sync middleware

mmcgahan:
This PR helps bridge the gap between post/delete middleware and the new apiActionCreators/sync system, and _greatly_ simplifies the post/delete middleware, removing their dependency on fetchUtils.

The only thing that post/delete did that wasn't supported by sync middleware was having custom `onSuccess` and `onError` handlers, but this PR allows those handlers to be passed as an additional `meta` argument to method-specific apiActionCreators (`meta` was already supported by the root `api.requestAll` action creator - this PR just surfaces that option to method-specific action creators).

### DIFF
--- a/src/actions/apiActionCreators.js
+++ b/src/actions/apiActionCreators.js
@@ -21,12 +21,12 @@ export function requestAll(queries, meta) {
 	};
 }
 
-export const _applyMethod = method => query => {
+export const _applyMethod = method => (query, meta) => {
 	query.meta = {
 		...(query.meta || {}),
 		method,
 	};
-	return requestAll([query]);
+	return requestAll([query], meta);
 };
 
 export const get = _applyMethod('get');
--- a/src/actions/apiActionCreators.js
+++ b/src/actions/apiActionCreators.js
@@ -21,7 +21,7 @@ export function requestAll(queries, meta) {
 	};
 }
 
-export const _applyMethod = method => (query, meta) => {
+const _applyMethod = method => (query, meta) => {
 	query.meta = {
 		...(query.meta || {}),
 		method,
--- a/src/epics/mutate.js
+++ b/src/epics/mutate.js
@@ -1,15 +1,7 @@
-import { Observable } from 'rxjs/Observable';
-import 'rxjs/add/observable/of';
-import 'rxjs/add/observable/from';
-import 'rxjs/add/observable/fromPromise';
-import 'rxjs/add/operator/catch';
+import * as api from '../actions/apiActionCreators';
 import 'rxjs/add/operator/do';
 import 'rxjs/add/operator/map';
 import 'rxjs/add/operator/filter';
-import 'rxjs/add/operator/mergeMap';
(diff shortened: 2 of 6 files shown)


src/epics/mutate.js
```diff
@@ -104,8 +47,14 @@ https://github.com/meetup/meetup-web-platform/blob/master/docs/Queries.md#recipe
 				);
 			}
 		})
-		.map(({ payload }) => payload)
-		.mergeMap(doFetch$(getMethodQueryFetch(method, fetchQueries, store)));
+		.map(({ payload: { query, onSuccess, onError } }) => ({
```
mmcgahan: this is the core of the refactor - instead of making the `fetch` call and parsing the results directly in this middleware, we leave `sync` to make the calls. There is a corresponding update in `sync` to call the custom `onSuccess`/`onError` handlers at the right time.

src/epics/sync.js
```diff
@@ -116,21 +116,33 @@ export const getFetchQueriesEpic = fetchQueriesFn => (action$, store) =>
...
 			.takeUntil(action$.ofType(LOCATION_CHANGE)) // cancel this fetch when nav happens
 			.mergeMap(({ successes = [], errors = [] }) => {
+				const deprecatedSuccessPayload = getDeprecatedSuccessPayload(
+					successes,
+					errors
+				);
+				const deprecatedActions = [apiSuccess(deprecatedSuccessPayload)];
+				if (meta && meta.onSuccess) {
+					deprecatedActions.push(meta.onSuccess(deprecatedSuccessPayload));
+				}
```
mmcgahan: new hook to inject `onSuccess` into emitted actions.

src/epics/mutate.js
```diff
@@ -37,65 +29,16 @@ import { getDeprecatedSuccessPayload } from '../util/fetchUtils';
...
-		.catch(err => {
-			const actions = [api.fail(err), apiError(err)];
-			if (onError) {
-				actions.push(onError(err));
-			}
-			return Observable.from(actions);
-		});
-
-const getMethodEpic = method => fetchQueries => (action$, store) =>
+const getMethodEpic = method => action$ =>
```
willh-meetup: Really like how much code we're deleting here!

Maybe it's worth adding some docs for `getMethodEpic` now?

src/actions/apiActionCreators.js
```diff
@@ -21,12 +21,12 @@ export function requestAll(queries, meta) {
 	};
 }
 
-export const _applyMethod = method => query => {
+export const _applyMethod = method => (query, meta) => {
```
willh-meetup: We probably don't need to export this?

src/epics/mutate.js
mmcgahan: done - this whole module is deprecated so hopefully no one will actually ever have to _read_ the docs, though :)

willh-meetup: 👍

PR #270, opened 2017-05-04
don't sort lang list

mattkime:
this started with a failing integration test in mup-web after adding a couple of languages -

```
 FAIL  tests/integration/routing.int.js
  ● Startup and render - using current /build/ files › returns a 404 for invalid route

    expect(received).toBe(expected)

    Expected value to be (using ===):
      404
    Received:
      302

      at responseTest (tests/integration/routing.int.js:40:33)
      at process._tickCallback (internal/process/next_tick.js:109:7)

  ● Startup and render - using current /build/ files › renders the group home page

    expect(received).toBe(expected)

    Expected value to be (using ===):
      200
    Received:
      302

      at responseTest (tests/integration/routing.int.js:49:33)
      at process._tickCallback (internal/process/next_tick.js:109:7)

  Startup and render - using current /build/ files
    ✕ returns a 404 for invalid route (100ms)
    ✕ renders the group home page (31ms)

Test Suites: 1 failed, 1 total
Tests:       2 failed, 2 total
```

for some reason `it-IT` is being selected as the language which i suspect is causing the redirect.

it looks like this is related to language sorting. i'm not sure why this is being done or what the desired effect is but its getting in the way of adding new languages. it appears that, by chance, with the lang list currently in master `en-US` sorts to the top and tests pass. if i add a couple (2) more languages to the list then `it-IT` is at the top of the list. 

Perhaps the solution is to document the language list and make clear the implications of language order.

before adding more languages -

lang list - 
```
    [ 'en-US',
      'en-AU',
      'de-DE',
      'fr-FR',
      'ja-JP',
      'it-IT',
      'pt-BR',
      'es-ES',
      'es' ]
```
sorted lang list - 
```
    [ 'en-US',
      'es',
      'es-ES',
      'pt-BR',
      'it-IT',
      'ja-JP',
      'fr-FR',
      'de-DE',
      'en-AU' ]
```
After adding new languages - 

lang list -
```
    [ 'en-US',
      'en-AU',
      'de-DE',
      'fr-FR',
      'ja-JP',
      'it-IT',
      'pt-BR',
      'es-ES',
      'es',
      'ko-KR',
      'tr-TR' ]
```

sorted lang list - 
```
[ 'it-IT',
      'en-US',
      'tr-TR',
      'ko-KR',
      'es',
      'es-ES',
      'pt-BR',
      'en-AU',
      'ja-JP',
      'fr-FR',
      'de-DE' ]
```

### DIFF
--- a/src/util/languageUtils.js
+++ b/src/util/languageUtils.js
@@ -39,9 +39,9 @@ export const getLanguage = (
 	supportedLangs,
 	defaultLang = LANG_DEFAULT
 ) => {
+	// places defaultLang at top of list
+	supportedLangs.sort(l => l === defaultLang ? -1 : 1);
 	// return the first language hit in the order of preference
-	// TODO - better comment
-	supportedLangs.unshift(defaultLang);
 	return (
 		getCookieLang(request, supportedLangs) ||
 		getUrlLang(request, supportedLangs) ||
--- a/src/util/languageUtils.js
+++ b/src/util/languageUtils.js
@@ -38,16 +38,10 @@ export const getLanguage = (
 	request,
 	supportedLangs,
 	defaultLang = LANG_DEFAULT
-) => {
-	// return the first language hit in the order of preference
-	supportedLangs.sort(l => l !== defaultLang);
-	return (
-		getCookieLang(request, supportedLangs) ||
+) => getCookieLang(request, supportedLangs) ||
 		getUrlLang(request, supportedLangs) ||
 		getBrowserLang(request, supportedLangs) ||
-		defaultLang
-	);
-};
+		defaultLang;
 
 const makeRedirect = (reply, originalUrl) => redirectPathname =>
 	reply.redirect(url.format({ ...originalUrl, pathname: redirectPathname }));
--- a/src/util/languageUtils.js
+++ b/src/util/languageUtils.js
@@ -38,10 +38,17 @@ export const getLanguage = (
 	request,
 	supportedLangs,
 	defaultLang = LANG_DEFAULT
-) => getCookieLang(request, supportedLangs) ||
+) => {
+	// return the first language hit in the order of preference
+	// TODO - better comment
+	supportedLangs.unshift(defaultLang);
+	return (
+		getCookieLang(request, supportedLangs) ||
 		getUrlLang(request, supportedLangs) ||
 		getBrowserLang(request, supportedLangs) ||
-		defaultLang;
+		defaultLang
+	);
+};
 
 const makeRedirect = (reply, originalUrl) => redirectPathname =>
 	reply.redirect(url.format({ ...originalUrl, pathname: redirectPathname }));
--- a/package.json
+++ b/package.json
@@ -12,10 +12,11 @@
     "coveralls": "cat coverage/lcov.info | coveralls",
     "flow": "flow",
     "lint": "eslint . --ext .js,.jsx --fix",
+    "precommit": "lint-staged",
     "test": "yarn run flow && yarn run test:unit && yarn run test:integration -- --silent",
     "test:base": "jest --config jest.config.json",
     "test:integration": "NODE_ENV=test jest --config jest.int.config.json",
-    "test:unit": "NODE_ENV=test yarn test:base",
+    "test:unit": "NODE_ENV=test yarn test:base -- --silent --coverage",
     "test:watch": "yarn test:base -- --watch"
   },
   "repository": {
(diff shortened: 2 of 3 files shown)


mattkime:
looks like i need to mind the project integration tests

src/util/languageUtils.js
```diff
@@ -38,16 +38,10 @@ export const getLanguage = (
 	request,
 	supportedLangs,
 	defaultLang = LANG_DEFAULT
```
sadafie: what's your defaultLang? `it-IT`?

src/util/languageUtils.js
```diff
@@ -38,16 +38,10 @@ export const getLanguage = (
...
-) => {
-	// return the first language hit in the order of preference
-	supportedLangs.sort(l => l !== defaultLang);
```
mmcgahan: Looks like I just needed to review the `array.sort` API - this sorting function should be

```js
supportedLangs.sort(l => l === defaultLang ? -1 : 1);
```

As the deleted comment suggests, the `supportedLangs` are being sorted by their order of preference. The only preference we have is that the default language take precedence over all others, so the sort function is intended to guarantee that it appears at the top of the list rather than assuming that the list will be manually defined in order of preference.

You should get the correct behavior just by adjusting the sort function

mattkime:
I'm not entirely certain about the role that defaultLang plays

src/util/languageUtils.js
mattkime: any reason not to just `unshift(defaultLang)`?

mmcgahan: you could, but then you have a redundant `defaultLang` key somewhere deeper in the array that isn't doing anything.

given that the 'verbose' alternative is `supportedLangs.sort(l => l === defaultLang ? -1 : 1);`, I'd prefer 'correct' over terse.

mmcgahan:
`defaultLang` is selected when the accept-language header and/or cookie request a language other than the ones supported, or don't specify a language at all.

src/util/languageUtils.js
```diff
@@ -40,7 +40,8 @@ export const getLanguage = (
 	defaultLang = LANG_DEFAULT
 ) => {
 	// return the first language hit in the order of preference
-	supportedLangs.sort(l => l !== defaultLang);
+	// TODO - better comment
```
mmcgahan: let's go ahead and flesh out the comment in this PR if it needs an update

src/util/languageUtils.js
sadafie: Ah I see. I was reading the sort API but not really following what this sort was intended to do. I agree that a comment update would be helpful here as well to explain the sort and `defaultLang`

mattkime: >given that the 'verbose' alternative is supportedLangs.sort(l => l === defaultLang ? -1 : 1);, I'd prefer 'correct' over terse.

but your 'correct' version also unnecessarily rearranges the rest of the list which is extremely confusing unless its been directly addressed. It appears to be a side effect but I only could have guessed...and hoped that the test suite complete.

src/util/languageUtils.js
mattkime: >defaultLang is selected when the accept-language header and/or cookie request a language other than the ones supported

Is `defaultLang` always `en-US`?  It looks like it can be overridden but i'm unclear about the circumstances in which it happens.

src/util/languageUtils.js
mmcgahan: this is definitely a good place to insert any other sorting rules that are relevant, but at the moment I think we only care that the `defaultLang` is first in the priority list? We could break it out into a function called `sortByPreference` or something if we need to be more specific.

> Perhaps the solution is to document the language list and make clear the implications of language order.

Your original suggestion to document the 'supported languages' setting and make it the responsibility of the caller to submit them in preference order might be the best long-term option. It's a slightly larger refactor, though, so maybe the best short term resolution is to fix the sort function definition and open a ticket to require that `supportedLangs` be provided in order of language preference, default lang first.

src/util/languageUtils.js
mmcgahan: yeah, for our apps `defaultLang` is always `en-US`, but the option of passing in a different default language makes this function more generic, and if I remember right the unit tests supply different default languages to test the behavior.

Eventually, when Meetup moves HQ to New Zealand, we'll want to update this to `en-NZ`, but it's not critical.

mattkime:
Incorporated requested changes and created new ticket as requested. I'm going to go ahead and merge this. if any changes are needed i'm happy to discuss further. need to support additional languages for onboarding launch.

PR #272, opened 2017-05-08
WP 302 GAE PubSub tracking

mmcgahan:
In order to get the tracking logs to the right place in GAE, we have to use Google PubSub. This PR looks for a GAE-supplied env variable to choose between regular stdout logging and PubSub logging.

I can't really test it outside of GAE, so the plan will be to

1. merge this branch
2. pull it into the starter kit
3. pull it into mup-web
4. update mup-web's `Makefile` to send the `GAE_INSTANCE` variable to the GAE `docker run`
5. Look for errors in the GAE log, and hopefully be able to check in with the data team to confirm that the tracking data is appearing where they expect GAE tracking data to appear

### DIFF
--- a/src/util/avro.js
+++ b/src/util/avro.js
@@ -1,4 +1,13 @@
+const log = require('./logger').default;
+const MOCK_PUBSUB = {
+	topic: () => ({ publish: () => Promise.resolve() }),
+};
+const pubsub = process.env.GAE === 'true'
+	? require('@google-cloud/pubsub')()
+	: MOCK_PUBSUB;
 const avro = require('avsc');
+
+const analyticsLog = pubsub.topic('analytics-log-json');
 
 // currently the schema is manually copied from
 // https://github.dev.meetup.com/meetup/meetup/blob/master/modules/base/src/main/versioned_avro/Click_v2.avsc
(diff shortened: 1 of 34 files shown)


src/util/avro.js
```diff
@@ -1,5 +1,27 @@
+const log = require('./logger').default;
 const avro = require('avsc');
 
+const getPlatformAnalyticsLog = (isGAE = process.env.GAE_INSTANCE) => {
```
mattkime: is there a way to pass an argument instead of using an env var? ideally config is managed by the app.

mmcgahan: we could translate the env var into a config var in a separate file, but this seems like a clear case of an environment dependency as opposed to a 'local configuration' dependency - the environment itself must be able to support the automatic authentication necessary to make pubsub calls, and the env var seems like a good way of indicating that.

mattkime: oh right, just add a comment about it.

mmcgahan:
I believe this is ready for final review - I was able to confirm that the tracking messages can be successfully sent through pubsub from a local machine that is authorized (as described in docs), but we'll need to set up mup-web to pass in the `GAE` env var to the Docker container and see whether the messages are correctly sent there when running in GAE - there's not a definitive way of proving performance outside of deployment.

Followups can be done in separate tickets - current behavior is unaffected by this new feature

mattkime:
I should have let @mmcgahan merge. i'm a bit out of it at the moment. should be okay though.

PR #276, opened 2017-05-11
WP 324 separate buildtime config from runtime config

mmcgahan:
This PR separates out the buildtime config from the runtime config so that Travis can build the app in `NODE_ENV=production` but still run all 'simulated runtime' tests in `NODE_ENV=test`

In the mup-web Node Convict PR (https://github.com/meetup/mup-web/pull/999), we'll need to update the `config` imports in the build scripts so that they are importing from `meetup-web-platform/lib/util/config/build` instead of `meetup-web-platform/lib/util/config`. This should allow Webpack to correctly transpile the app.

The dummy testing runtime in Travis will be handled by `config.test.json`.

The prod runtime in GCP will be handled by env vars.

### DIFF
--- a/src/util/config/build.test.js
+++ b/src/util/config/build.test.js
@@ -1,0 +1,12 @@
+import { PROTOCOL_ERROR, validateProtocol } from './build';
+
+describe('validateProtocol', () => {
+	it('does not error when protocol is `http` or `https`', () => {
+		expect(() => validateProtocol('http')).not.toThrow();
+		expect(() => validateProtocol('https')).not.toThrow();
+	});
+
+	it('throws error when the protocol is not `http` or `https`', () => {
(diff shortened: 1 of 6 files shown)


src/util/config/build.test.js
```diff
@@ -0,0 +1,12 @@
+import { PROTOCOL_ERROR, validateProtocol } from './build';
+
+describe('validateProtocol', () => {
+	it('does not error when protocol is `http` or `https`', () => {
+		expect(() => validateProtocol('http')).not.toThrow();
+		expect(() => validateProtocol('https')).not.toThrow();
+	});
+
+	it('throws error when the protocol is not `http` or `https`', () => {
+		expect(() => validateProtocol('ftp')).toThrowError(PROTOCOL_ERROR);
```
willh-meetup: Heh

PR #279, opened 2017-05-17
Flow typing

mmcgahan:
I wanted to try out the Atom+Nuclide editor to see how the Flow integration worked, and aim for 100% Flow coverage on `routeUtils.js` - this PR is the result.

I learned a lot about Flow, but there are no significant functional changes - mainly just safer code!


flow-typed/platform.js
```diff
@@ -1,7 +1,8 @@
 // @flow
-import type { Match } from 'react-router-dom';
```
mmcgahan: turns out you can't currently import a libdef into a libdef - this will just default to an `any` type

src/util/routeUtils.js
```diff
@@ -9,10 +9,16 @@ import matchPath from 'react-router-dom/matchPath';
...
 
 // munge a route's 'relative' `path` with the full matchedPath
-const routePath = (route: PlatformRoute, matchedPath: string): string =>
-	`${matchedPath}${route.path || ''}`.replace('//', '/');
-
-export const decodeParams = (params: Object) =>
+const routePath = (
+	route: PlatformRoute,
+	matchedPath: string
+): MatchPathOptions => ({
```
mmcgahan: although technically `matchPath` can take a string arg, it will convert it into the preferred `MatchPathOptions` object, so I've refactored this function to return a `MatchPathOptions` object directly

PR #280, opened 2017-05-18
Update config docs

sadafie:
Ran into a few issues with my environment this morning so wanted to clarify a few things in the docs, including precedence and adding another local config file example.

I plan on linking to the `Config.md` file in this repo from the starter repo, since it's not easily discoverable otherwise (unless you know to look for it).

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -15,7 +15,6 @@ Configuration values are read in the following order of increasing precedence:
 1. Default value - see `src/util/config/index.js` and `src/util/config/build.js`
 2. Application root `config.<NODE_ENV>.json` file
 3. Environment variable in `$HOME/.mupweb.config`
-4. Command line arguments
 
 **Note** that if a configuration variable is set in _BOTH_ a local config file AND
 in an environment var, the environment variable will take precedence.
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -32,7 +32,7 @@ with the format `config.<target NODE_ENV>.json`, e.g.
 The schema for all supported values is in `../src/util/config.js`.
 
 
-## Example local config files
+## Examples
 
 ### Sample `config.test.json` file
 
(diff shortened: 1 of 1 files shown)


sadafie:
@mmcgahan is this good to merge now?

docs/Config.md
```diff
@@ -1,35 +1,47 @@
...
+## Configuration precedence
 
-1. The OAuth consumer `key` for the platform
-2. The OAuth consumer `secret`
-3. The photo scaler salt for Meetup Classic
+Configuration values are read in the following order of increasing precedence:
+
+1. Default value - see `src/util/config/index.js` and `src/util/config/build.js`
+2. Application root `config.<NODE_ENV>.json` file
+3. Environment variable in `$HOME/.mupweb.config`
```
mmcgahan: Technically it doesn't matter where the env var comes from - not sure if it would be confusing to people if `.mupweb.config` wasn't automatically read, or conversely if they had set a relevant env var in another way that wasn't in `.mupweb.config`. Not a big deal.

docs/Config.md
```diff
@@ -1,35 +1,47 @@
...
+4. Command line arguments
```
mmcgahan: currently we haven't specified any command line args, so maybe this is confusing?

sadafie: i had the same thought, will remove for now

sadafie:
>  although one thing that I'd like to discuss is whether we should explicitly support env vars in development at all, i.e. we should encourage people to maintain a local config.development.json file that contains the secrets so that we never have to guide them through the env var setup at all.

Agreed, +1. I think the local config.developmet.json file would cause less potential confusion.

docs/Config.md
sadafie: I'd like to leave this is for now, since "historically" the configs were environment variables, and if they're still set in `.mupweb.config`, notes like this one might help debug issues.

PR #282, opened 2017-05-19
WP-318: activity tracking as Hapi plugin

mmcgahan:
This is a big PR because it involves moving + modifying code, but the modifications are _mostly_ related to adding Flow types to the activity tracking code - it's a bit verbose, which I'm not thrilled about, but the code is "better" for having types.

Two major functional changes

1. Tracking methods on the `reply` interface are now provided by a plugin, rather than a manually-applied `server.decorate` function in the startup script.
2. instead of `reply.track(response, <tracking type>, ...args)`, there are now specific `reply` methods for each tracking type:

    - `reply.trackApi`
    - `reply.trackSession`

I'll add a few more docs to try to make this plugin a little more clear, but I wanted to open this up for review early.

Although this ticket was initially going to also include a refactor of click tracking to be a plugin, that should be its own project because of how it is woven into the Redux system in addition to the server - it will probably require a more in-depth refactor to make it more modular, and potentially (although not certainly) pluggable.


src/plugins.js
```diff
@@ -1,8 +1,10 @@
+import config from './util/config';
```
mmcgahan: config returns to the plugins script - mainly to make sure the tracking plugin is not dependent on global config (it only needs 1 `isProd` property)

package.json
```diff
@@ -100,7 +100,7 @@
     "tough-cookie": "2.3.2"
   },
   "lint-staged": {
-    "*.{js,jsx}:-flow-typed/**/*": [
+    "src/**/*.{js,jsx}": [
```
sadafie: how about `tests/`?

mmcgahan: Yeah I was trying to get this to work with multiple directories, but the minimatch glob pattern (https://github.com/isaacs/minimatch) is surprisingly cumbersome and I couldn't get it to do multiple directories _and_ multiple file extensions. I've opened a ticket to address this directly https://meetup.atlassian.net/browse/WP-340

PR #283, opened 2017-05-20
WP 305 integration tests: PATCH, DELETE /mu_api

mmcgahan:
These supported request methods were not being tested - this allowed a break to happen awhile back and I promised I would add relevant unit tests.

This also consolidates our /mu_api endpoint integration tests by using a configurable `request` module mock, similar to what is used in mup-web

There is scope for more detailed testing, but the main objective here is coverage so that we know that there is basic request method support


mmcgahan:
@nlaz feel free to pass the review to someone else, but thought you might be interested

tests/integration/api.int.js
```diff
@@ -90,3 +114,143 @@ describe('API proxy endpoint integration tests', () => {
...
+		})
+		.then(test)
+		.then(() => server.stop())
+		.catch(err => {
+			server.stop();
+			throw err;
+		});
+};
+
+describe('API proxy POST endpoint integration tests', () => {
```
eilinora: General comment these look all like positive outcome tests, should we have negative tests. This is a bit different than unit tests but just commenting on that observation

mmcgahan: Yeah, in this case I think that's alright because I don't _think_ we actually care about cases where the handlers may or may not be called - we only care that they _are_ called for the specific request methods that we've defined. For example, I don't think the tests should fail if the POST handler ended up delegating to the GET handler, so we don't need to test that the GET handler _is not called_ in the POST case. Essentially, I'd argue that the handlers are not designed to have an opinion on incorrect/irrelevant requests, so it's not necessary to verify the behavior of incorrect/irrelevant requests. Let me know if you have a suggestion on what to add, though.

eilinora: Makes sense 🚀

PR #285, opened 2017-05-25
configure `supportedLocaleCodes` with consolidated CONFIGURE action

mmcgahan:
The main change is adding `supportedLocaleCodes` to `state.config` using the same Redux action pattern we've used for passing other config/environment data from the server to the client app.

I've also refactored the config action creators as a single `CONFIGURE` action creator that produces a single immutable object update to `state.config` since everything in `state.config` is a simple string-value map.

### DIFF
--- a/src/actions/configActionCreators.js
+++ b/src/actions/configActionCreators.js
@@ -1,15 +1,5 @@
 // @flow
-export const configure = (config: { [string]: mixed }) => ({
+export default (config: { [string]: mixed }) => ({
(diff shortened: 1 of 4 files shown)


mmcgahan:
> Will this affect anything on the consumer app side

It won't change anything for current app code, but check out https://meetup.atlassian.net/browse/WP-123 and the comments to see where the intended application is - group context 'canonical link' tags, which I think should be done by a group-context-related product team.

PR #286, opened 2017-05-25
WP 336 - clean up build config and make sure we're validating the right things

mmcgahan:
My original pass at hostname validation focused on the wrong thing - app server and asset server host names, which don't actually matter in prod vs dev (app server host is always locked to 0.0.0.0 in the code, actually, so it doesn't need to be part of the config).

Instead, we need to validate that the prod _API host_ is not used in dev, and vice versa.

This PR also moves the app_server and asset_server configs out of config/build.js and into config/index.js, because they're not actually needed at build time.

The result is simpler and more correct. Win.


src/util/config/build.js
```diff
@@ -84,23 +28,6 @@ export const schema = {
 };
 export const config = convict(schema);
 
-// Load environment dependent configuration
```
sadafie: We don't need a local config for `env, `isProd`, `isDev`, but removing this means that we can't change the value of `disable_hmr` with a local config file. Shouldn't we be able to?

src/util/config/build.js
```diff
@@ -84,23 +28,6 @@ export const schema = {
...
-const configPath = path.resolve(
```
sadafie: We don't need a local config for `env, `isProd`, `isDev`, but removing this means that we can't change the value of `disable_hmr` with a local config file. Shouldn't we be able to?

mmcgahan: I thought briefly about that, but was thinking that `disable_hmr` might generally be something you want to temporarily opt into rather than 'hard coding' it into a local config file, so maybe it would be simplest just to leave it as an env variable - the hoops we have to jump through to read a local config file are a little cumbersome for what IMO is an edge case usage. The original desire for `DISABLE_HMR` came from troubles with the old implementation of HMR, so I'm not sure if it will even be used these days...

I can add back support if it seems like it's too inconsistent or awkward to use.

sadafie: My vote would be to remove it if HMR is pretty reliable now. Otherwise, the documentation that says not to use ENV vars but rather to use local config for everything except `disable_hmr` would be confusing imo.

For the purpose of this PR though I think it's fine as is, and maybe we can follow up remove.

PR #287, opened 2017-05-26
Provide initial now

mmcgahan:
This provides the server-side render time to the client so that it can use a consistent 'relative time' reference on both sides of the divide and avoid client-server markup mismatches (MW-1117 - the fix is provided in a corresponding mup-web PR)

### DIFF
--- a/src/renderers/server-render.jsx
+++ b/src/renderers/server-render.jsx
@@ -184,6 +184,7 @@ const makeRenderer = (
 			baseUrl: host,
 			localeCode,
 			supportedLocaleCodes,
+			initialNow: new Date().getTime(),
 		})
 	);
 	// render skeleton if requested - the store is ready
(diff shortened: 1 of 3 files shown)


src/renderers/server-render.jsx
```diff
@@ -184,10 +178,15 @@ const makeRenderer = (
...
-	store.dispatch(configureBaseUrl(host));
-	store.dispatch(configureLocaleCode(localeCode));
-
+	store.dispatch(
+		configure({
+			apiUrl,
+			baseUrl: host,
+			localeCode,
+			supportedLocaleCodes,
+			initialNow: new Date().getTime(),
```
sadafie: Is there a test case that could be updated to at least check for this value, and possibly test whatever date format this is returning?

mmcgahan: Turns out we don't have a unit test for this module, so I've opened a ticket to write that -
 we can make sure `initialNow` is one of the expected props of this function call there https://meetup.atlassian.net/browse/WP-341

Issue #281, opened 2017-05-18
MWP modules


mmcgahan:
As this repo has stabilized, its underlying structure has become better defined, but the current module structure makes it difficult to work with features in isolation, and the scope of particular updates is initially difficult to determine because it's hard to get a sense of how many files might need to be touched.

I'd like to change that by restructuring the platform to be more clearly modular, with explicit entry points for consumer applications rather than the 'export all the things' setup we currently have. The module organization would also aim to have as few dependencies between modules as possible, so that updates can be confined to individual modules - in effect, it would be multiple independent 'packages' in a single repo.

This work would follow on from refactoring the server-side modules as plugins (outline copied over from [this Google Doc](https://docs.google.com/document/d/1o14_iO1Ak_QksizmKRI2KYv-IWB_Antg_wFCYBjgYNU/edit)):

### `server` (Hapi)
- [x] startup utils - https://meetup.atlassian.net/browse/WP-417

#### Hapi Plugins

- [x] ~~tracking - https://meetup.atlassian.net/browse/WP-318~~
- [x] ~~api - https://meetup.atlassian.net/browse/WP-319~~
- [x] react app renderer - https://meetup.atlassian.net/browse/WP-320
- [x] ~~service worker - wait for https://meetup.atlassian.net/browse/WP-321~~
- [x] click tracking - this interacts with Redux, fetchUtils, and the server routes - need to figure out how to make it a more coherent module

### ~~`router`~~
- [x] ~~browser: React Router helpers - https://meetup.atlassian.net/browse/WP-337~~

### `render` (React+Redux helpers)
- https://meetup.atlassian.net/browse/WP-416
- [x] `server`
  - ServerApp (TBD)
  - `dom.jsx`
  - `createServerStore`
- [x] `browser`
  - BrowserApp
  - `createBrowserStore`
- [x] `core`
  - `<PlatformApp />`
  - [ ] middleware/epics
  - [ ] platform reducer
  - [ ] `fetchUtils`

### ~~(optional) MWP build/run CLI~~
- [x] ~~described in Google doc~~

mattkime:
nothing to add except that it looks like a good plan.

eilinora:
@mmcgahan, Are there any down the chain changes that teams next be aware of, thinking of usages in Pro and MUP repo's?

mmcgahan:
[...]

PR #289, opened 2017-06-01
WP 331 `<Redirect>` component

mmcgahan:
This PR introduces a `<Redirect>` component that behaves similarly to [React Router's `<Redirect>`](https://reacttraining.com/react-router/web/api/Redirect) but includes support for redirecting to external URLs.

Docs have been added to `Routing.md` and the component itself, with full Flow type annotations. Unit tests and integration tests have been added.


mattkime:
likely out of scope, but could something like this be its own package?

mmcgahan:
@mattkime definitely - I think it will go nicely with the other Router-related tools into a proper `routing` module/package (mentioned in the [MWP modules issue](https://github.com/meetup/meetup-web-platform/issues/281)), and it would likely make sense to release it as a standalone repo/package at that point

.eslintrc.js
```diff
@@ -34,15 +34,6 @@ const config = {
 		'no-useless-escape': 2,
 		'no-whitespace-before-property': 2,
 		'prefer-template': 2,
-		'require-jsdoc': [
```
mmcgahan: At this point, we would prefer Flow over jsdoc for type annotations

package.json
```diff
@@ -58,15 +58,6 @@
     "url-regex": "3.2.0",
     "uuid": "3.0.1"
   },
-  "peerDependencies": {
```
eilinora: are these supposed to all be removed?

mmcgahan: nope! had to remove them temporarily in order to run `flow-typed` but forgot to put them back, thanks!

flow-typed/platform.js
```diff
@@ -71,3 +71,20 @@ declare module 'react-router-dom/matchPath' {
 	declare function matchPath(pathname: string, options: MatchPathOptions): null | Match;
 	declare module.exports: typeof matchPath;
 }
+
+declare type LocationShape = {
+  pathname?: string,
```
eilinora: more a flow question what does the `?` mean?

mmcgahan: optional prop. Any key that is not explicitly mentioned in the type definition will throw an error

src/components/redirect.test.jsx
```diff
@@ -0,0 +1,52 @@
...
+	ReactDOMServer.renderToString(
+		<StaticRouter location={new URL('http://example.com')} context={context}>
+			<Redirect to={to} />
+		</StaticRouter>
+	);
+	return context;
+};
+
+describe('Server rendering', () => {
+	it('sets React Router context for "local" redirect, nothing from Redirect.rewind()', () => {
```
eilinora: is this because of withSideEffect?

src/components/redirect.test.jsx
```diff
@@ -0,0 +1,52 @@
...
+		const internalTo = 'foo/bar';
+		const httpContext = renderToContext(internalTo);
+		expect(httpContext.url).toEqual(internalTo);
+		expect(Redirect.rewind()).toBeUndefined();
+	});
+	it('provides external redirect URL from Redirect.rewind(), nothing from React Router context', () => {
```
eilinora: nitpick alert! might be good to split this and the next `it` into separate tests

src/components/redirect.test.jsx
mmcgahan: sort of. `Redirect.rewind()` is provided by `withSideEffect`, and should only return a value when the underlying React Router `<Redirect>` can't handle the URL - i.e. `rewind()` should only return a value when the Redirect is to an external URL. I'll add another inline comment [here](https://github.com/meetup/meetup-web-platform/pull/289/files/892b6732a97a369d4a6f005d8f7dd98f8af9b22e#diff-d88069df2a21944fc5327a27969009f7R65) to make it more explicit

src/components/redirect.test.jsx
mmcgahan: You mean separate the https and http urls? I feel like that's not central to the behavior of the method, but I can split it out

src/components/Redirect.jsx
```diff
@@ -0,0 +1,80 @@
...
+type RouterTo = string | LocationShape | URL;
+type RedirectProps = {
+	to: RouterTo,
+	push?: boolean,
+};
+
+/*
+ * A routing component that, when rendered, will redirect the application to
+ * another internal route _or_ external URL. On the server, the redirect will
+ * be handled with a 302 redirect HTTP response to the browser. On the client,
```
eilinora: What if we need to do a 301? I'm soon going to be migrating a page that will live at 1 location to a new one, ideally i would use the Redirect to send people on their way but would be good to have it recorded as 301 so SEO picks that up

mmcgahan: I started adding in that functionality this morning but it requires some non-trivial refactoring that I think would be easier to implement and review in a separate ticket - I'll do it today, but I'll make a new PR to keep this one simple and make it easier to see the code specifically relevant to 301/302 support

src/components/redirect.test.jsx
mmcgahan: done

src/components/Redirect.jsx
eilinora: Sounds good! Thanks Mike!

PR #292, opened 2017-06-02
WP 265 clear inflight refs

mmcgahan:
This PR updates the platform reducer to clear out data old that is being refreshed by a new query.

This will prevent flashes of content from prior navigation on things like group or event pages when moving around the app. For repeated queries, the cache middleware will ensure that the UI get 'instant' updates of the cleared state.

### DIFF
--- a/src/reducers/platform.js
+++ b/src/reducers/platform.js
@@ -27,6 +27,15 @@ type ApiState = {
 export const DEFAULT_API_STATE: ApiState = { inFlight: [] };
 export const DEFAULT_APP_STATE = {};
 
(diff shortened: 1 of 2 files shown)


src/reducers/platform.js
```diff
@@ -50,7 +62,13 @@ export function api(
 				return { ...DEFAULT_API_STATE, inFlight };
 			}
 
-			return { ...state, inFlight };
+			// remove any `ref`s that are being refreshed - eliminate stale data
```
eilinora: Are we also clearing inFlight when errors occur, i _think_ i observed that not to be the case

PR #294, opened 2017-06-02
revert the catch-all unicode delete, convert all unicode newlines

mmcgahan:
it's only those unicode newlines that are a problem. other unicode is (hopefully) safe. This PR catches the unicode newlines when they first appear in API responses.


mmcgahan:
I tested this on http://beta2.dev.meetup.com:8000/TimerliitXolos/events/235131434/ and the original issue is still resolved

PR #295, opened 2017-06-05
Mw 586 ssl in dev

willh-meetup:
This gets SSL working in dev with Hapi. Also need to make some small changes to webpack dev server in mup-web to get it reading in the right config options into the right place.

Gonna leave some comments below on some choices I made that I could go either way on.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -101,7 +101,7 @@ need to add these values to both the `asset_server` and the `app_server`.
   "csrf_secret": "asdfasdfasdfasdfasdfasdfasdfasdf",
   "cookie_encrypt_secret": "asdfasdfasdfasdfasdfasdfasdfasdf",
   "app_server": {
-    "protocol": "http",
+    "protocol": "https",
     "host": "beta2.dev.meetup.com",
     "port": 0,
     "crt_file":"/path/to/certificate.crt",
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -75,3 +75,45 @@ The secrets in this example are dummy values and won't work in an actual app.
   }
 }
 ```
+
+### Running over HTTPS
+
+You need to change change the `protocol` to `https`, and tell the server where
+to find the certificate and the private key (`crt_file` and `key_file`). You'll
+need to add these values to both the `asset_server` and the `app_server`.
+
+```json
+{
+  "api": {
+    "protocol": "https",
+    "host": "www.api.dev.meetup.com",
+    "timeout": 10,
+    "root_url": "https://www.api.dev.meetup.com"
+  },
+  "asset_server": {
+    "protocol": "https",
+    "host": "beta2.dev.meetup.com",
+    "port": 0,
+    "crt_file":"/path/to/certificate.crt",
+    "key_file":"/path/to/certificate.key",
+
+  },
+  "csrf_secret": "asdfasdfasdfasdfasdfasdfasdfasdf",
+  "cookie_encrypt_secret": "asdfasdfasdfasdfasdfasdfasdfasdf",
+  "app_server": {
+    "protocol": "http",
+    "host": "beta2.dev.meetup.com",
+    "port": 0,
+    "crt_file":"/path/to/certificate.crt",
+    "key_file":"/path/to/certificate.key",
+  },
+  "disable_hmr": false,
+  "photo_scaler_salt": "asdfasdfasdfasdfasdfasdfasdfasdf",
+  "oauth": {
+    "auth_url": "https://secure.dev.meetup.com/oauth2/authorize",
+    "access_url": "https://secure.dev.meetup.com/oauth2/access",
+    "key": "asdfasdfasdfasdfasdfasdfasdfasdf",
+    "secret": "asdfasdfasdfasdfasdfasdfasdfasdf"
+  }
+}
+```
--- a/src/server.js
+++ b/src/server.js
@@ -37,22 +37,10 @@ export default function start(
 
 	const baseRoutes = getRoutes(renderRequestMap);
 	const finalRoutes = [...routes, ...baseRoutes];
-	let tls = null;
-	if (
-		appConfig.app_server.protocol === 'https' &&
-		appConfig.app_server.key_file &&
-		appConfig.app_server.crt_file
-	) {
-		tls = {
-			key: fs.readFileSync(appConfig.app_server.key_file),
-			cert: fs.readFileSync(appConfig.app_server.crt_file),
-		};
-	}
 
 	const connection = {
 		host: '0.0.0.0',
 		port: appConfig.app_server.port,
-		tls: tls,
 		routes: {
 			plugins: {
 				'electrode-csrf-jwt': {
@@ -61,6 +49,13 @@ export default function start(
 			},
 		},
 	};
+
+	if (appConfig.app_server.protocol === 'https') {
+		connection.tls = {
+			key: fs.readFileSync(appConfig.app_server.key_file),
+			cert: fs.readFileSync(appConfig.app_server.crt_file),
+		};
+	}
 
 	const finalPlugins = [...plugins, ...getPlugins()];
 
--- a/src/server.js
+++ b/src/server.js
@@ -1,5 +1,7 @@
 import './util/globals';
 
+import fs from 'fs';
+
 import appConfig from './util/config';
 import getPlugins from './plugins';
 import getRoutes from './routes';
@@ -35,10 +37,22 @@ export default function start(
 
 	const baseRoutes = getRoutes(renderRequestMap);
 	const finalRoutes = [...routes, ...baseRoutes];
+	let tls = null;
+	if (
+		appConfig.app_server.protocol === 'https' &&
+		appConfig.app_server.key_file &&
+		appConfig.app_server.crt_file
+	) {
+		tls = {
+			key: fs.readFileSync(appConfig.app_server.key_file),
+			cert: fs.readFileSync(appConfig.app_server.crt_file),
+		};
+	}
 
 	const connection = {
 		host: '0.0.0.0',
 		port: appConfig.app_server.port,
+		tls: tls,
 		routes: {
 			plugins: {
 				'electrode-csrf-jwt': {
--- a/src/util/config/build.js
+++ b/src/util/config/build.js
@@ -33,12 +33,12 @@ export const schema = {
 		key_file: {
 			format: String,
 			default: '',
-			env: 'KEY_FILE',
+			env: 'ASSET_KEY_FILE',
 		},
 		crt_file: {
 			format: String,
 			default: '',
-			env: 'CRT_FILE',
+			env: 'ASSET_CRT_FILE',
 		},
 	},
 	env: {
@@ -76,6 +76,16 @@ if (asset_server) {
 
 config.set('isProd', config.get('env') === 'production');
 config.set('isDev', config.get('env') === 'development');
+
+const assetConf = config.get('asset_server');
+
+if (
+	assetConf.protocol === 'https' &&
+	(!fs.existsSync(assetConf.key_file) || !fs.existsSync(assetConf.crt_file))
+) {
+	throw new Error('Missing HTTPS cert or key!');
+}
+
 config.validate();
 
 export default config.getProperties();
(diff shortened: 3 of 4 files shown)


docs/Config.md
```diff
@@ -75,3 +75,45 @@ The secrets in this example are dummy values and won't work in an actual app.
...
+    "protocol": "https",
+    "host": "www.api.dev.meetup.com",
+    "timeout": 10,
+    "root_url": "https://www.api.dev.meetup.com"
+  },
+  "asset_server": {
+    "protocol": "https",
+    "host": "beta2.dev.meetup.com",
+    "port": 0,
+    "crt_file":"/path/to/certificate.crt",
```
willh-meetup: `crt_file` and `key_file` are attached to both `asset_server` and `app_server`. I like that they are in the same place we're setting the protocol, but it's almost always going to be a duplicate config.

docs/Config.md
```diff
@@ -75,3 +75,45 @@ The secrets in this example are dummy values and won't work in an actual app.
   }
 }
 ```
+
+### Running over HTTPS
+
+You need to change change the `protocol` to `https`, and tell the server where
```
willh-meetup: Where you get certs for Meetup is actually in `mup-web`, since that's where the config goes.

src/server.js
```diff
@@ -35,10 +37,22 @@ export default function start(
...
 	const baseRoutes = getRoutes(renderRequestMap);
 	const finalRoutes = [...routes, ...baseRoutes];
+	let tls = null;
+	if (
+		appConfig.app_server.protocol === 'https' &&
+		appConfig.app_server.key_file &&
+		appConfig.app_server.crt_file
+	) {
+		tls = {
+			key: fs.readFileSync(appConfig.app_server.key_file),
```
willh-meetup: I'm reading in these files because they have newlines (that need to be preserved), and that makes them super hard to add to JSON.

docs/Config.md
```diff
@@ -75,3 +75,45 @@ The secrets in this example are dummy values and won't work in an actual app.
...
+    "host": "beta2.dev.meetup.com",
+    "port": 0,
+    "crt_file":"/path/to/certificate.crt",
+    "key_file":"/path/to/certificate.key",
+
+  },
+  "csrf_secret": "asdfasdfasdfasdfasdfasdfasdfasdf",
+  "cookie_encrypt_secret": "asdfasdfasdfasdfasdfasdfasdfasdf",
+  "app_server": {
+    "protocol": "http",
```
mattkime: should this be https? looks like a cery and key are defined.

docs/Config.md
mattkime: When this is set to https, does ordinary http work as well?

mattkime:
if i'm understanding correctly, the `https` config is used only to tell code to check for the cert and key files. if thats the case then (in dev) i'd find it preferable to 

a) check to see if the cert and key are set
b) try/catch around loading the cert and key
c) the above succeed, pass them into the config
d) otherwise log whats missing and that https will not be available.

we do have http2 around the corner so it would be nice to sidestep any confusion over setting the protocol.

...could always be a follow up.

src/util/config/build.js
```diff
@@ -25,6 +25,21 @@ export const schema = {
 			arg: 'asset-port',
 			env: process.env.NODE_ENV !== 'test' && 'ASSET_SERVER_PORT', // don't read env in tests
 		},
+		protocol: {
```
mattkime: i'm confused as to why these are build config changes. ideally this would be in `src/util/config/`, right?

willh-meetup: This is just adding some defaults to the schema. This prevents the build from complaining about these values not being defined when it starts the server.

docs/Config.md
willh-meetup: Yeeeeeep.

docs/Config.md
willh-meetup: Nope. Should I clarify here?

src/util/config/build.js
willh-meetup: Oh, maybe you mean in `index.js`? I think it might make sense to add some validation in there, but I still think it needs to get added to the schema.

willh-meetup:
@mattkime So, `protocol` was already defined in our config (and does control the actual protocol of the URL we're serving), so I'm backing it off that.

You're right, though, that it would be nice to complain more loudly about needing all three.

docs/Config.md
mattkime: That would be helpful but I also wonder if this is something we should address. 

I am lazy. when i type `beta2.meetup.com` i see a broken page. Then i type `https://beta2.meetup.com.` Its kinda annoying.

src/util/config/build.js
mattkime: I think I'm just confused about what goes in which config file although it doesn't need to be settled here.

docs/Config.md
willh-meetup: Just to expand on that: Hapi is an application server and you need to start up multiple version to listen on different ports, so we'd be starting an extra web server that would just be redirecting traffic to the SSL version (I assume we want  redirect, because the whole point of doing this is so that all traffic goes over https).

mattkime: yeah, definitely, and can be left alone for now

mattkime:
A few thoughts that are coming in a bit late - 

What is the value of SSL in dev? 

Initially, it seems the value would be 'we have ssl in prod so we should have it in dev' which makes sense but google provides the ssl layer so this is similar but not the same as prod. 

Is there value in using SSL in prod at the container level? Does GAE even support that? Would there be a performance hit?

We can also think of this ticket as an intermediate step to supporting http2. http2 may or may not provide performance improvements in GAE (perhaps the proxy could reuse connections, etc). Basic http2 push support looks like it will be delayed until browsers can cancel unnecessary pushes. 

---

How does this fit into the dev work flow? Will it be used by default or for troubleshooting an https specific problem?

...anyway, i know this strays pretty far from proper PR discussion but i'm happy to take the conversation elsewhere.

src/server.js
```diff
@@ -35,10 +37,22 @@ export default function start(
 
 	const baseRoutes = getRoutes(renderRequestMap);
 	const finalRoutes = [...routes, ...baseRoutes];
+	let tls = null;
+	if (
+		appConfig.app_server.protocol === 'https' &&
+		appConfig.app_server.key_file &&
+		appConfig.app_server.crt_file
```
mmcgahan: as part of validation, we should make sure that `key_file` and `crt_file` are populated when `protocol` is https - that's a pretty clear error condition. Then, we'll be able to assume that if `protocol === 'https'`, we're good to go.

src/server.js
```diff
@@ -35,10 +37,22 @@ export default function start(
 
 	const baseRoutes = getRoutes(renderRequestMap);
 	const finalRoutes = [...routes, ...baseRoutes];
+	let tls = null;
```
mmcgahan: nitpick, but in general I'd aim to just 'inject' the `connection.tls` property directly inside an `if` block after `connection` is first defined in order to avoid having a re-assignable `let` declaration and to avoid any ambiguity around passing in a `tls` key/val that may or may not be defined.

mmcgahan:
I'm eager to force us into SSL-only all through the pipeline, and I'd strongly favor disabling HTTP completely. The main downside to that is our current deployment setup that uses HTTP-behind-HTTPS. I suspect that might require a bit more effort to resolve, but we should do that as a close followup if possible.

So, what do you think about programmatically setting the protocol config (similar to what we do for `isProd`/`isDev`) rather than allowing it to be set by an environment config file/variable? Something like

```js
const protocol = config.get('isProd') ?  'http': 'https';
config.set('asset_server.protocol', protocol);
```

Once we sort out the prod setup to be HTTPS end-to-end, we can get rid of any reference to HTTP/S throughout the codebase and assume it's all SSL, which would make things simpler long term.

mmcgahan:
> What is the value of SSL in dev?

It was mentioned in a few meetings, but HTTP2 and Service Workers require SSL, so it's a blocker for our PWA work. It also make dev more consistent with prod, which is a Good Thing.

> Is there value in using SSL in prod at the container level?

Yes, for HTTP2 - Hapi won't allow things like server push if you're running on a non-SSL connection.

> Does GAE even support that?

Yes - there's some general port/protocol docs here - https://cloud.google.com/appengine/docs/flexible/custom-runtimes/configuring-your-app-with-app-yaml#network_settings

> Would there be a performance hit?

Some, but unlikely to be significant. Lots of articles about it - https://www.google.co.nz/search?q=ssl+performance

> How does this fit into the dev work flow? Will it be used by default or for troubleshooting an https specific problem?

We should aim to be SSL-only. 'Advanced' web features don't work the same (or at all) over non-SSL, and I can imagine the inconsistencies creating a lot of avoidable debug issues in dev.

mmcgahan:
@mattkime we can probably set up Hapi to auto-redirect to HTTPS in dev to address the 'lazy url' concern

willh-meetup:
@mmcgahan we would need to start the http server on another port, which if we aren't using 80 and 443, will no redir automatically.

src/server.js
willh-meetup: 👍 Was thinking about adding a custom validator that makes sure the file exists.

willh-meetup:
@mmcgahan 👍 to doing https more programmatically, generally speaking. 

One thing about that: If we do that, everyone will need to put the certs on their machine, otherwise everything will be broken. So, might want to do as a follow up once we've worked out any kinks.

src/server.js
willh-meetup: 👍

docs/Config.md
willh-meetup: Are we phasing out webpack dev server?

mattkime:
Spoke to mikem and had my questions answered, not that it had much bearing on this pr.

src/server.js
mmcgahan: sounds good. Unfortunately it looks like the custom validation callbacks defined by node-convict only take the current `value` as an argument, so I don't see a built-in way of validating interdependent values, but it's easy enough to do it outside of `config.validate()`. I'm considering whether to submit a PR to node-convict to add that functionality, though - seems like it would be pretty useful.

mmcgahan:
> If we do that, everyone will need to put the certs on their machine, otherwise everything will be broken

Yeah - probably necessary in the long run but definitely a sticking point in the short term. It would be pretty sweet to have a key/cert server available over VPN or something that could supply the data during a build step... maybe something to think about as a long-term solution that wouldn't require everyone keeping highly-sensitive files on their work (and sometimes personal) laptops

mattkime:
>wouldn't require everyone keeping highly-sensitive files on their work (and sometimes personal) laptops

Are they highly sensitive if they're just for `*.dev.meetup.com`?

mmcgahan:
> Are they highly sensitive if they're just for *.dev.meetup.com?

Ah, maybe not. In which case, we just need a simple distribution mechanism :)

src/util/config/index.js
```diff
@@ -203,5 +214,12 @@ config.set(
 config.set('isProd', config.get('env') === 'production');
 config.set('isDev', config.get('env') === 'development');
 config.validate();
+const appConf = config.get('app_server');
```
willh-meetup: @mmcgahan Let me know what you think of this.

src/util/config/build.js
```diff
@@ -61,6 +76,16 @@ if (asset_server) {
 
 config.set('isProd', config.get('env') === 'production');
 config.set('isDev', config.get('env') === 'development');
+
+const assetConf = config.get('asset_server');
```
willh-meetup: This is a little duplication, but I think I kind of prefer it to creating another thing to import.

mmcgahan: agreed since it's so quick-and-specific

src/util/config/index.js
mmcgahan: LGTM

docs/Config.md
mmcgahan: no - I was thinking more about the prod setup. If/when prod static files are served from the app server, we'll probably want to revisit using the WebpackDevServer middleware in dev instead of the standalone CLI-based tool just to keep the dev setup more consistent with prod, but yeah that would need to be sorted out before the idea of an `asset_server` could go away entirely.

PR #297, opened 2017-06-06
WP 334 promises from api requests

### LINKED ISSUE #271: Async API validation with redux-form
We recently did some validation via an API request for checking whether a particular pro network name and urlname are available ([PR](https://github.com/meetup/pro-web/pull/94)). This is fairly straight forwards with redux-form: return a Promise from a submit handler that resolves or rejects based on whether the form is valid or not. For a simple example, this is from their [documentation](http://redux-form.com/6.6.3/examples/submitValidation/):

```jsx
import { SubmissionError } from 'redux-form'

const sleep = ms => new Promise(resolve => setTimeout(resolve, ms))

function submit(values) {
  return sleep(1000) // simulate server latency
    .then(() => {
      if (![ 'john', 'paul', 'george', 'ringo' ].includes(values.username)) {
        throw new SubmissionError({ username: 'User does not exist', _error: 'Login failed!' })
      } else if (values.password !== 'redux-form') {
        throw new SubmissionError({ password: 'Wrong password', _error: 'Login failed!' })
      } else {
        window.alert(`You submitted:\n\n${JSON.stringify(values, null, 2)}`)
      }
    })
}

```

However, the platform doesn't support `fetch` type api requests that return a promise, so this is roughly the workaround approach we had to take:

```jsx
export class MyReduxFormContainer extends React.Component {

	validating = false;

	componentWillReceiveProps(props) {
		// if the component should be validating and the api request is not in flight
		if (this.validating && !props.inFlight.includes(VALIDATE_FROM_API_REF)) {
			let validationErrors = {};

			// handle generic api failures
			if (props.validatedFromApi.error) {
				validationErrors = {
					_error: 'Sorry, something went wrong. Please try again.',
				};
			} else {
				// api request success

				// get the values we need from the response
				const {
					name_available,
					urlname_available,
				} = props.validatedFromApi.value;

				if (!name_available) {
					validationErrors = {
						...validationErrors,
						pro_account: 'This name is not available',
					};
				}

				// check that the urlname is valid
				if (!urlname_available) {
					validationErrors = {
						...validationErrors,
						urlname: 'This URL is not available',
					};
				}
			}

			if (Object.keys(validationErrors).length) {
				// if there are validation or api errors, reject the promise that was
				// returned from the submit handler
				this.rejectValidationPromise(new SubmissionError(validationErrors));
			} else {
				// if everything is valid, resolve the promise returned from the submit
				// handler
				this.resolveValidationPromise();
			}

			// set the validating flag back to false
			this.validating = false;
		}
	}

	handleSubmit = (formValues) => {
		// set a flag while we are supposed to be validating
		this.validating = true;

		// call an action that triggers the API_REQUEST
		this.props.actions.validateFromAPI(formValues);

		// redux-form expects a promise returned from a submit handler function for
		// async validation.
		return new Promise((resolve, reject) => {
			// expose the resove/reject functions to resolve the promise when the
			// api request completes or fails.
			this.resolveValidationPromise = resolve;
			this.rejectValidationPromise = reject;
		});
	};

	render() {
		return (
			<MyReduxForm onSubmit={this.handleSubmit} />
		);
	}
}

export function mapStateToProps(state) {
	return {
		validatedFromApi: state.api[VALIDATE_FROM_API_REF] || {},
		inFlight: state.api.inFlight || [],
	};
}

export function mapDispatchToProps(dispatch) {
	return {
		actions: bindActionCreators(actionCreators, dispatch),
	};
}

export default connect(mapStateToProps, mapDispatchToProps)(MyReduxFormContainer);
```

If there were something in the platform that supported making promised based api requests that resolved with the response then we could skip a lot of the workaround code (eg: exposing resolve/reject from the promise, setting a validating flag) and do something more like:

```jsx
export class MyReduxFormContainer extends React.Component {
	handleSubmit(formValues) {
		// maybe some sort of platform defined `fetch` function that acts like fetch,
		// but takes a query and goes through the platforms api layer and may even
		// update the redux store if needed. The important thing here though (no matter the 
		// details of the implementation) is that the `thing` doing the api request
		// returns a promise that resolves with the response.
		return this.props.actions.fetch(validateFromAPIQuery(formValues))
			.then((response) => {
				let validationErrors = {};

				// handle generic api failures
				if (response.error) {
					validationErrors = {
						_error: 'Sorry, something went wrong. Please try again.',
					};
				} else {
					// api request success

					// get the values we need from the response
					const {
						name_available,
						urlname_available,
					} = response.value;

					if (!name_available) {
						validationErrors = {
							...validationErrors,
							pro_account: 'This name is not available',
						};
					}

					// check that the urlname is valid
					if (!urlname_available) {
						validationErrors = {
							...validationErrors,
							urlname: 'This URL is not available',
						};
					}
				}

				if (Object.keys(validationErrors).length) {
					throw new SubmissionError(validationErrors);
				}
			});
	};

	render() {
		return (
			<MyReduxForm onSubmit={this.handleSubmit} />
		);
	}
}

export function mapStateToProps(state) {
	return {};
}

export function mapDispatchToProps(dispatch) {
	return {
		actions: bindActionCreators(actionCreators, dispatch),
	};
}

export default connect(mapStateToProps, mapDispatchToProps)(MyReduxFormContainer);
```

To me that is much more cleaner. If redux-form is something that may be used more widely on platform based apps, I think we should talk about (or at least start thinking about) ways to simplify this type of async validation.

mmcgahan:
The code updates are actually pretty simple, but support a pretty sophisticated set of use cases. The one potential risk I can see in this system is that it could be used to side-step Redux-managed state, since the request responses can be gathered directly from the resolved Promise rather than pulled from the Redux `state` object. I'm not sure if that's a big practical step backward, but it does introduce some accidental complexity if some requests are managed in a "pure" Redux way and others are managed as Promises... I think Redux Form's Promise-based validation API is a poor architectural choice, but that is compensated by other features of the lib, so... what can ya do.

Closes #271


docs/Queries.md
```diff
@@ -193,6 +218,41 @@ const deleteAction = api.del(deleteQuery);  // note `api.del` not `api.delete` b
...
+import * as api from 'meetup-web-platform/lib/actions/apiActionCreators';
+
+function mapDispatchToProps(dispatch) {
+  return bindActionCreators({ get: api.get }, dispatch);
+}
+
+class Example extends React.Component {
+  onSubmit() {
+    const formQuery = { ... };
+    const apiRequest = api.post(formQuery); // this triggers the POST
```
mikespencer: Just trying to follow along and was wondering if this should be calling something like `this.props.get` instead of `api.post` in your example?

mmcgahan: ah yes, good catch. will fix

PR #298, opened 2017-06-06
do not allow undefined params in encoded queries

mmcgahan:
This will permanently fix a bug encountered here https://github.com/meetup/mup-web/pull/1122. When queries are url-encoded, we need to make sure that there aren't any keys with `undefined` values because our URL-encoding package (rison) will throw an error.

This PR allows `null` values to get through, but skips `undefined` values

### DIFF
--- a/src/components/BrowserApp.test.jsx
+++ b/src/components/BrowserApp.test.jsx
@@ -1,8 +1,8 @@
 import React from 'react';
-import TestUtils from 'react-dom/test-utils';
+import ShallowRenderer from 'react-test-renderer/shallow';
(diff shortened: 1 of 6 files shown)


src/components/BrowserApp.test.jsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
-import TestUtils from 'react-dom/test-utils';
+import ShallowRenderer from 'react-test-renderer/shallow';
```
mmcgahan: you can ignore this change - just fixing a deprecation warning in tests now that the shallow renderer has been moved to a different package.

PR #299, opened 2017-06-06
WP 271 tracking log referrer bug

mmcgahan:
2 related fixes here

1. Ensure that the first navigation action in the browser populates a valid `referrer` property to the `API_REQ`. In the current codebase, `referrer` only appears on the _second_ navigation action. Sadface for tracking.

2. Make sure that the tracking log is generated for the initial server-side request as well as subsequent AJAX requests. The current code only calls `trackApi` for AJAX requests. Sadface for tracking


mmcgahan:
Sorry for the late commit - I hadn't pushed the fix for issue number 1 in the description

PR #300, opened 2017-06-08
WP 356 301 redirect

mmcgahan:
Follow-on from #289, which provided basic HTTP 302 redirect support. This adds support for 301 (Permanent) redirects using a new `permanent` prop on the `<Redirect />` component

### DIFF
--- a/src/components/Redirect.jsx
+++ b/src/components/Redirect.jsx
@@ -50,6 +50,11 @@ type RedirectState = {
  * <Router to='http://example.com' />
  * // external URL, URL object
  * <Router to={new URL('http://example.com')} />
+
+ * // internal route, permanent redirect
+ * <Router to='http://example.com' permanent />
+ * // external URL, permanent redirect
+ * <Router to='http://example.com' permanent />
  * ```
  */
 class Redirect extends React.Component {
--- a/src/components/Redirect.jsx
+++ b/src/components/Redirect.jsx
@@ -52,7 +52,7 @@ type RedirectState = {
  * <Router to={new URL('http://example.com')} />
 
  * // internal route, permanent redirect
- * <Router to='http://example.com' permanent />
+ * <Router to='/foo/bar' permanent />
  * // external URL, permanent redirect
  * <Router to='http://example.com' permanent />
  * ```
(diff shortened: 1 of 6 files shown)


src/components/Redirect.jsx
```diff
@@ -43,30 +50,49 @@ type RedirectProps = {
  * <Router to='http://example.com' />
  * // external URL, URL object
  * <Router to={new URL('http://example.com')} />
+
+ * // internal route, permanent redirect
+ * <Router to='http://example.com' permanent />
```
eilinora: Was this supposed to be relative? Not seeing a diff with below

src/components/Redirect.jsx
```diff
@@ -43,30 +50,49 @@ type RedirectProps = {
...
 	render() {
-		const { to, push } = this.props;
+		const { to, push, permanent } = this.props;
 		if (to instanceof URL || testForExternal(to)) {
+			// this is an external URL, so let `withSideEffect` handlers do redirect
 			return null;
 		}
-		return <RouterRedirect to={to} push={push} />;
+		// internal route URL
+		return (
```
eilinora: Could you explain this a bit? Sorry. Trying to understand shy you need to add a route to do the redirect inside of it.

tests/integration/render.int.js
```diff
@@ -174,4 +174,46 @@ describe('Full dummy app render', () => {
...
+				.catch(err => {
+					server.stop();
+					throw err;
+				});
+		}));
+	it('redirects permanently (301) to external route for <Redirect to={URL object} permanent />', () =>
+		start(getMockRenderRequestMap(), {}).then(server => {
+			const request = {
+				method: 'get',
+				url: '/redirect/external/permanent',
```
eilinora: If this is an external redirect test shouldn't it start with http? Not sure why this is passing :-/

mmcgahan: sorry, being a little 'clever' here in order to use the URLs to conditionally render different kinds of `<MockRedirect>` elements in `mockApp.jsx` - the last two section of the URL path are used to change the props passed to the `<Redirect>` element [here](https://github.com/meetup/meetup-web-platform/pull/300/files/a01145cb76ee165ca2b2f93cd04cc189ddf51347#diff-c22e2430a4ffb0ef9932d7e94b5330f1R20)

src/components/Redirect.jsx
mmcgahan: added more code comments - hopefully it's more clear now. Essentially, the outer `<Route>` is providing a hook into the server-side rendering context in order to inject the `permanent` flag. The `<Redirect>` component doesn't have the same hook, but the `<Route>` will always render.

src/components/Redirect.jsx
mmcgahan: oops - copy-pasta fail. fixed now

tests/integration/render.int.js
eilinora: Ooohhh 👍 :)

PR #304, opened 2017-06-12
WP 354 tracking cleanup

mmcgahan:
The current code requires that a `response` exist for activity tracking, but all we really need is the `request`. The `response` is only used to set new tracking cookies, and that can be done separately in an `onPreResponse` handler.

This PR implements that refactor to make the core plugin simpler - about half the changes are related to refactoring tests to use the new, simpler `request` arguments.

### DIFF
--- a/src/plugins/tracking/_activityTrackers.js
+++ b/src/plugins/tracking/_activityTrackers.js
@@ -10,10 +10,6 @@ const parseUrl = url.parse;
  * This module exports specific tracking functions that consume the `request`
  * object and any additional arguments that are relevant to generating the
  * tracking record.
- *
- * Note that some trackers use cookies to maintain state across sessions - these
- * functions are not guarantees to be pure because they may use the
- * `request.state` interface to set long-lived cookies
  */
 
 /*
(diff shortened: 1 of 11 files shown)


src/plugins/tracking/_activityTrackers.js
```diff
@@ -7,13 +7,9 @@ import { updateTrackId, newSessionId } from './util/idUtils';
...
 
 /*
- * This module exports specific tracking functions that consume the `response`
+ * This module exports specific tracking functions that consume the `request`
  * object and any additional arguments that are relevant to generating the
  * tracking record.
- *
- * Note that some trackers use cookies to maintain state across sessions - these
- * functions are not guarantees to be pure because they may use the
- * `response.state` interface to set long-lived cookies
```
willh-meetup: I know this is based off of `request.response.state` now, but should we still mention here that we are still setting those cookies?

mmcgahan: not in this module - since the functions are now pure and only add some metadata to `request.plugins.tracking`, the actual cookie setting side effect is isolated to the `onPreResponse` handler registered in `activity.js`, which is the main entrypoint to the plugin and documents the cookie behavior

mmcgahan:
In setting up this PR, I again ran into some refactoring requirements of the current login/logout support, which was a little annoying since the whole login/logout flow isn't really mature, so I've filed a Jira ticket to roll back all direct login/logout support - I think it would be best to leave that for a dedicated product-team project at some point, and plan on making a solid 'redirect to meetup classic login' in the interim, which is what the Pro team is doing currently. It will make some of our module refactoring significantly simpler

https://meetup.atlassian.net/browse/WP-362

willh-meetup:
👍 Yeah, that sounds good for whenever we can dedicate time to replatforming login.

PR #306, opened 2017-06-13
WP-352 supply match object to query functions

mmcgahan:
The `match` object from React Router contains some useful info beyond the `location` and `params` that we've been supplying until now - we already have it available in the router, so might as well pass it along to query functions.

@eilinora had simulated this behaviorin one of her mup-web PRs, which was the motivation for the update - we shouldn't need to re-calculate the `match` inside a query function any more.

### DIFF
--- a/src/util/routeUtils.js
+++ b/src/util/routeUtils.js
@@ -106,7 +106,7 @@ const _matchedRouteQueriesReducer = (location: URL) => (
 	// call the query functions with non-url-encoded params
 	const params = decodeParams(match.params);
 	const routeQueries = routeQueryFns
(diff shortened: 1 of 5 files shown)


src/util/routeUtils.js
```diff
@@ -106,7 +106,7 @@ const _matchedRouteQueriesReducer = (location: URL) => (
 	// call the query functions with non-url-encoded params
 	const params = decodeParams(match.params);
 	const routeQueries = routeQueryFns
-		.map(queryFn => queryFn({ location, params }))
+		.map(queryFn => queryFn({ ...match, location, params }))
```
sadafie: 👍

PR #308, opened 2017-06-21
Update docs and config management to default HTTPS

mmcgahan:
A few updates, with docs.

I wanted to make it 'easy' to configure the platform to use the SSL certs, so I've updated the config files to

1. Default to HTTPS everywhere
2. assume that the cert files are located at `~/.certs/star.dev.meetup.com.{crt|key}`

The `scp` command for getting those files onto your dev machine is in the docs, so hopefully it's easy for people to do the Right Thing. It's also possible to put them somewhere custom and supply a path the certs in a config file, but hopefully most people won't need to do that.

In development, the config module will also automatically fall back to HTTP if things are not set up correctly, but it will print a warning. In prod, an invalid SSL config will throw an Error on startup.

I think this implementation will make it quick and easy for people to convert to SSL in dev.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -75,8 +75,9 @@ The secrets in this example are dummy values and won't work in an actual app.
 ### Running over HTTPS
 
 By default, the app server and asset server will run over HTTPS, but you must
-supply the cert files (`.key` and `.crt`). By default, the platform will look
-for these files in `~/.certs/` and warn you if they are not found.
+supply the cert files (`.key` and `.crt`). The platform will look for these
+files in `~/.certs/` and warn you if they are not found, although you can supply
+a custom file path if desired.
 
 To get the cert files, you can copy them from the chapstick repo at
 `/util/conf/localdev/ssl/`. You should copy the
@@ -87,6 +88,7 @@ your dev machine.
 checked into version control
 
 ```
+$ mkdir ~/.certs # if it doesn't exist
 $ scp www.dev.meetup.com:/usr/local/meetup/util/conf/localdev/ssl/star.dev.meetup.com.* ~/.certs/
 ```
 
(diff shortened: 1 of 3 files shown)


src/util/config/index.js
```diff
@@ -213,13 +216,24 @@ config.set(
...
 config.set('isDev', config.get('env') === 'development');
-config.validate();
 const appConf = config.get('app_server');
 if (
 	appConf.protocol === 'https' &&
 	(!fs.existsSync(appConf.key_file) || !fs.existsSync(appConf.crt_file))
 ) {
-	throw new Error('Missing HTTPS cert or key!');
+	const message = 'Missing HTTPS cert or key for application server!';
+	if (config.isProd) {
```
willh-meetup: I think this could break prod.

RIght now we're not including certs in the app server, only in the configs for the nginx server. This was OK before, because `http` was the default for production. If we're definitely overwriting this value in prod, it's OK, but I'm not sure if that's the case.

docs/Config.md
```diff
@@ -78,42 +74,40 @@ The secrets in this example are dummy values and won't work in an actual app.
 
 ### Running over HTTPS
 
-You need to change change the `protocol` to `https`, and tell the server where
-to find the certificate and the private key (`crt_file` and `key_file`). You'll
-need to add these values to both the `asset_server` and the `app_server`.
+By default, the app server and asset server will run over HTTPS, but you must
+supply the cert files (`.key` and `.crt`). The platform will look for these
+files in `~/.certs/` and warn you if they are not found, although you can supply
```
willh-meetup: I like this.

src/util/config/index.js
mmcgahan: ah, yes - was just fixing that up. I'm going to keep the 'http in prod' behavior for now and wait for a separate bug to enable https behind the load balancer since it's not strictly necessary right now and I don't want to make this more complicated than necessary

willh-meetup: 👍

PR #309, opened 2017-06-22
add app-server host config

sadafie:
Not exactly sure what the reasoning was with not having a `app_server.host` config, but this just adds it back in.

The start info script in `mup-web` expects a value. So we should either add is back in, or edit mup-web to remove or hardcode that line in the script.

### DIFF
--- a/src/util/config/index.js
+++ b/src/util/config/index.js
@@ -107,7 +107,11 @@ export const config = convict({
 				: 'https',
 			env: 'DEV_SERVER_PROTOCOL', // legacy naming
 		},
-		// host: '0.0.0.0', ALWAYS 0.0.0.0
+		host: {
+			format: validateServerHost,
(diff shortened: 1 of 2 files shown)


mmcgahan:
`start-info` is going away with MWP-CLI, so the buggy-looking `http://undefined...` output will go away then.

`app_server.host` isn't used functionally (it's always hard-coded to '0.0.0.0'), so it's good to keep our config values as lean as possible to avoid inconsistencies that might be confusing. The host value that we would want printed is something determined entirely by a `etc/hosts/` or similar external DNS definition - the platform can't really know what the desired `host` value is, so it probably shouldn't be part of the codebase

mmcgahan:
start-info is now dead in mup-web - I recommend closing this PR

mmcgahan:
Feel free to re-open if/when there's a need

PR #310, opened 2017-06-22
domain relative url for mu_api

mattkime:


mattkime:
Testing locally.

src/util/fetchUtils.js
```diff
@@ -87,8 +87,8 @@ export const getFetchArgs = (apiUrl, queries, meta) => {
 	const isFormData = queries[0].params instanceof FormData;
 	const isDelete = method === 'DELETE';
 
-	const fetchUrl = new URL(apiUrl);
-	fetchUrl.searchParams.append('queries', rison.encode_array(queries));
+	const searchParams = new URLSearchParams();
```
mmcgahan: `URL` is used because `URLSearchParams` doesn't have a good browser polyfill. Instead, I recommend doing something like

```js
const fetchUrl = new URL(window.location.origin);
fetchUrl.pathname = `apiUrl`;
```

and leaving the rest of the implementation intact

willh-meetup: Does this part only run in the browser, or does `window.location` get populated on the server?

Curious if it's the latter, because this might not work. We're routing traffic on GAE a little differently, which I think is what was causing `info.host` to be the wrong thing (we are setting a host header on requests to new platform requests on www.meetup.com to the appspot url).

mmcgahan: `fetchUtils` only runs in the browser - the app server exclusively uses the `request` lib for server-to-server requests, not `fetch` - `window` is safe to use here. Probably should move this comment up to the top of the module, but browser-only usage is declared [here](https://github.com/meetup/meetup-web-platform/blob/6fd9c7ba88dd3c1a2e7c85094f1413f7c47cebfa/src/util/fetchUtils.js#L154-L155)

mattkime: it looks like URLSearchParams gets polyfilled when the URL polyfill is supplied - https://github.com/inexorabletash/polyfill#url-api

mmcgahan: Ah okay cool.

willh-meetup: 👍 (RE only running in the browser)

src/util/fetchUtils.js
```diff
@@ -137,11 +137,10 @@ export const getFetchArgs = (apiUrl, queries, meta) => {
...
 	};
 	if (hasBody) {
-		config.body = isFormData
-			? queries[0].params
-			: fetchUrl.searchParams.toString();
+		config.body = isFormData ? queries[0].params : searchParams.toString();
 	}
-	const url = isFormData || !hasBody ? fetchUrl.toString() : apiUrl;
+	const url =
+		isFormData || apiUrl + (hasBody ? '' : `?${searchParams.toString()}`);
```
mmcgahan: watch out for shortcircuit here - `url` will be set to `true` when `isFormData` is true.

```js
const url = isFormData || !hasBody ? `${apiUrl}?${searchParams}` : apiUrl;
```

mattkime: I don't quite understand. Did I make a mistake? I'm not seeing it after rereading my code.

mmcgahan: yes - if `isFormData` is `true`, then `url` will be set to `true` instead of a URL string.

```js
const apiUrl = '/mu_api';
const hasBody = false;
const searchParams = new URLSearchParams();

// when `isFormData` is `false`, no problem:
let isFormData = false;
console.log(isFormData || apiUrl + (hasBody ? '' : `?${searchParams.toString()}`));
// > /mu_api?

// when `isFormData` is `true`, the assignment short-circuits and returns `true`:
isFormData = true;
console.log(isFormData || apiUrl + (hasBody ? '' : `?${searchParams.toString()}`));
// > true
```

mattkime: oh, it never occurred to me to read the code that way. I thought the `||` would prevent execution of any code that came after it if isFormData was true.

mmcgahan: it does stop execution, but it also 'returns' the first truthy value it finds.

mattkime: yes, i know this now but a couple of parens would have made it clear from first read.

mmcgahan: Prettier will actually strip out unnecessary parens like that - one option to make it more obvious is to set a new var like `const useQueryString = isFormData || !hasBody`

mattkime: I take strong disagreement with prettier

mmcgahan: looks like there's a lot of discussion about it:

https://github.com/prettier/prettier/issues/1304
https://github.com/prettier/prettier/issues/187

PR #311, opened 2017-06-23
ServiceWorker plugin

mmcgahan:
Ignore the `flow-typed/` updates - just adding flow support for mwp-cli and updating the libdefs for other packages

2 updates in this PR:

1. a ServiceWorker plugin that registers the `asset-service-worker.{localeCode}.js` route on the server
2. a `<ServiceWorker>` wrapping component to register the service worker in `componentDidMount`.


src/components/PlatformApp.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Provider } from 'react-redux';
 import SyncContainer from './SyncContainer';
+import ServiceWorker from './ServiceWorker';
```
mmcgahan: start review here

src/plugins/service-worker/index.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+const buildPaths = require('mwp-cli/src/commands/buildCommands/config/paths');
```
mattkime: is there a way we can avoid interdependencies between modules?

src/plugins/service-worker/index.js
```diff
@@ -0,0 +1,45 @@
...
+	next: () => void
+) {
+	/*
+	 * Route for service worker script at top-level path. Depends on `Inert`
+	 * `path` must match client `serviceWorker.register` call - MWP provides this
+	 * in the `<ServiceWorker>` component
+	 */
+	server.route({
+		method: 'GET',
+		path: '/asset-service-worker.{localeCode}.js',
```
mattkime: i'm curious why we need a service worker js for each locale. it appears to be part of how the webpack plugin works but still wonder about the underlying need.

src/plugins/service-worker/index.js
mmcgahan: Not entirely. The main challenge here is that the service worker is a result of a build plugin - it can't be separated from the build config - so we need to depend on the builder for info about where the file ends up. There might be some options for making the dependency structures more clear.

src/plugins/service-worker/index.js
mmcgahan: the service worker is generated by a plugin that is defined once-per-build, which ends up meaning once-per-language. We actually want to keep that separation, because the way the plugin works is to check the filenames of all chunks emitted by the build, and define them all as 'precache' resources - we wouldn't want a single service worker to try to precache all versions of the app in all languages, which would be many megabytes of assets.

src/plugins/service-worker/index.js
mattkime: I do have a recommendation and I should make a ticket as a follow up -

we need a method of providing project level config. babel and jest have ways of doing it (package.json content or separate config files) so it should be possible with our own modules.

src/plugins/service-worker/index.js
mattkime: I'm realizing that I don't understand the caching behavior that this implements. Before this is used in production we need all devs to be aware of how this will affect caching in production. 

I did do a quick google search and its my impression that caching behavior is implementation specific.

mmcgahan: > caching behavior is implementation specific.

I'm not sure what this means. The precache behavior is described here https://github.com/GoogleChrome/sw-precache, but essentially the service worker will download _all_ the files generated by the webpack build in the background when it first installs. This means that async chunks and some of the images that are generated by the bundle will load instantly when they're needed on subsequent navigation.

One of my tickets for this sprint is to document current caching behavior in the platform - I'll make sure to include a section about this service worker.

src/plugins/service-worker/index.js
mmcgahan: yeah, we thought a lot about this when setting up the node-convict utility, but there is probably more that can be done with that - we use `config.{env}.json` files to define project-level config. As @sadafie can attest, it gets complicated very quickly.

src/plugins/service-worker/index.js
mattkime: The goal of the ticket seems to have migrated a bit (which is fine) and its taking me a moment to catch on. 

My initial expectation of service worker implementation was the basic caching the browser already does, but functional without a network connection. I thiiink thats what this does, in part - https://github.com/GoogleChrome/samples/blob/gh-pages/service-worker/selective-caching/service-worker.js

Anyway, thats what I mean by 'caching behavior is implementation specific' - its not a service worker thing, but how you implement the service worker.

Precaching - a term I was misunderstanding - means front loading assets. I suspect we don't want to front load all assets for our site, at least not in the long term. Changing asset loading like this is worthy of group discussion.

Finally, I'm confused about how the service worker cache is cleared.

---

While there are items to discuss, my primary concern is "will this confuse developers?" if the answer is no, then carry on and we can decide later whether or not to change course.

mmcgahan: The original ticket was implementing a webpack offline plugin, so the implementation is still well-aligned with that that goal - I've included comments to update what I learned when investigating options https://meetup.atlassian.net/browse/WP-321

> I suspect we don't want to front load all assets for our site, at least not in the long term

With our current assets, it's not an issue because it's only a few hundred kB of assets (much less than the image payload for many of our pages), but we can definitely tweak [the  plugin config](https://github.com/meetup/mwp-cli/blob/270cb2bbd84d8455c695e1583c1798f1ac73e0e8/src/commands/buildCommands/config/webpack/browserAppConfig.js#L143-L152) as desired over time to tune performance.

It might be worth looking up, but I'm pretty sure the service worker cache is effectively the same as the browser cache, so we don't need to manually manage purging it ourselves.

Since the service worker will only run in prod and there is no consumer-app code needed to make it work, I think we don't need to worry too much about confusion for developers. However, I will document some of this behavior in the caching docs that I put together this sprint.

src/plugins/service-worker/index.js
mattkime: I'd like to take a look at what can be done here, even if its just showing me all the hazards of going down this path. 

Interdependencies can get complicated very quickly as well.

I'll create a follow up ticket.

mmcgahan: what do you mean by 'interdependencies'? Currently there is a one-way dependency between mwp and mwp-cli, which seems reasonable to me - the platform depends on the platform builder for build config info.

PR #312, opened 2017-06-26
call trackSession with request, not response

mmcgahan:
`request.trackSession` was being called with the wrong argument. The simple fix would be to just change the argument, but Hapi has a better option for injecting the `request` object into methods that have been 'decorated' on the request, so this PR is a light refactor to change the function signature of the request `track___` methods - the `request` object is now injected as a curried argument. The overall structure change makes it easier to work with 'child tracking' methods like `trackLogin` and `trackApi` that are delegated to by `trackApi`, as well as making it much more difficult to accidentally refactor the trackers in such a way that the `request` object becomes inaccessible, which was the problem in the first place.

### DIFF
--- a/src/plugins/tracking/_activityTrackers.js
+++ b/src/plugins/tracking/_activityTrackers.js
@@ -17,9 +17,7 @@ const parseUrl = url.parse;
  * link (JOIN) between track ids pre/post login/logout, but does not carry over
  * to a new browser session
  */
-export const getTrackSession: TrackGetter = (trackOpts: TrackOpts) => (
-	request: Object
-) => {
+export const getTrackSession: TrackGetter = trackOpts => request => () => {
 	const { log, sessionIdCookieName } = trackOpts;
 	if (request.state[sessionIdCookieName]) {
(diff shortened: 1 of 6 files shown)


src/plugins/tracking/_activityTrackers.js
```diff
@@ -95,36 +89,31 @@ export const getTrackApiRequest: TrackGetter = (trackOpts: TrackOpts) => (
...
-	const trackLogout = getTrackLogout(trackOpts);
-
-	return (request: Object, queryResponses: Array<Object>) => {
-		const payload = request.method === 'post' ? request.payload : request.query;
+export const getTrackApi: TrackGetter = trackOpts => request => (
+	queryResponses: Array<Object>
+) => {
+	const payload = request.method === 'post' ? request.payload : request.query;
 
-		const metadataRison = payload.metadata || rison.encode_object({});
```
mmcgahan: this is all an indentation change because we no longer need to create local `track{ApiRequest|Login|Logout}` functions - they are now `request` methods

PR #318, opened 2017-07-06
Dedicated caching docs

mmcgahan:
Probably best to view [the formatted doc](https://github.com/meetup/meetup-web-platform/blob/4765e4c9a60b5705f698151f9c64c7e5b28a135f/docs/Caching.md) rather than the raw markup. This document provides an overview of current platform caching tools and discusses future possibilities once we have `Cache-Control` headers available from REST API responses. We can start to get very sophisticated with our caching once that happens, but for now we're in a world full of HTTP requests.

### DIFF
--- a/docs/Caching.md
+++ b/docs/Caching.md
@@ -8,7 +8,7 @@ responsiveness. There are two primary types of content that are cached:
 
 ## Overview
 
-![Cache diagram](https://user-images.githubusercontent.com/1885153/27892985-d75546fe-6256-11e7-8272-66252caa5c85.png)
+![Cache diagram](https://user-images.githubusercontent.com/1885153/28044152-c9f076ce-6629-11e7-9c6b-1bdf31fb6b50.png)
 
 ([from web platform diagrams](https://docs.google.com/presentation/d/1c8jf8UtGa81Ay4oqlbHfoP4Ile2VkiRVrCMVag6cRBQ/#slide=id.g1f921bb8da_0_0))
 
--- a/docs/Caching.md
+++ b/docs/Caching.md
@@ -1,6 +1,6 @@
 # Caching
 
-Web platform apps cache data at a few different levels in order to improve UI
+Network data is cached at a few different levels in order to improve UI
 responsiveness. There are two primary types of content that are cached:
 
 1. REST API responses
@@ -32,12 +32,6 @@ discussion](#future) for potential changes to the behavior.
 
 ### Browser
 
-The browser uses
-[IndexedDB](https://developer.mozilla.org/en/docs/Web/API/IndexedDB_API) for
-local storage of API response data - it caches the JSON response rather than the
-full HTTP response in order to avoid complexity arising from the batched API
-request/response routine.
-
 #### Service Worker
 
 The service worker does not cache API requests/responses (and probably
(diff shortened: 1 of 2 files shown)


docs/Caching.md
```diff
@@ -0,0 +1,158 @@
+# Caching
+
+Web platform apps cache data at a few different levels in order to improve UI
```
mattkime: I'd go with `Data is cached at a few different levels…` since the platform isn't always responsible for caching

docs/Caching.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Overview
+
+![Cache diagram](https://user-images.githubusercontent.com/1885153/27892985-d75546fe-6256-11e7-8272-66252caa5c85.png)
+
+([from web platform diagrams](https://docs.google.com/presentation/d/1c8jf8UtGa81Ay4oqlbHfoP4Ile2VkiRVrCMVag6cRBQ/#slide=id.g1f921bb8da_0_0))
+
+## 1. API responses
+
+Currently the Meetup REST API does not provide a
```
mattkime: There's no particular reason why we're dependent upon caching headers to determine how mup-web should cache api responses, although we may decide it would be desirable to act upon them.

docs/Caching.md
```diff
@@ -0,0 +1,158 @@
...
+discussion](#future) for potential changes to the behavior.
+
+### App server
+
+The app server does not currently cache any API responses. [See the 'Future' 
+discussion](#future) for potential changes to the behavior.
+
+### Browser
+
+The browser uses
```
mattkime: How long is it cached for? When is the cache purged? May it be purged manually?

docs/Caching.md
```diff
@@ -0,0 +1,158 @@
...
+
+## 2. Static assets
+
+Static assets generated by the build are cached in our CDN and in the browser.
+Because each filename includes a hash of the file contents, the file can have an
+infinite cache lifetime.
+
+### CDN
+
+(CDN cache invalidation policy?)
```
mattkime: we might ask @alessandrobologna about the fastly cdn cache invalidation process

docs/Caching.md
```diff
@@ -0,0 +1,158 @@
...
+All static assets are delivered to the browser with a `Cache-Control` header
+with a a 10 year expiration. For
+[browsers that do not support Service Workers](https://jakearchibald.github.io/isserviceworkerready/),
+normal browser cache rules apply - static assets will be cached as they are
+requested by the app, but not before. For browsers that _do_ support service
+workers, the platform service worker will 'pre-cache' all bundled assets
+asynchronously after the in-page assets load, and serve those assets from the
+browser cache for all future requests.
+
+When new assets are generated, a new service worker is generated. When a client
```
mattkime: how does the client know to download the new service worker? does it have a hashed filename like other built assets?

docs/Caching.md
```diff
@@ -0,0 +1,158 @@
...
+
+Web platform apps cache data at a few different levels in order to improve UI
+responsiveness. There are two primary types of content that are cached:
+
+1. REST API responses
+2. Static assets (files)
+
+## Overview
+
+![Cache diagram](https://user-images.githubusercontent.com/1885153/27892985-d75546fe-6256-11e7-8272-66252caa5c85.png)
```
mattkime: It looks like all data goes through the service worker in the browser. I think this is a bit misleading because

a) not all browsers have a service worker
b) the service worker is only used to provide cached static assets
c) the browser uses IndexedDB to cache API responses

docs/Caching.md
```diff
@@ -0,0 +1,158 @@
...
+Because each filename includes a hash of the file contents, the file can have an
+infinite cache lifetime.
+
+### CDN
+
+(CDN cache invalidation policy?)
+
+### Browser
+
+All static assets are delivered to the browser with a `Cache-Control` header
```
mattkime: might be worth having headers for 'Static Asset Cache' and 'Service worker cache' to emphasize the different tech at play. Wouldn't need any more explanation.

docs/Caching.md
```diff
@@ -0,0 +1,158 @@
...
+
+The cache is stored locally in the user's browser using IndexedDB so that it
+will survive multiple sessions. All API queries are stored, and the data is
+refreshed each time the query is re-requested.
+
+In dev, be aware that the cache may be masking failed API requests - keep your
+network dev tools open and watch the server logs to make sure you haven't broken
+anything.
+
+### Future
```
mattkime: I'm a little confused about including the Future section here, checked into the repo. Seems like this is more 'Issues' territory.

I'd like to have some conversation about which options we may consider in the future but would like to find another venue to discuss it.

docs/Caching.md
mmcgahan: I'll update/clarify. Technically, all data _does_ go through the service worker, but the service worker passes through un-cached data transparently. The IndexedDB cache appears at the Redux middleware layer, not the 'browser' as a whole, so I'll make that more explicit

docs/Caching.md
mmcgahan: We definitely should rely on the data source to tell the platform how long to cache the data - i.e. the platform shouldn't be creating its own cache lifetime policy independent of the API, but maybe you're thinking of something else?

docs/Caching.md
mmcgahan: I'll add more info about how the cache is purged - it basically lasts as long as the logged-in session lasts. It can be manually purged in a similar way to clearing cookies, so I'll add that info

docs/Caching.md
mmcgahan: my understanding is that essentially service worker scripts are _never_ cached - they are always re-downloaded and the running service worker instance is 'upgraded' if the new script is different than the old script.

docs/Caching.md
mattkime: I think there's a lot to be worked out here and I suspect there will be some back and forth. The UI dictates data freshness needs - is it the home page were we need a simple list of groups or is an admin editing group content? In the first example we could use rather stale data but in the second one we'd want whatever is up to date. This is why I pause at the idea that the back end would drive data freshness concerns.

docs/Caching.md
mmcgahan: Moved to https://github.com/meetup/meetup-web-platform/issues/322

docs/Caching.md
mmcgahan: Yeah, definitely lots of things to consider as we flesh out caching behavior. I've adjusted this wording to avoid saying the data 'cannot' be cached, and instead just state that it is not currently cached.

docs/Caching.md
mmcgahan: Thinking about this a bit more, I think we actually don't have to worry about invalidating the fastly cache since it only uses hashed filenames - the cache doesn't need to expire, so I've added that statement

mmcgahan:
new link to formatted doc: https://github.com/meetup/meetup-web-platform/blob/4765e4c9a60b5705f698151f9c64c7e5b28a135f/docs/Caching.md

docs/Caching.md
mattkime: I agree with you, but i'm still curious in case things go off the rails. but this could be addressed another time.

PR #319, opened 2017-07-07
WP 379 default SSL port 443 in dev

mmcgahan:
This is sort of a weird solution because of some OS-level limits on binding to ports < 1024.

In order to use port 443 and thereby be able to run the dev server at https://beta2.dev.meetup.com rather than https://beta2.dev.meetup.com:8000, the `node` command needs to be run with `sudo`, so I've made it an opt-in port assignment that will only be applied when the server is run with `sudo` - in consumer apps, you'll have to run something like

```sh
$ sudo yarn start
# or assign the current env vars to the `sudo` context with `-E`
$ sudo -E yarn start
```

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -41,7 +41,7 @@ export default function start(
 
 	const connection = {
 		host: '0.0.0.0',
-		port: appConfig.app_server.port, // needed for custom port behind SSL proxy in prod
+		port: appConfig.app_server.port, // needed for custom port in prod
 		routes: {
 			plugins: {
 				'electrode-csrf-jwt': {
--- a/src/server.js
+++ b/src/server.js
@@ -41,7 +41,7 @@ export default function start(
 
 	const connection = {
 		host: '0.0.0.0',
-		port: appConfig.app_server.port,
+		port: appConfig.app_server.port, // needed for custom port behind SSL proxy in prod
 		routes: {
 			plugins: {
 				'electrode-csrf-jwt': {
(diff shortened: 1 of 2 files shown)


mattkime:
How would this functionality be exposed to developer? Would you need to type your password every time you start the server? seems like that would be a deal breaker.

mmcgahan:
@mattkime normal `sudo` rules would apply - I think there's a 15 minute timeout or something for re-entering your password on mac? I also found that the files that are created in the `build/` directory can only be deleted with `sudo`, which is pretty annoying.

I'm not convinced this change is worth it, but it's also a pretty trivial code change, and doesn't require maintenance. @eilinora I know you were initially interested in this change, but do you think it's worth the annoyance of having to run with `sudo`? The 'invisible' solution would require running a proxy server in the background that has root privileges, which adds a whole bunch of complexity/maintenance that I'm not interested in officially supporting

eilinora:
awww man! yeah, it would be nice to have this work, but not if it means `sudo`. I think the password entering would be far more annoying over time then entering `:8000` when needed.

mattkime:
i'm ambivalent about this. it doesn't sound like it will be used but its not complicated. the reason not to add it is because it might confuse someone down the road

mmcgahan:
@mattkime agreed. I think I'll close without merging for now - it's simple to re-implement it in the future if there really is a growing desire for it, but it's more of a distraction for now.

PR #321, opened 2017-07-10
yarn upgrade july 2017

mattkime:


package.json
```diff
@@ -72,29 +72,29 @@
...
-    "mwp-cli": "1.0.71",
-    "prettier": "1.4.4",
-    "react": "15.5.4",
-    "react-addons-test-utils": "15.5.1",
+    "lint-staged": "4.0.1",
+    "meetup-web-mocks": "1.0.194",
+    "mwp-cli": "1.0.74",
+    "prettier": "1.5.2",
+    "react": "15.6.1",
+    "react-addons-test-utils": "15.6.0",
```
mmcgahan: this is an orphaned dependency and can be removed (it was moved into `react-dom` a month or so ago and import paths were updated

PR #324, opened 2017-07-13
Support api meta request headers

akdetrick:
Adds support for api custom meta request headers. See also: https://www.meetup.com/meetup_api/#meta-headers

### DIFF
--- a/src/util/apiUtils.js
+++ b/src/util/apiUtils.js
@@ -247,6 +247,10 @@ export const buildRequestArgs = externalRequestOpts => ({
 		headers['X-Meetup-Request-Flags'] = (flags || meta.flags).join(',');
 	}
 
+	if (meta.apiMetaHeaders) {
+		headers['X-Meta-Request-Headers'] = meta.apiMetaHeaders.join(',');
+	}
+
 	if (meta.variants) {
 		headers['X-Meetup-Variants'] = Object.keys(
 			meta.variants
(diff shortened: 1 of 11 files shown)


src/util/apiUtils.test.js
```diff
@@ -37,6 +37,8 @@ import {
 	groupDuotoneSetter,
 } from './apiUtils';
 
+const API_META_HEADER = 'X-Meta-Request-Headers';
```
mmcgahan: might be better to define this constant in `apiUtils.js` and then import it here

src/util/apiUtils.js
```diff
@@ -247,6 +247,10 @@ export const buildRequestArgs = externalRequestOpts => ({
 		headers['X-Meetup-Request-Flags'] = (flags || meta.flags).join(',');
 	}
 
+	if (meta.apiMetaHeaders) {
```
mmcgahan: what do you think about calling this `metaRequestHeaders`? Just checking [the relevant Meetup API docs](https://www.meetup.com/en-AU/meetup_api/#meta-headers) and it looks like there are a couple other headers that are also considered 'meta headers', so I think this one should be named something more specific.

mmcgahan:
This can merge once you add a heading/description for

##### `metaRequestHeaders`

in `Query.md`

akdetrick:
Because I already added docs in `master`, I added the heading in the Queries.md docs in `master` as well.

PR #326, opened 2017-07-18
Wp 408 api proxy module

mmcgahan:
This ticket packages all the api-proxy code into a well-contained module. **There is no new functionality here**, just file and method organization. The plugin is not currently imported directly by consumer applications - the calls are all internal to the platform. The 'public interface' is _just_ the plugin `register` function.

Highlights:

- the `mu_api/` route is registered as part of the plugin rather than being defined separately in the platform `routes.js` file. The route's `path` is configurable as a plugin option.
- `request.proxyApi$(queries)` will make API requests based on the queries passed in as an array argument
- duotone utilities are consolidated into the plugin - they shouldn't be needed anywhere else
- 95% unit test coverage

---

**Info for reviewers**

The plugin code is all inside `src/plugins/api-proxy/`, and consists of

1. `index.js` - the plugin 'register' function that applies the features to the Hapi `server` object
2. `routes.js` - supplies the GET/POST routes for `/mu_api` (minimal change - this file was moved)
3. `handler.js` - supplies the API proxy route handler. This file was moved - very little modification. It is separate from `routes.js` so that it is easy to mock in integration tests
4. `proxy.js` - this is the `request.proxyApi$()` - mostly just a file move, with light modification
5. `util/*.js` - the helpers for making the API requests based on the `queries` provided to `proxyApi$`

Essentially, the biggest part of the cleanup/consolidation was changing the `proxyApiRequest$(request)` function into `request.proxyApi$()` - relatively small surface change, but much better internal organization.


src/plugins/api-proxy/handler.js
```diff
@@ -0,0 +1,35 @@
+import Joi from 'joi';
```
mmcgahan: This file was moved from src/apiProxy/apiProxyHandler.js, but `parseRequestQueries` was moved into it from `src/util/apiUtils.js` because we only need to parse the request queries in the api proxy endpoint (handled here), not anywhere else

src/plugins/api-proxy/proxy.js
```diff
@@ -1,53 +1,34 @@
...
 // Implicit dependency: tracking plugin providing request.trackApi method
 import { Observable } from 'rxjs/Observable';
 import 'rxjs/add/observable/zip';
 
 import {
 	buildRequestArgs,
 	makeApiRequest$,
-	parseRequest,
-} from '../util/apiUtils';
+	getExternalRequestOpts,
```
mmcgahan: `parseRequest` was slightly refactored to _only_ generate `externalRequestOpts`, which are the 'options' supplied to the `request` call to the REST API - essentially porting the incoming request attributes to the outgoing request.

mmcgahan:
I'm going to split up `api-proxy/util/index.js` a little further - there are some duotone methods in there that could be moved to `api-proxy/util/duotone.js`, and there might be some other logical grouping that could be split out as well

mmcgahan:
Actually, I'm going to leave as-is. The duotone cleanup is actually better done in a followup because it enables a much bigger and better simplification that is beyond the scope of this ticket.

mmcgahan:
https://meetup.atlassian.net/browse/WP-421

mmcgahan:
Need to add the file upload logic to this ticket, e.g. `serverUtils.js:onResponse`

mmcgahan:
@willh-meetup I know this PR is a pain to read, thanks so much for digging into it. I'll follow up

src/plugins/api-proxy/index.js
```diff
@@ -0,0 +1,74 @@
+// @flow
+import fs from 'fs';
+import util from 'util';
+import { duotones, getDuotoneUrls } from './util/duotone';
+import getApiProxyRoutes from './routes';
+import proxyApi$ from './proxy';
+
+const API_PROXY_PLUGIN_NAME = 'api-proxy';
+const API_ROUTE_PATH = '/mu_api';
```
mmcgahan: it definitely could be - I've basically set this as the plugin's default `path`, but it can be overridden by passing in `path` in the plugin options. I think I would lean toward keeping the default defined here, because it's unlikely to change and it feels simpler to avoid requiring an explicit setting to be passed down from the consumer. The downside is that it's a little hidden.

It would be pretty simple to add it to mwp-cli/src/config where we have most of our app config rules defined, but I almost feel like putting it there would make it easier to accidentally change and break stuff?

src/plugins/api-proxy/util/duotone.js
```diff
@@ -70,6 +70,16 @@ export const duotones = [
...
  */
 
+export const makeSign = (salt, ref) => rx => {
+	const spec = `event/${rx}/${ref}`;
+	const signature = crypto
+		.createHash('sha1')
+		.update(`${spec}${salt}`)
+		.digest('hex')
+		.substring(0, 10);
+	return `https://a248.e.akamai.net/secure.meetupstatic.com/photo_api/${spec}/sg${signature}`;
```
mmcgahan: 👍 Yeah Lori opened a ticket about that for me, so it's updated in `master` and I'll make sure the merges bring the update here

src/util/createStoreServer.js
```diff
@@ -2,7 +2,7 @@ import { applyMiddleware, createStore } from 'redux';
 import { parseQueryResponse } from './fetchUtils';
 import getEpicMiddleware from '../middleware/epic';
 import catchMiddleware from '../middleware/catch';
-import apiProxy$ from '../apiProxy/api-proxy';
+import 'rxjs/add/operator/map';
```
mmcgahan: it was always using the RxJS map, but we were missing the explicit operator import - it just happened to still work because a parent dependency must have also added the operator. This import just enforces the dependency properly.

willh-meetup: 👍

src/plugins/api-proxy/index.js
willh-meetup: It's also defined in `src/renderers/server-render.jsx`, so it ends up not really being entirely configurable, right? 

That was more the angle that I was coming from, that we're defining it in two places, but maybe only having it defined once is probably better, and some sort of config seemed like maybe an easy way to do that. 

I think you're right that making it easily configurable probably isn't the way to go, though, since we probably won't want to change it much.

mmcgahan: ah yes, you're right. I think it would make sense in the short term to kill the 'configurable' code, and just `import { API_ROUTE_PATH }` from the api-proxy module into `server-render.jsx` - that's a reasonable dependency requirement.

willh-meetup: 👍

src/plugins/api-proxy/util/index.test.js
```diff
@@ -561,71 +560,4 @@ describe('logApiResponse', () => {
...
-				oauth_token: 'foo',
-			},
-			server: getServer(),
-		};
-		expect(() =>
-			parseRequest(getRequest, 'http://dummy.api.meetup.com')
-		).toThrow();
-	});
-});
+describe('getExternalRequestOpts', () => {});
```
mmcgahan: yeah the commented out import and this stubbed out test were supposed to remind me to finish the test :). Since this function gets separated out into a standalone `send` module in #330, which I'm hoping to merge soon after this initial PR, I think I'll make the testing update there to avoid merge conflicts.

willh-meetup: 👍

src/plugins/api-proxy/util/index.test.js
```diff
@@ -28,14 +26,14 @@ import {
 	getLanguageHeader,
 	injectResponseCookies,
 	logApiResponse,
-	parseRequest,
+	// getExternalRequestOpts,
```
mmcgahan: removed

PR #327, opened 2017-07-19
WP 398 platform agent

mmcgahan:
The latest activity tracking schema allows us to specify `platform_agent: 'MUP_WEB' // or 'PRO_WEB'`, so I've updated the code to read from `config.agent` in `package.json`. It's a breaking change because each consumer app will need to specify `config.agent`, but that's pretty simple (already done in mup-web, and in PR in pro-web).

This PR also removes `config/build.js`, which is now supplied by `env` from `mwp-cli/src/config`


willh-meetup:
☝️ Looks like a unit test failure

src/plugins.js
```diff
@@ -1,8 +1,8 @@
 import Inert from 'inert';
 import HapiPino from 'hapi-pino';
 import CsrfPlugin from 'electrode-csrf-jwt';
+import config from 'mwp-cli/src/config';
```
mattkime: i think we have a weird dependency relationship between consumers, meetup-web-platform, and mwp-cli. It strikes me as an unwanted side effect that mwp-cli should need to be updated to supply config to meetup-web-platform, particularly since the value itself is coming from the consumer. 

Taking a look at meetup-web-platform, the other place where this occurs is providing the browser js output dir to the service worker code. This value could be provided by the consumer repo. Both values could be read directly from the consumer repos and the dependency between meetup-web-platform and mwp-cli could be removed.

src/util/config/index.js
```diff
@@ -5,9 +5,12 @@ import path from 'path';
 import chalk from 'chalk';
 import convict from 'convict';
 
-import buildConfig, { schema as buildSchema } from './build';
+import { env } from 'mwp-cli/src/config';
```
mattkime: my previous comment was based on looking at meetup-web-platform master.

I'm not 100%, but I suspect that similar techniques might help eliminate this dependency as well. I doubt asset_server.host, asset_server.port, asset_server.protocol are needed by meetup-web-platform, etc, i'm sorting through to figure out what is actually used.

src/plugins.js
```diff
@@ -104,12 +104,12 @@ function getServiceWorkerPlugin() {
 }
 
 export default function getPlugins() {
-	const { platform_agent, isProd } = config;
```
willh-meetup: This seems to be the culprit for the failed unit test.

We should either mock or add a config.agent to mwp's `package.json`

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## [5.0]
+
+### BREAKING CHANGE
```
willh-meetup: 👍

mmcgahan:
@mattkime re: config

Yeah, there are different approaches to simplifying config, but after a few false starts, I think the direction we're going now is probably the best compromise. The 2 ends of the spectrum are

1. Let each repo manage its own config
  - Pro: no dependency issues
  - Con: lots of redundancy
2. Consolidate all config into a single repo
  - Pro: no redundancy - 'single source of truth' for config info
  - Con: potentially complex dependencies

I've been moving us more toward option 2 because of the difficulty and awkwardness of keeping config in sync using option 1. I'm hoping/expecting that mwp-cli will remain at the top of the dependency chain for a long time, which means that it can be used as the single source of config truth across all repos without needing circular dependencies or redundant config modules, and making it easier to find where config values come from - each of those problems has appeared with the more 'localized config' approach that we started with.

It's actually sort of similar to the approach taken by Redux - one way data flow - where even though consumer applications get to define some of their config values (e.g. in `package.json`), those values are only _read_ from `mwp-cli/src/config` - data can be created downstream, but it is only consumed from upstream because upstream is responsible for coordinating the data from all sources

mmcgahan:
@willh-meetup ah, hadn't pushed that to mup-web yet. all set now, thanks.

PR #329, opened 2017-07-21
WP-424 app route plugin

mmcgahan:
Main changes involve consolidating the 'app route' into a single Hapi plugin

1. `src/routes/appRouteHandler.js` -> `src/plugins/app-route/handler.js`
2. `src/routes/appRoute.js` -> `src/plugins/app-route/route.js`
3. the server app rendering functions created in the server-app build process are now passed to the app route as a plugin option to the `getPlugins()` function. Previously, they were passed in to the `getRoutes()` function. Minor difference.
4. the entire plugin is covered by flowtype
5. `renderRequestMap` renamed to `languageRenderers`, which is a bit more clear about what is contained in the object
6. app route test moved from general routes module test to a plugin-route-specific test file.


tests/integration/server.int.js
```diff
@@ -1,6 +1,6 @@
 import { mockConfig } from '../mocks';
 import start from '../../src/server';
-import * as appRouteHandler from '../../src/routes/appRouteHandler';
+import * as appRouteHandler from '../../src/plugins/app-route/handler';
```
willh-meetup: Another thing probably outside the scope of this...

Might be nice to have an integration test for the language redirect behavior.

src/plugins/app-route/route.js
```diff
@@ -41,12 +37,8 @@ export const onPreResponse = {
  * - `state`: skip cookie validation because cookies can be set by other services
  *   on the same domain and may not conform to the validation rules that are
  *   used by Hapi
- *
- * @param {Object} renderRequestMap a map of 'localeCode' strings to request-
- *   rendering functions that return an HTML string
- * @return {Object} a Hapi route configuration object
  */
-const getApplicationRoute = renderRequestMap => ({
+export default (languageRenderers: { [string]: LanguageRenderer$ }) => ({
```
willh-meetup: 👍  `languageRenderers` is a better name than `renderRequestMap`

src/plugins/app-route/handler.js
```diff
@@ -0,0 +1,27 @@
+// @flow
```
willh-meetup: Might be worth adding a comment to this file explaining exactly what it's handling (language redirects and session tracking).

src/plugins/app-route/handler.js
```diff
@@ -0,0 +1,27 @@
...
+import url from 'url';
+
+export default (languageRenderers: { [string]: LanguageRenderer$ }) => (
+	request: HapiRequest,
+	reply: HapiReply
+) => {
+	const pathname = request.getPrefixedPath();
```
mmcgahan: Makes sense - I'll do that in a separate branch and merge straight to `next` before this PR merges

PR #330, opened 2017-07-21
WP 421 api proxy cleanup

mmcgahan:
Some further refinement to #326 - didn't want to expand that diff too much, but this PR provides better organization by separating out the 'Send' and 'Receive' logic of the API proxy routine - they are independent operations that should be organized into separate modules rather than being combined in a single `util/index` file.

The diff is large, but there are no functional changes - mainly just creating two separate files from one monolithic file, plus splitting out tests

`src/api-proxy/util/index.js` -> `src/api-proxy/util/send.js` + `src/api-proxy/util/receive.js`


src/plugins/api-proxy/util/send.js
```diff
@@ -0,0 +1,325 @@
...
+		agentOptions: {
+			rejectUnauthorized: baseUrl.indexOf('.dev') === -1,
+		},
+	};
+	if (request.mime === 'multipart/form-data') {
+		// the parsed payload includes string key-value pairs for regular inputs,
+		// and file descriptors for file upload inputs { filename, path, headers }
+		// The file descriptors can be converted to readable streams for the API
+		// request.
+		externalRequestOpts.formData = parseFormData(
```
mmcgahan: this is a change - breaking out the file-transforming logic into its own pure function

mmcgahan:
This is almost ready - I just need to add some unit tests for a few more send/receive methods that are new or didn't have coverage before.

src/plugins/api-proxy/util/send.js
```diff
@@ -0,0 +1,331 @@
...
+		'accept-language': getLanguageHeader(request),
+	};
+
+	delete externalRequestHeaders['host']; // let app server set 'host'
+	delete externalRequestHeaders['accept-encoding']; // let app server set 'accept'
+	delete externalRequestHeaders['content-length']; // original request content-length is irrelevant
+	delete externalRequestHeaders['content-type']; // the content type will be set in buildRequestArgs
+
+	// cloudflare headers we don't want to pass on
+	delete externalRequestHeaders['cf-ray'];
```
willh-meetup: We should be able to get rid of these now, since we aren't behind cloudflare anymore.

src/plugins/api-proxy/util/receive.test.js
```diff
@@ -0,0 +1,299 @@
...
+import { MOCK_GROUP } from 'meetup-web-mocks/lib/api';
+
+import { getServer, MOCK_LOGGER } from '../../../util/testUtils';
+
+import { API_PROXY_PLUGIN_NAME } from '../';
+import {
+	makeApiResponseToQueryResponse,
+	makeInjectResponseCookies,
+	makeLogResponse,
+	// getExternalRequestOpts,
```
willh-meetup: 🔪 ?

src/plugins/api-proxy/util/receive.js
```diff
@@ -0,0 +1,250 @@
...
+	};
+}
+
+/**
+ * @return {Object} { value, error? }
+ */
+export const parseApiValue = ([response, body]) => {
+	// treat non-success HTTP code as an error
+	if (response.statusCode < 200 || response.statusCode > 299) {
+		return formatApiError(new Error(response.statusMessage));
```
mmcgahan: happy to change, although I have no idea how the api proxy would handle a redirect response - make another request to the redirected endpoint? I don't think it would actually return a value, but I guess it's fine to let a 3xx return a null `value` for now instead of erroring. (I actually need to merge #336 in order to get some of this behavior smoothed out).

src/plugins/api-proxy/util/receive.js
```diff
@@ -0,0 +1,250 @@
...
+};
+
+export const makeReceive = request => {
+	const logResponse = makeLogResponse(request);
+	const injectResponseCookies = makeInjectResponseCookies(request);
+	return query => {
+		const parseApiResponse = makeParseApiResponse(query);
+		const apiResponseToQueryResponse = makeApiResponseToQueryResponse(query);
+		return response => {
+			logResponse(response); // this will leak private info in API response
```
mmcgahan: It's still pretty useful because it's the simplest way to what data is coming from the REST API for a particular request - Ivy used it just today to debug an issue, and I think it's probably used pretty regularly in dev. The privacy concern is partly limited by the fact that we truncate the 'body' of the response to 256 characters, and I think it's probably okay but imperfect to have this data in the logs - in theory it's as protected as any of our other prod data (and less of a privacy concern than QLing in my opinion), and you have to do some specific log parsing to expose it more generally. Definitely happy to have a followup discussion or just bring it up in Slack or something

mmcgahan:
Going to merge this now so it's not hanging around through hackathon

src/plugins/api-proxy/util/receive.js
willh-meetup: 👍

PR #331, opened 2017-07-24
add gae health checks

mattkime:
pro-web needs them as well.


mattkime:
@mmcgahan test added!

PR #332, opened 2017-07-24
`large` and `small` duotoned image urls for group object

mmcgahan:
`group.duotoneUrl` is now an object instead of a string, and has a `large` and `small` photo URL

This is for Foundation, who would like a higher-resolution duotoned image for the group context header/hero. The `GroupCard` will need to be updated to point to the `small` URL in mup-web.

### DIFF
--- a/src/util/apiUtils.js
+++ b/src/util/apiUtils.js
@@ -451,7 +451,10 @@ export const groupDuotoneSetter = duotoneUrls => group => {
 		);
 	const duotoneUrlRoot = duotoneKey && duotoneUrls[duotoneKey];
 	if (duotoneUrlRoot && photo.id) {
-		group.duotoneUrl = `${duotoneUrlRoot}/${photo.id}.jpeg`;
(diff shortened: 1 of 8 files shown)


src/util/apiUtils.js
```diff
@@ -451,7 +451,10 @@ export const groupDuotoneSetter = duotoneUrls => group => {
 		);
 	const duotoneUrlRoot = duotoneKey && duotoneUrls[duotoneKey];
 	if (duotoneUrlRoot && photo.id) {
-		group.duotoneUrl = `${duotoneUrlRoot}/${photo.id}.jpeg`;
+		group.duotoneUrl = {
+			small: `${duotoneUrlRoot.small}/${photo.id}.jpeg`,
```
eilinora: curiosity why do we use `jpeg` vs just `jpg`?

mmcgahan: photo scaler behavior - `.jpg` won't actually be generated/served for some reason

PR #333, opened 2017-07-30
WP-436 click tracking module

mmcgahan:
Click tracking has two parts: writing the click data to Redux state and then to a cookie in browser code, and reading the click data from the cookie in server code. This module consolidation brings together all relevant modules into `src/plugins/tracking/`.


src/plugins/index.js
```diff
@@ -106,6 +107,12 @@ export function getActivityTrackingPlugin({ agent, isProd }) {
 	};
 }
 
+export function getClickTrackingPlugin() {
```
mattkime: just curious why this little preamble is here. i see that other plugins have it as well but it remains mysterious.

mmcgahan: It's mostly just to have a consistent pattern for registering the plugin and keeping the main `getPlugins` function succinct - calling a function for each plugin from the `getPlugins` function just makes it easier to add/change 'options' for each plugin over time.

mmcgahan: also, having the 'getters' makes it slightly easier to mix-and-match plugins for integration testing.

src/plugins/tracking/util/clickState.js
```diff
@@ -0,0 +1,56 @@
...
+
+/*
+ * This module provides utilities for managing click tracking data in Redux
+ * state an saving that data to a browser cookie
+ */
+
+export const COOKIE_NAME = 'click-track'; // must remain in sync with Meetup Classic implementation
+
+const BrowserCookies = JSCookie.withConverter({
+	read: (value, name) => value,
```
willh-meetup: We should drop `name` from here and from `write` since we don't use that variable.

src/plugins/tracking/util/clickState.js
```diff
@@ -0,0 +1,56 @@
...
+export const setClickCookie = clickTracking => {
+	const domain = window.location.host.replace(/[^.]+/, ''); // strip leading subdomain, e.g. www or beta2
+	const cookieVal = JSON.stringify(clickTracking);
+	BrowserCookies.set(COOKIE_NAME, cookieVal, { domain });
+};
+
+export const CLICK_TRACK_ACTION = 'CLICK_TRACK';
+export const CLICK_TRACK_CLEAR_ACTION = 'CLICK_TRACK_CLEAR';
+
+export const actions = {
```
willh-meetup: I noticed later we're renaming `actions` and `reducer` as we import them. 

Is this a naming convention thing, or something that needs to happen?

src/plugins/tracking/click.js
```diff
@@ -0,0 +1,37 @@
+// @flow
+import clickReader from './util/clickReader';
+
+export const CLICK_PLUGIN_NAME = 'mwp-click-tracking';
+
+export function onPreHandlerExtension(request: HapiRequest, reply: HapiReply) {
+	try {
+		clickReader(request, reply);
+	} catch (err) {
+		console.error(err);
```
mmcgahan: I don't remember, but the `request.server.app.logger` logging should be fine

src/plugins/tracking/util/clickState.js
mmcgahan: ok - I was sort of thinking that it might be useful for future maintainers to know that these callbacks do have access to the cookie name if needed, but that's easy to find out from the JSCookie docs if this needs to be tweaked for some reason.

src/plugins/tracking/util/clickState.js
mmcgahan: I'm not really sure what the 'best' solution is for this since this module lives inside the 'click'-related code, so it's implicit that the `actions` are '`clickActions` and the `reducer` is a `clickReducer` - I could rename but it's a little redundant from the perspective of the module. Also, it makes sense if the whole module is imported and the named exports get referenced from a top-level object, e.g.

```js
import * as clickState from '.../clickState';

clickState.actions.click();
clickData: clickState.reducer(state);
```

Since the name of the typically ends up being where it stores its corresponding `state` data, the consumer app will probably always want to alias it in some way just so that it has control over where the data appears, but the name of the `actions` export could be made more unique. I'm ambivalent.

willh-meetup: No, I think this is fine, just wanted to make sure I fully understood.

PR #335, opened 2017-08-02
WP 415 selector helpers moved from mup-web

mmcgahan:
No new code here, just some Flow types - feel free to review that, but everything else is verbatim from mup-web.

Once this merges, I'll set up a PR in mup-web to remove the existing selectorsHelper.js script and update the import paths for modules that currently rely on it.

### DIFF
--- a/flow-typed/platform.js
+++ b/flow-typed/platform.js
@@ -1,6 +1,7 @@
 // @flow
 declare var Intl: Object;
 
+declare type NonNull = string | number | boolean | {} | [];
 declare type Params = { [string]: string | number };
 
 declare type FluxStandardAction = {
@@ -11,6 +12,8 @@ declare type FluxStandardAction = {
 };
 
 declare type Reducer = (state: ?Object, action: FluxStandardAction) => Object;
+// Selector is paramaterized by its output type
+declare type Selector<T> = (state: ?Object) => T;
 
 // API query structure
 declare type Query = {
@@ -24,9 +27,10 @@ declare type Query = {
 	},
 };
 
+declare type QueryResponseValue = Object | Array<Object>;
 declare type QueryResponse = {
 	ref: string,
-	value: Object | Array<Object>,
+	value: QueryResponseValue,
 	type?: string,
 	flags?: Array<string>,
 	meta?: Object,
@@ -62,29 +66,32 @@ declare type Match = {
 
 declare type MatchedRoute = { route: PlatformRoute, match: Match };
 declare type MatchPathOptions = {
-  path: string,
-  exact?: boolean,
-  strict?: boolean,
-}
+	path: string,
+	exact?: boolean,
+	strict?: boolean,
+};
 
 declare module 'react-router-dom/matchPath' {
-	declare function matchPath(pathname: string, options: MatchPathOptions): null | Match;
+	declare function matchPath(
+		pathname: string,
+		options: MatchPathOptions
+	): null | Match;
 	declare module.exports: typeof matchPath;
 }
 
 declare type LocationShape = {
-  pathname?: string,
-  search?: string,
-  hash?: string,
-  state?: any,
-}
+	pathname?: string,
+	search?: string,
+	hash?: string,
+	state?: any,
+};
 
 declare class RouterRedirect extends React$Component {
 	props: {
 		to: string | LocationShape,
 		push?: boolean,
-	}
+	},
 }
 declare module 'react-router-dom/Redirect' {
-	declare export default typeof RouterRedirect;
+	declare export default typeof RouterRedirect
 }
--- a/src/reducers/selectors.js
+++ b/src/reducers/selectors.js
@@ -1,0 +1,124 @@
+// @flow
+
+import { createSelector } from 'reselect';
+
+export const EMPTY_OBJ = {};
+export const EMPTY_AR = [];
+
+/**
+ * Checks a response object for errors and returns them
+ * @param  {Object} responseValue Response value from Api
+ * @return {Object} an object containing errors parsed from the response
+ */
+export const processErrors = (responseValue: ?QueryResponse): Object => {
+	if (!responseValue || !responseValue.error) {
+		return {};
+	}
+	return responseValue;
+};
+
+export const processErrorsMaybe = (responseValue: ?QueryResponse) => {
+	const processed = processErrors(responseValue);
+	return Object.keys(processed).length ? processed : null;
+};
+
+// Checks if object is empty
+export const isEmpty: Selector<boolean> = state => {
+	if (!state) {
+		return true;
+	}
+	return Object.keys(state).length === 0;
+};
+
+// Checks if object is has errors
+export const hasErrors: Selector<boolean> = state =>
+	Object.keys(processErrors(state)).length > 0;
+
+// Combo check if the object is either empty or has errors
+export const hasValidValue: Selector<boolean> = state =>
+	!isEmpty(state) && !hasErrors(state);
+
+/*
+ * returns the QueryResponse.value, or an object containing a single `error`
+ * property. An empty response returns the defaultValue.
+ */
+export const getValue = (
+	resp: ?QueryResponse,
+	defaultValue: mixed = EMPTY_OBJ
+): mixed => {
+	if (!resp || isEmpty(resp)) {
+		return defaultValue;
+	}
+	if (hasErrors(resp)) {
+		return resp; // this is an object like { error, meta }
+	}
+	return resp.value; // the 'unwrapped' value
+};
+
+/**
+ * Returns a property from response.value in state. If response.error or
+ * response is empty, returns default value
+ */
+export const getProperty = (
+	resp: ?QueryResponse,
+	prop: string,
+	defaultValue: mixed = EMPTY_OBJ
+): mixed => {
+	const value = getValue(resp, defaultValue);
+
+	if (value === defaultValue || value.error) {
+		return value;
+	}
+	if (value instanceof Object && value[prop]) {
+		return value[prop];
+	}
+
+	return defaultValue;
+};
+
+/**
+ * Pulls the value out of the state object provided or if error returns default value
+ * @param  {Object}  defaultValue    fallback value if value is inValid
+ * @param  {Object}  selector    selector to be pulled off state object
+ * @return {Object}  returns selected value
+ */
+export const getSelectOrFallback = (
+	defaultValue: mixed,
+	selector: Selector<mixed> = s => s
+) => (state: ?Object) => {
+	if (isEmpty(state)) {
+		return defaultValue;
+	}
+	return selector(state);
+};
+
+// selects the inFlight part of state
+const getInFlightResp = state => state.api.inFlight;
+
+export const getInFlight = createSelector(
+	getInFlightResp,
+	getSelectOrFallback(EMPTY_AR, resp => resp)
+);
+
+// Reads response and builds error object based on response.error
+export const getResponse = (
+	resp: ?QueryResponse,
+	type: string,
+	defaultValue: mixed = EMPTY_OBJ
+) => {
+	// empty check
+	if (!resp || isEmpty(resp)) {
+		return defaultValue;
+	}
+	// error object
+	if (resp.error) {
+		return {
+			error: {
+				type,
+				message: resp.error,
+			},
+		};
+	}
+
+	return resp.value;
+};
--- a/src/reducers/selectors.js
+++ b/src/reducers/selectors.js
@@ -3,22 +3,38 @@
 import { createSelector } from 'reselect';
 
 export const EMPTY_OBJ = {};
-export const EMPTY_AR = [];
+export const EMPTY_ARR = [];
 
-// Checks a response object for errors and returns them
-export const processErrors = (responseValue: ?QueryResponse): Object => {
-	if (!responseValue || !responseValue.error) {
-		return {};
+/*
+ * Checks a query response object for an error prop and the response if there
+ * are errors, or an empty object if no errors are found.
+ *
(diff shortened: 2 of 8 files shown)


src/reducers/selectors.js
```diff
@@ -0,0 +1,114 @@
+// @flow
+
+import { createSelector } from 'reselect';
+
+export const EMPTY_OBJ = {};
+export const EMPTY_AR = [];
+
+// Checks a response object for errors and returns them
+export const processErrors = (responseValue: ?QueryResponse): Object => {
+	if (!responseValue || !responseValue.error) {
```
nlaz: I think the `processErrors` method is left over from a time ages ago where errors were nested in the `value` object and we had to process multiple requests for any possible error. Now errors match 1 to 1 with requests and the `error` isn't nested in value anymore, which leads to confusing behavior with this method. 

For example,
```js
const response = { ref: 'REF', error: 'ERROR', meta: {} };
processErrors(response); // returns `{ ref: 'REF', error: 'ERROR', meta: {} }`
processErrors({}); // returns `{}`
```
It should return just the error right? Now all we need to do is check if `error` is present and return the error. Also, the name `responseValue` is confusing now that we have distinguished between the response (ie. `resp`) and the response's value object.

src/reducers/selectors.js
```diff
@@ -0,0 +1,114 @@
...
+// selects the inFlight part of state
+const getInFlightResp = state => state.api.inFlight;
+
+export const getInFlight = createSelector(
+	getInFlightResp,
+	getSelectOrFallback(EMPTY_AR, resp => resp)
+);
+
+// Reads response and builds error object based on response.error
+export const getResponse = (
```
nlaz: What is the difference b/t `getResponse` and `getValue`? It is not obvious to me what the use case for each method is. Maybe we can consolidate them?

src/reducers/selectors.js
```diff
@@ -0,0 +1,114 @@
...
+ */
+export const getValue = (
+	resp: ?QueryResponse,
+	defaultValue: mixed = EMPTY_OBJ
+): mixed => {
+	if (!resp || isEmpty(resp)) {
+		return defaultValue;
+	}
+	if (hasErrors(resp)) {
+		return resp; // this is an object like { error, meta }
```
nlaz: Why would this return a response object when there are errors? If I call `getValue`, I would expect a response's value object or the default value.

nlaz:
Can we add a version of `getValue` that accepts a ref and returns that ref's value from state?

Something like
```
export const getValueByRef = (state, REF, defaultValue = EMPTY_OBJ) => {
   const resp = state.api[REF] || defaultValue;
   return resp.value || defaultValue;
};
// ... somewhere else
const mapStateToProps = createStructuredSelector({
	events: getValueByRef(UPCOMING_EVENTS_REF),
});
```
Not sure how to get this to work inside `reselect` methods since it passes arguments to the selectors, but that's the sort of thing I'm looking for.

nlaz:
Also, I would _love_ if we could beef up the docs here. Not sure what some of these methods are used for.

src/reducers/selectors.js
```diff
@@ -0,0 +1,114 @@
...
+
+/*
+ * returns the QueryResponse.value, or an object containing a single `error`
+ * property. An empty response returns the defaultValue.
+ */
+export const getValue = (
+	resp: ?QueryResponse,
+	defaultValue: mixed = EMPTY_OBJ
+): mixed => {
+	if (!resp || isEmpty(resp)) {
```
nlaz: Why do we need the extra `isEmpty` check? If the returned value is an array, it will return an empty object if `defaultValue` is not set. Seems like it could lead to weird bugs?

mmcgahan:
@nlaz I'm happy to do a bit of syntax/var name cleanup here, but actual functionality changes (and method removal) are probably out of scope for this ticket, which is specific to moving the code, not changing the implementation. I'll try to add some docs, but since I didn't write most of these, I think that might be a good followup ticket to file for @eilinora.

If I had to re-implement these helpers, I think they could benefit from locking down the type definitions a little more fully - there are a number of places where the returned value can be any number of things, which makes the selectors a little tricky to reason about and use, but I don't want to risk breaking any existing usage from a file-organization ticket :)

src/reducers/selectors.js
```diff
@@ -0,0 +1,114 @@
+// @flow
+
+import { createSelector } from 'reselect';
+
+export const EMPTY_OBJ = {};
+export const EMPTY_AR = [];
```
mmcgahan: yes, I'll make a note to update the corresponding `import`s in mup-web

src/reducers/selectors.js
mmcgahan: This was always a confusing method to me as well. It's used in 11 places in mup-web (some comments, conversations, photos code), so I'm not confident in removing/changing its behavior in a way that will guarantee to work with those implementations - definitely something that could/should be cleaned up in a separate ticket, though.

src/reducers/selectors.js
```diff
@@ -0,0 +1,114 @@
...
+
+// Checks a response object for errors and returns them
+export const processErrors = (responseValue: ?QueryResponse): Object => {
+	if (!responseValue || !responseValue.error) {
+		return {};
+	}
+	return responseValue;
+};
+
+export const processErrorsMaybe = (responseValue: ?QueryResponse) => {
```
mmcgahan: in `communicationSelectors`, there is a function that needs to return a falsey value if there are no errors, whereas `processErrors` always returns something truthy (`{}`). I'd like to see this go away just like `processErrors` to avoid the confusion.

src/reducers/selectors.js
```diff
@@ -0,0 +1,114 @@
...
+ * property. An empty response returns the defaultValue.
+ */
+export const getValue = (
+	resp: ?QueryResponse,
+	defaultValue: mixed = EMPTY_OBJ
+): mixed => {
+	if (!resp || isEmpty(resp)) {
+		return defaultValue;
+	}
+	if (hasErrors(resp)) {
```
mmcgahan: looks reasonable.

src/reducers/selectors.js
mmcgahan: 100% agree, but this is used in a number of places in mup-web that presumably depend on this 'mixed response type' behavior.

src/reducers/selectors.js
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	return defaultValue;
+};
+
+/*
+ * reads the value out of the state object provided using an optional selector,
+ * or, if response is empty, returns the provided default value
+ */
+export const getSelectOrFallback = (
```
mmcgahan: capitalization is correct - this function returns ('gets') a function that applies a selector ('selects') or falls back to the provided default.

src/reducers/selectors.js
mmcgahan: Followup for @eilinora - out of scope here.

flow-typed/platform.js
```diff
@@ -24,9 +27,10 @@ declare type Query = {
 	},
 };
 
+declare type QueryResponseValue = Object | Array<Object>;
```
mmcgahan: not yet - relatively easy to write, though.

src/util/testUtils.js
```diff
@@ -108,3 +109,20 @@ export function testCreateStore(createStoreFn) {
...
+ * @param {String} ref Reference to use in object creation
+ * @param {Object} value value to use in object creation
+ * @return {Object} mock object
+ */
+export const generateMockState = (ref, value = { id: '🦄' }) => ({
+	api: {
+		...MOCK_APP_STATE.api,
+		[ref]: {
+			ref: ref,
+			meta: { statusCode: 204 },
```
mmcgahan: applying defaults for multiple arguments gets a little awkward when you want to override one default but not another, so I think I'll need to change the signature of this function from `(ref, value?)` to `({ ref, value?, meta? })` - I'll make a note to update that breaking change in mup-web as well.

src/reducers/selectors.js
mmcgahan: `isEmpty` checks for more than just `!resp`, but I had to explicitly add `!resp` here because Flow wasn't smart enough to know that `isEmpty` handles the null case, unfortunately. A little more verbose, but I've convinced myself that is also is more obvious to someone reading the code without being too verbose, which is a good thing.

yeah - the expectation is that if you are using looking for an array value, you will need to provide the `defaultValue`. I wouldn't mind making the `defaultValue` a required argument, but that would be a followup ticket with a breaking change.

mmcgahan:
@nlaz I think I've addressed the issues you raised that don't affect functionality, and added some code docs to flag some implementation concerns

nlaz:
@mmcgahan I see where you are coming from. If the ticket is to simply migrate the code in `mup-web` to the `meetup-web-platform` codebase then I'm satisfied that it can be followed up on later. 

However, I'm a little confused by the breaking changes argument. Introducing any implementation changes now in this initial migration while deprecating the old methods in `mup-web` seems like it would cause the least amount of headache while rolling it out. Let me know what your thoughts are. I'm happy to put some love into this to make it work well for everyone.

src/reducers/selectors.js
eilinora: I'd argue for not making it required as we more often than not expect an object than an array.

src/util/testUtils.js
```diff
@@ -108,3 +109,24 @@ export function testCreateStore(createStoreFn) {
...
 	});
 }
+
+/**
+ * Useful function when generating fake state objects
+ * @param {String} ref Reference to use in object creation
+ * @param {Object} value value to use in object creation
+ * @return {Object} mock object
+ */
+export const generateMockState = ({
```
eilinora: there were also some generateMockError helpers which are handy

mmcgahan:
@nlaz My sense is that refactoring these methods will require a careful review of current usage and some thorough planning of how to set up a coherent set of complementary helpers that compose well, followed by case-by-case refactoring of the code that consumes the helpers in mup-web. Hopefully the existing unit tests are sufficient to catch refactoring errors, but whenever you're dealing with state, there tend to be a lot of edge cases so I'm not very well set up to do that kind of refactoring at the moment.

Since this ticket just changes the import path, the required changes in mup-web are minimal - I'm comfortable making that kind of change myself without input from another team. Functional changes would definitely require a deeper level of support.

If you feel that the helpers should change significantly before being used more widely, we can close this PR without merging. The only reason to move them into MWP is to allow pro-web to use them, so if they're going to change we should probably wait until they're 'stable'.

mmcgahan:
@eilinora @nlaz would you two be able to decide on an up/down vote for this? I think the main benefit was being able to share these helpers with pro-web, but I agree that a review/refactor would be useful so it's probably best to wait until the helpers stabilize in mup-web before we bring them into MWP. If they are 'stable enough' or there is demand for them from pro-web, then I can wrap up this PR and merge, but you both have a better perspective on that than I do.

If there's not pressing demand from pro-web and you don't think you'll be able to review the current implementation in mup-web for awhile, I'll just close and leave things be for now.

nlaz:
I probably won't be using these helpers until they are more stable and updated so there is no pressure from me. I'll vote down for these changes for now but I will try to get a PR to update `mup-web` (or here) in by next week. I created a placeholder ticket so I can add to my next sprint. I don't want to stand in the way of progress so feel free to proceed however and I'll support when I have some bandwidth to add here.

eilinora:
@mmcgahan i still think these are useful as it's helped me from not making new functions over and over again like we were. Yes they can be improved but still usable

mmcgahan:
@eilinora ok - I think I'll close this PR for now until usage can be better documented or refactored, unless you know of helpers that pro-web would like to start using right away. I think `getValue`, `getProperty`, and `getResponse` could be clarified/consolidated in a way that would make it easier to understand the cases in which each one should be used, and ideally refactored in a way that makes their return 'types' more predictable.

eilinora:
They had specifically asked to make these more available :(

eilinora:
If anything it would be good to define the jira tasks around the work around what it would take to make this migratable.

mmcgahan:
@eilinora the type confusion that I'm referring to is more related to Flow types - much more specific than `Object` vs `Array` JS types. Thinking with Flow types helps clarify why something like `getValue` is hard to work with:

the function takes a `resp` argument that is an object with this type:

```js
{
  value: Object | Array,
  error?: string, // optional
  meta: Object
}
```

the `defaultValue` argument is literally _anything_ (`any`)

This means that the return value could be one of 3 things:

1. `resp.value` - Object or Array corresponding to the 'real' API response nested in the original `resp`
2. `{ value, error, meta }` - the original `resp` argument in full if there is an `error` prop - `value` will be an object containing individual errors, `error` will be a string like `'Bad Request'`. `meta` will contain all the usual `meta` info from that specific REST API request.
3. `any` - whatever `defaultValue` is. This could be used to simulate API data, or it could be used in some other way.

This is complex to reason about, and is not clear from the name `getValue`, because 'value' could mean a number of different things. In the error case, `value` is a nested property, but it's the top-level value of the non-error case. It appears that the caller would always need to then inspect the return value to determine what to do with it, because `resp.value`, `error`, and `default` return values that should generally be handled in different ways.

`getProperty` has similar challenges, and has code paths that may not end up getting a property at all, which is confusing.

`getResponse` is very similar to `getValue`, but creates a custom `error` response, which makes the return types more complicated - it's not clear when I would choose `getResponse` instead of `getValue`, which should be clarified either through the implementation or documentation.

IMO the complexity of these selectors makes them difficult to reason about and therefore prone to being used incorrectly, leading to bugs. I think it would be easier to use these selectors if they had narrower scope and didn't try to manage default values - you might want to keep `getSelectOrFallback` as the preferred helper for applying fallback behavior across the board.

PR #338, opened 2017-08-07
WP-440: shared PageWrap (+ Browser polyfill & NewRelic polyfill)

sadafie:
Fixes https://meetup.atlassian.net/browse/WP-440

Move PageWrap from `mup-web` to platform repo in order to share across consumer repos including `pro-web`.


src/util/browserPolyfill.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * @param {String} localeCode the 'xx-XX' language code for the app
+ * @return {String} the polyfill.io cdn string
+ */
+export function polyfillServiceUrl(localeCode) {
+	const features = [
```
mattkime: we should use a default set - `default-3.6` 

Then we can remove - 
Promise
URL
Array.from
String.prototype.endsWith
String.prototype.startsWith


docs - https://polyfill.io/v2/docs/features/#default-sets

sadafie: Yes, would make a good follow up! I'd like to limit the scope of this and there are more changes coming in.

sadafie:
Need some help --

I added a `src/assets/` directory for the logos, but I need them to get built to `lib/`. Currently the build step for this repo is just a babel command. What's the best way of building these assets with the new `mwp-cli`?

mattkime:
@sadaf are you having a problem when you try to use these changes in mup-web/pro-web?

sadafie:
Yes, the `PageWrap` module that gets built to `meetup-web-platform/lib/components` errors because it can't resolve the asset modules. Specifically:

```
>> browserApp bundle built
Child
       1264 modules

    ERROR in ../meetup-web-platform/lib/components/PageWrap.js
    Module not found: Error: Can't resolve '../assets/favicon.ico' in '/Users/sadaf/git_repo/meetup-web-platform/lib/components'

    ERROR in ../meetup-web-platform/lib/components/PageWrap.js
    Module not found: Error: Can't resolve '../assets/logos/m_swarm_120x120.png' in '/Users/sadaf/git_repo/meetup-web-platform/lib/components'

    ERROR in ../meetup-web-platform/lib/components/PageWrap.js
    Module not found: Error: Can't resolve '../assets/logos/m_swarm_128x128.png' in '/Users/sadaf/git_repo/meetup-web-platform/lib/components'

    ERROR in ../meetup-web-platform/lib/components/PageWrap.js
    Module not found: Error: Can't resolve '../assets/logos/m_swarm_152x152.png' in '/Users/sadaf/git_repo/meetup-web-platform/lib/components'

    ERROR in ../meetup-web-platform/lib/components/PageWrap.js
    Module not found: Error: Can't resolve '../assets/logos/m_swarm_180x180.png' in '/Users/sadaf/git_repo/meetup-web-platform/lib/components'

    ERROR in ../meetup-web-platform/lib/components/PageWrap.js
    Module not found: Error: Can't resolve '../assets/logos/m_swarm_196x196.png' in '/Users/sadaf/git_repo/meetup-web-platform/lib/components'
webpack: Failed to compile.
```

Which makes sense because the new `src/assets/` directory currently isn't being build to `meetup-web-platform/lib`.

sadafie:
In all honesty, my preference would be to not have the assets in this repo but after a brief conversation with adamD he says it's the best place for now (i.e., i'm not sure if we have any other common repo where it could live).

mattkime:
i suspect its erroring because babel doesn't know how to resolve the require statements that depend upon loaders.

this might do the job - https://github.com/morlay/babel-plugin-transform-require-ignore

sadafie:
@mattkime I'm not sure I understand what that plugin does.... it seems to claim to ignore extensions but I'm not sure how that would help here

sadafie:
Oh, but maybe this might work: https://babeljs.io/docs/usage/cli/#babel-copy-files

`--copy-files` for files that aren't compiled.

mattkime:
ah, gotcha. that might make more sense than what i suggested.

sadafie:
`--copy-files` works for assets! LMK if you have any questions or concerns about doing it this way but it seems pretty straightforward. Only drawback I can see is that it might copy over extra files that are unnecessary but aside from overall size that seems ok to me.

Next error:

```
Error: Cannot find module 'raw-loader!swarm-icons/dist/sprite/sprite.inc'
    at Function.Module._resolveFilename (module.js:485:15)
    at Function.Module._load (module.js:437:25)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/Users/sadaf/git_repo/meetup-web-platform/lib/components/PageWrap.js:46:18)
```

mattkime:
it might be good to copy the asset dir outside of babel so we're not copy over unneeded code but i don't have a strong opinion. 

i wonder how webpack is resolving `raw-loader!swarm-icons/dist/sprite/sprite.inc` - either relative to the package or the project.

src/components/PageWrap.jsx
```diff
@@ -0,0 +1,124 @@
+import React from 'react';
+import Helmet from 'react-helmet';
+import PropTypes from 'prop-types';
+
+import { polyfillServiceUrl } from '../util/browserPolyfill';
```
mmcgahan: In working to make things more modular in MWP, I've realized that almost everything in `src/util/` would be better placed somewhere else (usually closer to the scripts that consume the utils).

In this case, I would recommend a new `/src/components/util/` (or `/src/components/pageUtils/`) directory to put `browserPolyfill.js` and `newrelicBrowser.js` - since those are only used in this component, that organization keeps those modules better 'hidden' from the rest of the application.

I have a modularization ticket to move all the PlatformApp/BrowserApp-related components to their own module, so this code will move along with that sometime soon, so keeping everything close together now will help with that.

---

thinking about this a bit more, since there are quite a few files that need to move around with `PageWrap`, I think I'd recommend making a single directory that contains _all_ the related scripts, e.g. `src/components/page-wrap`, so that the related files can be moved around together easily - no need to spread things among `src/components/`, `src/util/` and `src/asset/`, which makes the 'system' unwieldy

sadafie: Sure, i can move the files.

sadafie:
>I assume pro-web should use the same logo files? If not, we might want to make those configurable as props in PageWrap

They are not in the pro-web version, however, I'm not sure if there omission is intentional or if the changes just never got pulled into that repo. @mattkime @mikespencer @acmei   Do any of you know if the favicon logos in your version of PageWrap was intentionally removed?

sadafie:
Just realized, this repo doesn't use webpack at all correct?

Anyone have any suggestions for alternatives to `file-loader` and `raw-loader`? I tried using https://github.com/quadric/babel-plugin-inline-import but so far i'm unsuccessful.

mmcgahan:
> Just realized, this repo doesn't use webpack at all correct?

Correct, I should have flagged that. I think the best short-term option is to require that _all_ static asset paths be passed in as `prop`s from the consumer app rather than baking them into this component - trying to work around that limitation is probably not worth the effort right now

src/components/PageWrap.jsx
```diff
@@ -0,0 +1,124 @@
...
+					/>
+					<link rel="shortcut icon" sizes="196x196" href={swarmIcon196x196} />
+
+					<script type="text/javascript" src={polyfillServiceUrl(localeCode)} />
+					<script type="text/javascript">
+						{newrelicBrowserJS}
+					</script>
+				</Helmet>
+
+				<div
```
akdetrick: For the icon sprite, the thing we really need to do here is dump the entire contents of `swarm-icons/dist/sprite/sprite.inc` into this div.

mattkime:
@sadafie 

re: the loaders. 

did you try my previous recommendation? it would just ignore imports relying on loaders. useful for building this package which will be consumed by an app that will provide the loaders.

mikespencer:
> >I assume pro-web should use the same logo files? If not, we might want to make those configurable as props in PageWrap

> They are not in the pro-web version, however, I'm not sure if there omission is intentional or if the changes just never got pulled into that repo. @mattkime @mikespencer @acmei   Do any of you know if the favicon logos in your version of PageWrap was intentionally removed?

As far as I know the favicon was not intentionally removed. I'm guessing those changes just never made in to the app for whatever reason

src/components/PageWrap.jsx
akdetrick: `raw-loader` and `dangerouslySetInnerHTML` are only a means to this end, not a requirement for the icon sprite

mmcgahan:
@mattkime the consumer apps are set up to use the pre-compiled (`/lib/`) version of MWP, so `require-loader` and `file-loader` will not be processed by webpack (MWP is an `external` lib). This saves a couple minutes of compile time in Travis and reduces some build uncertainty, so I'm reluctant to start transpiling MWP as part of the consumer app builds.

mmcgahan:
Since @sadafie is out the rest of the day and would like to wrap this up before I sign in tomorrow, here are the options:

1. Keep the static assets and their `file-loader`/`raw-loader` `require` statements as-is in this PR
  - We'll need to figure out the right babel configuration so that it doesn't stumble on these import statements, and copies the source files to `/lib/` or whatever
  - in MWP-CLI, update the build configuration to transpile `meetup-web-platform` imports just like the rest of the application so that the loaders will be applied
  - we'd also probably want to make the webpack loaders (and webpack) a `peerDependency` of MWP
2. treat static assets paths as `prop`s to `PageWrap` provided by the consumer apps
  - PageWrap itself becomes simpler, `AppContainer` gets a little bigger along the lines of what is happening in this PR https://github.com/meetup/mup-web/pull/1416/files#diff-60cd712125cee88fc20d9aaddb0e4c51
  - No build changes to either repo

I favor option 2 because build changes (1) introduce much more uncertainty (i.e. are more bug-prone and require more 'magic' in the form of plugins) than `prop` changes, and IMO it keeps the responsibilities of the two sides more consistent - the applications are responsible for webpack-related code including compiling static assets, while the platform provides 'dumb' components that primarily structure the data/assets coming from the apps rather than directly providing data/assets - it's a loose principle, but I think it's useful.

As a followup, I think it would be relatively straightforward to write a `mwp-static` package that could contain shared static assets like the logos. From using Lerna in my hackathon, I think it about to become trivially easy to publish new packages from MWP

sadafie:
Thanks @mattkime @mmcgahan @akdetrick @mikespencer !!

I'm now remembering how this repo is compiled based on all the feedback. I think mostly I forgot that there wasn't a webpack build with this repo until after I had moved all the assets over.

I agree that Option 2 is easier at the moment. Unfortunately, it will make `AppContainer` a bit heavier because not only is it the static asset that needs loading but also meta data that lives in attributes for their unique `link` tags. i.e. `rel`, `sizes`, etc.:

```
+					<link
 +						rel="apple-touch-icon"
 +						sizes="180x180"
 +						href={swarmIcon180x180}
 +					/>
```

>As a followup, I think it would be relatively straightforward to write a mwp-static package that could contain shared static assets like the logos.

I agree. Based off of Mike Spencer's feedback, seems like the both pro-web and mup-web will rely on the same set of icon sprites and favicons, so if we can either consume or even build+consume (i.e. a shared module that exports all the link tags) from a central place, we'd save on the redundancy.

I'll get on this asap and hopefully we can merge today. Thanks!

sadafie:
Ready for final review. Thanks!

src/components/pageWrap/PageWrap.jsx
```diff
@@ -0,0 +1,100 @@
...
+		const lang = localeCode.substring(0, 2);
+
+		return (
+			<div
+				id="root"
+				className={`column lang_${lang}`}
+				style={{ minHeight: '100vh' }}
+			>
+				<Helmet defaultTitle="Meetup" titleTemplate="%s - Meetup">
+					{css && this.renderHeadLinks(css)}
```
mmcgahan: I think we can make this component simpler and easier to work with by avoiding having separate `css`, and `swarmLogos` props - instead, what do you think about just accepting a single `<Helmet>` element as a prop, which can contain whatever _additional_ `<head>` content that an application needs? This `PageWrap` component would then render its 'standard'/base `<Helmet>` content, and then render the passed-in Helmet `prop` as a sibling - I believe that the contents of both elements will be merged correctly by Helmet.

The benefits of that approach are that `PageWrap` can be simpler because it doesn't have to parse two different props using `renderHeadLinks`, and the parent component can have as much custom control over `<head>` as it wants without ever having to modify the `<PageWrap>` interface to support more props.

```jsx
// parent component
const head = (
  <Helmet>
    <link ...swarm stuff.../>
    <link ...swarm stuff.../>
    <link ...css stuff.../>
  </Helmet>
);
return (
  <PageWrap head={head}>
    ...
  </PageWrap>
);
```

package.json
```diff
@@ -42,6 +42,7 @@
     "chalk": "2.0.1",
     "convict": "4.0.0",
     "electrode-csrf-jwt": "1.2.1",
+    "enzyme": "2.9.1",
```
mmcgahan: enzyme should be a dev dependency

package.json
```diff
@@ -89,7 +91,7 @@
     "jest": "20.0.4",
     "lint-staged": "4.0.1",
     "meetup-web-mocks": "1.0.203",
-    "mwp-cli": "1.0.101",
+    "mwp-cli": "1.1.114",
```
mmcgahan: the `peerDependencies` version should also be updated

src/components/pageWrap/PageWrap.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+					<meta name="viewport" content="width=device-width, initial-scale=1" />
+					<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+					<meta http-equiv="X-UA-Compatible" content="IE=edge" />
+					<meta name="robots" content="index,follow" />
+					<meta
+						name="verify-v1"
+						content="h5EhuAEkLFlZmMxwpH5wnRaoDEmqYCCEUE+FLcrRNvE="
+					/>
```
sadafie: Do you know who would know this or how to follow up and find out?

src/components/pageWrap/PageWrap.jsx
sadafie: Along the same lines I was thinking about just having a single prop that passes the css and icon objects but I wasn't sure if the order of the tags in the head matter.

I'll see if I can get one or the other of the approaches working.

src/components/pageWrap/PageWrap.jsx
mmcgahan: `git blame`. Fine to do as followup

src/components/pageWrap/PageWrap.jsx
mmcgahan: pretty sure order won't matter

sadafie:
PR feedback updated. I'm going to try the idea of a head prop. If it works I'll update the PR.

src/components/pageWrap/newrelicBrowser.js
```diff
@@ -0,0 +1,2 @@
...tElementsByTagName("script")[0];if(setTimeout(u.abort,3e4),!(t&&t.licenseKey&&t.applicationID&&n))return u.abort();f(y,function(n,e){t[n]||(t[n]=e)}),c("mark",["onload",a()+b.offset],null,"api");var e=l.createElement("script");e.src="https://"+t.agent,n.parentNode.insertBefore(e,n)}}function o(){"complete"===l.readyState&&i()}function i(){c("mark",["domContent",a()+b.offset],null,"api")}function a(){return E.exists&&performance.now?Math.round(performance.now()):(s=Math.max((new Date).getTime(),s))-b.offset}var s=(new Date).getTime(),c=t("handle"),f=t(15),u=t("ee"),d=window,l=d.document,p="addEventListener",h="attachEvent",m=d.XMLHttpRequest,w=m&&m.prototype;NREUM.o={ST:setTimeout,CT:clearTimeout,XHR:m,REQ:d.Request,EV:d.Event,PR:d.Promise,MO:d.MutationObserver};var v=""+location,y={beacon:"bam.nr-data.net",errorBeacon:"bam.nr-data.net",agent:"js-agent.newrelic.com/nr-1026.min.js"},g=m&&w&&w[p]&&!/CriOS/.test(navigator.userAgent),b=n.exports={offset:s,now:a,origin:v,features:{},xhrWrappable:g};t(12),l[p]?(l[p]("DOMContentLoaded",i,!1),d[p]("load",r,!1)):(l[h]("onreadystatechange",o),d[h]("onload",r)),c("mark",["firstbyte",s],null,"api");var x=0,E=t(17)},{}]},{},["loader",2,10,4,3]);
```
mmcgahan: this file doesn't pass `yarn lint` - did you do a `--no-verify` when committing it? I'll go ahead and add it to `.eslintignore`, but I'd be interested to find out why the linting safeguard failed

PR #341, opened 2017-08-09
consolidate side-effect handling - ensure `.rewind()` is called on error

mmcgahan:
When there was an error thrown while rendering the application, `Helmet` and a few other 'react-side-effect' components were not getting correctly re-set on the server. When there was an error _inside_ one of these components in the application, all subsequent requests would try to re-render the same erroring instances, and the server could not recover.

This PR wraps the app-rendering call in a `try`/`catch` that will force the react-side-effect classes to re-set in the case of a rendering error. It also consolidates the references to these classes in a `resolveSideEffects` function so that it easier to keep track of this special behavior.

### DIFF
--- a/tests/integration/render.int.js
+++ b/tests/integration/render.int.js
@@ -217,7 +217,32 @@ describe('Full dummy app render', () => {
 			};
 			return server
 				.inject(request)
+				.then(response => {
+					expect(response.statusCode).toBe(500);
+				})
+				.then(() => server.stop())
+				.catch(err => {
+					server.stop();
+					throw err;
+				});
+		}));
+	it('returns 200 OK after a 500 error', () =>
+		start(getMockRenderRequestMap(), {}).then(server => {
+			const errorRequest = {
+				method: 'get',
+				url: '/badImplementation',
+				credentials: 'whatever',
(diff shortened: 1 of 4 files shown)


tests/integration/render.int.js
```diff
@@ -217,7 +217,32 @@ describe('Full dummy app render', () => {
...
+				.then(response => {
+					expect(response.statusCode).toBe(500);
+				})
+				.then(() => server.stop())
+				.catch(err => {
+					server.stop();
+					throw err;
+				});
+		}));
+	it('returns 200 OK after a 500 error', () =>
```
willh-meetup: Nice.

Hopefully this catches other uses of "react-side-effects" where there is potential for a similar error.

mmcgahan: I'll add a followup ticket to write a dedicated set of 'side effect integration tests' to make sure that errors in `NotFound`, `Forbidden`, and `Redirect` are also cleaned up correctly. Unfortunately there's not a clear way to automatically apply the cleanup to any future react-side-effect component, or write a generic test that will check all instances, so I'll need to write separate tests.

PR #342, opened 2017-08-10
WP-441 render components

mmcgahan:
no more `src/components/` directory!

All the remaining platform components have been consolidated into a `src/render/` 'package' that is the starting point for a complete React+Redux platform package.

public modules: `ServerApp`, `BrowserApp`, `Dom`, and `PageWrap`.

README added.


src/render/README.md
```diff
@@ -0,0 +1,33 @@
+# Render
```
willh-meetup: 🙌  docs 🙌

PR #343, opened 2017-08-11
WP 456 'API state' module

mmcgahan:
This PR consolidates the platform modules related to populating API data into app state - the API middleware (sync, cache, mutate [deprecated]), reducers (api, app [deprecated]), and Redux action creators.

External dependencies are relatively minimal - rxjs, redux-observable, mwp-router, mwp-tracking, and neither of the mwp dependencies are circular


src/api-state/reducer.js
```diff
@@ -0,0 +1,126 @@
...
+		case API_RESP_ERROR:
+		case CACHE_SUCCESS: // fall through
+			// each of these actions provides an API response that should go into app
+			// state - error responses will contain error info
+			delete state.fail; // if there are any values, the API is not failing
+			return {
+				...state,
+				...responseToState((action.payload || {}).response),
+			};
+		case API_RESP_FAIL:
```
eilinora: How does fail differ from error? I see this becomes a higher level state item but curious about why.

src/api-state/reducer.js
```diff
@@ -0,0 +1,126 @@
...
+			return { ...state, fail: action.payload };
+		case API_RESP_COMPLETE: {
+			// allways called - clean up inFlight
```
eilinora: Nitpick typo allways->always

src/render/components/Dom.jsx
```diff
@@ -33,10 +33,9 @@ const DOM = props => {
 		appMarkup = '',
 		assetPublicPath,
 		baseUrl,
-		clientFilename,
 		head,
 		initialState = {},
-		scripts = [],
+		scripts,
```
eilinora: If scripts aren't passed I see we would get prop errors but would this _maybe_ cause server errors of empty?

src/router/routeActionCreators.js
```diff
@@ -0,0 +1,9 @@
+export const LOCATION_CHANGE = '@@router/LOCATION_CHANGE';
+export const SERVER_RENDER = '@@router/INITIAL_RENDER';
```
eilinora: Rando question is the @@ thing a like stand industry thing?

src/render/components/Dom.jsx
mmcgahan: definitely, which is a good thing - there's no way to gracefully handle that kind of programmer error, so best to crash hard.

src/api-state/reducer.js
```diff
@@ -0,0 +1,126 @@
...
+import {
+	API_REQ,
+	API_RESP_SUCCESS,
+	API_RESP_ERROR,
+	API_RESP_FAIL,
+	API_RESP_COMPLETE,
+} from './sync/apiActionCreators';
+import { CACHE_SUCCESS } from './cache/cacheActionCreators';
+
+type ApiState = {
```
mmcgahan: Yeah it would definitely be worth creating a 'libdef' that is specific to this module - you have to put shared type definitions in a particular place (`/flow-typed/`) and prefix them with `declare` but it's pretty straightforward and would definitely be useful here - will do

src/api-state/index.js
```diff
@@ -1,8 +1,23 @@
...
 
-import getSyncEpic from '../epics/sync';
-import getCacheEpic from '../epics/cache';
-import { postEpic, deleteEpic } from '../epics/mutate'; // DEPRECATED
+import getSyncEpic from './sync';
+import getCacheEpic from './cache';
+import { postEpic, deleteEpic } from './mutate'; // DEPRECATED
+
+// export specific values of internal modules
+export {
```
mmcgahan: yes - the `index.js` in the root folder for the module is effectively the 'public interface' for the module. The expectation is that dependent modules will _only_ use `require('./src/api-state')` or `import ___ from './src/api-state'` - importing from deeper modules is discouraged and undocumented, although still technically possible because JS.

This `src/api-state/index.js` module will eventually be published to npm as `mwp-api-state`, so imports will be simple/standard.

The _major_ exception to this pattern is when you want to export a bunch of things that are designed to be used independently, like MWC. Using an `index.js` public interface means that each time you import one thing, you have to bundle _everything_ - this can be particularly problematic when some of the exports are designed to be used exclusively on the server, which often has very large third-party dependencies.

For something like `api-state`, that's not a problem because it's designed to be used all-or-nothing, server side and browser side.

src/api-state/reducer.js
mmcgahan: since we're batching requests to the API, we need a way of distinguishing a failed batch request from individual failed API requests - i.e. if the Node API proxy throws an error, it's a 'failure' that needs to be handled at the root application level, whereas REST API response errors should generally be dealt with by the particular requesting container.

This is something that I expect we'll need to build a generic toast/UI for sometime soon - it essentially means the app is broken and probably can't recover (although it may just be a temporary network issue)

src/router/routeActionCreators.js
mmcgahan: Both [Redux](https://github.com/reactjs/redux/blob/c9ad6a40fe501919d359b910ea7c7f34532e11d9/examples/todos-flow/src/types/index.js#L11) and [react-router-redux](https://github.com/reactjs/react-router-redux/blob/a304bda820ba57f68961035eff1950e618047b81/src/reducer.js#L5) use this convention for actions that are triggered by middleware rather than the UI, so I've mirrored it here - specifically `@@router/LOCATION_CHANGE` is the same as the action from react-router-redux, and the purpose of it is identical here.

tests/mockApp.jsx
```diff
@@ -36,7 +36,10 @@ const MockRedirect = props => {
 export const routes = [
 	{
 		path: '/',
-		component: 'div',
+		component: props =>
```
mmcgahan: these updates were just to prevent an irrelevant warning from being thrown in integration testing

src/api-state/sync/index.js
```diff
@@ -8,18 +8,40 @@ import 'rxjs/add/operator/catch';
...
-} from '../actions/syncActionCreators';
-import { actions as clickActions } from '../plugins/tracking/util/clickState';
-import { getRouteResolver, getMatchedQueries } from '../router/util';
-import { getDeprecatedSuccessPayload } from '../util/fetchUtils';
+} from './syncActionCreators';
+
+/**
+ * @deprecated
+ */
+export function getDeprecatedSuccessPayload(successes, errors) {
```
mmcgahan: I moved this function over from `fetchUtils` because it's only used here and `fetchUtils` doesn't need to be a dependency

mmcgahan:
good to go

PR #344, opened 2017-08-12
WP 442 platform store modules

mmcgahan:
This is the second part of the effort to modularize the Redux-related parts of the platform. I found that it made sense to separate the 'store' modules from the 'state' modules, and just set up 'state' as a dependency of 'store' so that the store could incorporate the api middleware from 'state' - there is no circular dependency.

The 'store' module is actually a package with two public modules - 'store/server' and 'store/browser' - because the dependencies for the browser module are not as heavy as the server-side dependencies (e.g. rxjs).

Structurally, this refactor moves the existing `fetchQueries` function into `store/browser`, which is now the only place that it is used. A corresponding `fetchQueries` function for `store/server` is also defined - the two store implementations are as similar as possible so to make the key differences (hopefully) more obvious.


src/middleware/catch.js
```diff
@@ -1,15 +0,0 @@
-/**
```
mmcgahan: moved to `src/store/middleware/` + converted to Flow type

src/plugins/requestAuthPlugin.js
```diff
@@ -21,6 +20,26 @@ import {
...
  */
 
+/**
+ * Attempt to JSON parse a Response object from a fetch call
+ *
+ * @param {String} reqUrl the URL that was requested
+ * @param {Response} response the fetch Response object
+ * @return {Promise} a Promise that resolves with the JSON-parsed text
+ */
+export const tryJSON = reqUrl => response => {
```
mmcgahan: moved from `src/util/fetchUtils/` - it's only used here.

src/plugins/requestAuthPlugin.test.js
```diff
@@ -351,3 +352,42 @@ describe('getAuthenticate', () => {
 		);
 	});
 });
+
+describe('tryJSON', () => {
```
mmcgahan: move only, no new tests

src/store/browser/fetchQueries.js
```diff
@@ -1,40 +1,12 @@
 import JSCookie from 'js-cookie';
```
mmcgahan: As mentioned in the description, there are now two `fetchQueries` modules - one for browser, one for server. This one is marked as a 'move' because it is most similar to the old `fetchUtils`  module.

src/store/middleware/injectPromise.test.js
```diff
@@ -0,0 +1,38 @@
+import { requestAll } from '../../api-state'; // mwp-api-state
+import injectPromise from './injectPromise';
+
+describe('injectPromise middleware', () => {
```
mmcgahan: new tests for `injectPromise` middleware

src/store/server/fetchQueries.js
```diff
@@ -0,0 +1,18 @@
+import 'rxjs/add/operator/map';
```
mmcgahan: this is a new module that just encapsulates the server-side equivalent of `fetchQueries` - the code was moved out of `src/store/server/index.js` just to keep the structure parallel with `src/store/browser/index.js`.

src/renderers/server-render.jsx
```diff
@@ -9,11 +9,10 @@ import ReactDOMServer from 'react-dom/server';
 import Helmet from 'react-helmet';
 import { API_ROUTE_PATH } from '../plugins/api-proxy'; // mwp-api-proxy
 import { Forbidden, NotFound, Redirect, SERVER_RENDER } from '../router'; // mwp-router
+import { getServerCreateStore } from '../store/server'; // mwp-store
+import Dom from '../render/components/Dom'; // mmwp-render/components/Dom
```
willh-meetup: `mmwp`

Do these comments have any programmatic significance now, or just planning for where we want them in the future?

src/store/middleware/injectPromise.test.js
willh-meetup: 👍

src/store/server/fetchQueries.js
```diff
@@ -0,0 +1,17 @@
...
+ * request to the server's own API proxy endpoint
+ */
+export default (request: HapiRequest) => () => (
+	queries: Array<Query>
+): Promise<ParsedQueryResponses> =>
+	request
+		.proxyApi$(queries)
+		.map(responses => ({ responses })) // package the responses in object like the API proxy endpoint does
+		.toPromise()
+		.then(parseQueryResponse(queries));
```
willh-meetup: if `parseQueryResponse` throws an error, does it get caught somewhere? Or do we not care about that here?

src/store/README.md
```diff
@@ -0,0 +1,15 @@
+# Store
+
+The Redux `createStore` helpers for MWP applications - `store/browser` and
+`store/server`. Both modules are implemented as 'getters' with a similar
+signature - the main difference is that the `store/server` function takes an
+additional `request` parameter corresponding to a Hapi request.
+
+## Dependencies
```
willh-meetup: I like that we're talking about dependencies here!

src/store/util/fetchUtils.js
```diff
@@ -0,0 +1,27 @@
+// @flow
+
+/**
+ * A module for middleware that would like to make external calls through `fetch`
+ */
+export const parseQueryResponse = (queries: Array<Query>) => (
```
willh-meetup: Is it worth writing more tests for this, or do we feel like we've got pretty good coverage with the fetchQueries tests?

src/renderers/server-render.jsx
mmcgahan: just planning, and making some of these lines more grep-able when we do make the conversion

src/store/util/fetchUtils.js
mmcgahan: I'll add a couple tests and then merge

PR #347, opened 2017-08-15
update regex (copied from url-regex, modified to not select closing p…

jaclynj:
…arentheses); remove url-regex package

### DIFF
--- a/src/util/linkify.js
+++ b/src/util/linkify.js
@@ -1,5 +1,7 @@
 // @flow
-import urlRegex from 'url-regex';
+var urlRegex = new RegExp(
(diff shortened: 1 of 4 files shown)


src/util/linkify.js
```diff
@@ -1,5 +1,7 @@
 // @flow
-import urlRegex from 'url-regex';
+var urlRegex = new RegExp(
+	/(?:(?:(?:[a-z]+:)?\/\/)|www\.)(?:\S+(?::\S*)?@)?(?:localhost|(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9][0-9]|[0-9])(?:\.(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9][0-9]|[0-9])){3}|(?:(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)(?:\.(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)*(?:\.(?:[a-z\u00a1-\uffff]{2,})))(?::\d{2,5})?(?:[/?#][^\s)"]*)?/gi
+);
```
mmcgahan: probably would be good to document where this came from

src/util/linkify.test.js
```diff
@@ -43,4 +43,13 @@ describe('linkify', () => {
...
 		expect(linkify(plainBase)).toBe(expectedLink);
 	});
+	it('should select and create a link around only the url', () => {
+		const plainText =
+			'Check out my Meetup (http://www.meetup.com/hq-faff), it is awesome.';
+		const expectedHTML =
+			'Check out my Meetup (' +
+			'<a class="link" href="http://www.meetup.com/hq-faff" title="http://www.meetup.com/hq-faff" target="" >' +
+			'http://www.meetup.com/hq-faff</a>), it is awesome.';
+		expect(linkify(plainText)).toBe(expectedHTML);
```
mmcgahan: you could technically use a snapshot for this so you don't have to manually write the result string, but no biggie

PR #349, opened 2017-08-15
error path implemented. should throw

mattkime:


src/routes/index.js
```diff
@@ -23,7 +23,13 @@ export default function getRoutes(renderRequestMap, apiProxyFn$ = apiProxy$) {
 		method: 'GET',
 		path: '/_ah/{param*}',
 		config: { auth: false },
-		handler: (request, reply) => reply('OK'),
+		handler: (request, reply) => {
+			if (request.params.checkType === 'error') {
```
mmcgahan: I think this would be `request.params.param === 'error'` if you want to test `/_ah/error` or `request.query.checkType === 'error'` if you want to test `/_ah/?checkType=error`

mattkime: yes, def. this needs another look

mattkime: should be better now. any advice on testing this?

mmcgahan: yeah, you should be able to make a direct copy of [this integration test](https://github.com/meetup/meetup-web-platform/blob/master/tests/integration/render.int.js#L211-L228) and just change the request `url` to the expected-to-fail url. Probably best to put the integration test into a new `tests/integration/gaeRoute.int.js` script, but adding it to `render.int.js` is fine.

src/routes/index.js
```diff
@@ -23,7 +23,13 @@ export default function getRoutes(renderRequestMap, apiProxyFn$ = apiProxy$) {
 		method: 'GET',
 		path: '/_ah/{param*}',
 		config: { auth: false },
-		handler: (request, reply) => reply('OK'),
+		handler: (request, reply) => {
+			if (request.params.param == !'error') {
```
mmcgahan: looks like a typo here - I think it should be `request.params.param !== 'error'`

src/routes/index.js
```diff
@@ -23,7 +23,13 @@ export default function getRoutes(renderRequestMap, apiProxyFn$ = apiProxy$) {
...
+				reply('OK');
+			} else {
+				throw new Error('Simulated error via url');
```
mmcgahan: A helpful convention is to handle errors at the top of your function and then keep the 'happy path' left-aligned - it reduces nesting and makes the error conditions a little easier to identify

```js
if (request.params.param == 'error') {
  throw new Error();
}
reply('OK');
```

PR #353, opened 2017-08-18
Wp 463 stop enforcing dev api

mattkime:


willh-meetup:
There are a few considerations with this:

1) If someone forgets they are hooked up to the LIVE api in dev, they could do some real damage to prod, so I think we'd want to have a visible warning somewhere about this.
2) If they aren't logged in to prod, but are on dev, cookies and such will not work as expected.

mattkime:
I think we can probably close this until there's more discussion

willh-meetup:
I'm gonna close, since I know we need to talk about exactly what needs to happen here.

PR #358, opened 2017-08-28
WP-468 Add support for browser ID

mmcgahan:
This PR probably looks much more complicated than necessary to add browser ID tracking, BUT I went ahead and did a significant refactor because there was some legacy 'session id' behavior that I had left in place from an early iteration of the code that is no longer necessary, and made the 'browser id' behavior more difficult to define.

### DIFF
--- a/src/plugins/app-route/handler.js
+++ b/src/plugins/app-route/handler.js
@@ -30,7 +30,6 @@ export default (languageRenderers: { [string]: LanguageRenderer$ }) => (
 					.redirect(renderResult.redirect.url)
 					.permanent(Boolean(renderResult.redirect.permanent));
 			}
-			request.trackSession();
 			return reply(renderResult.result).code(renderResult.statusCode);
 		},
 		err => reply(err) // 500 error - will only be thrown on bad implementation
--- a/src/plugins/tracking/_activityTrackers.js
+++ b/src/plugins/tracking/_activityTrackers.js
@@ -1,8 +1,7 @@
 // @flow
 import url from 'url';
 import rison from 'rison';
-import { parseMemberCookie } from '../../util/cookieUtils';
-import { updateTrackId, newSessionId } from './util/idUtils';
+import { parseIdCookie, updateId } from './util/idUtils';
 
 const parseUrl = url.parse;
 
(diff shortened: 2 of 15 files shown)


src/plugins/app-route/handler.js
```diff
@@ -30,7 +30,6 @@ export default (languageRenderers: { [string]: LanguageRenderer$ }) => (
 					.redirect(renderResult.redirect.url)
 					.permanent(Boolean(renderResult.redirect.permanent));
 			}
-			request.trackSession();
```
mmcgahan: no longer making a separate tracking call on first render - tracking is handled entirely by `trackApi`, which is called on first render and subsequent API proxy calls

src/plugins/tracking/activity.js
```diff
@@ -84,23 +83,23 @@ const onRequest = (request, reply) => {
...
+	const { browserIdCookieName, trackIdCookieName } = cookieConfig;
+	const pluginData = request.plugins[ACTIVITY_PLUGIN_NAME];
+	const browserId = pluginData.browserIdCookieName;
+	const trackId = pluginData.trackIdCookieName;
+
+	const FOREVER: CookieOpts = {
 		encoding: 'none',
 		path: '/',
 		isHttpOnly: true,
-		isSecure: isProd,
```
mmcgahan: `isSecure` is the default, and is now supported dev and prod, so we can leave it out

src/plugins/tracking/activity.js
```diff
@@ -117,13 +116,21 @@ export default function register(
 ) {
 	const { agent, isProd } = options;
 
-	const sessionIdCookieName: string = isProd ? 'SESSION_ID' : 'SESSION_ID_DEV';
-	const trackIdCookieName: string = isProd ? 'TRACK_ID' : 'TRACK_ID_DEV';
+	const trackIdCookieName: string = isProd
+		? 'MEETUP_TRACK'
```
mmcgahan: Note that we weren't actually using chapstick's `MEETUP_TRACK` cookie until now - we now consume it, and only create a new one if it doesn't exist in the initial request.

src/plugins/tracking/_activityTrackers.js
```diff
@@ -12,68 +11,53 @@ const parseUrl = url.parse;
...
-		sessionId: newSessionId(request),
-		url: request.url.path,
-	});
-};
-
 export const getTrackApiResponses: TrackGetter = trackOpts => request => (
 	queryResponses: Array<Object>,
-	url: string,
-	referrer: string
+	url: ?string,
```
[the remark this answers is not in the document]
mmcgahan: yeah, I think there were cases where these were `undefined` - there's an existing empty-string fallback for each - so I wanted to just make sure the flowtype was accurate. I didn't actually do much more digging to verify how it's being called, but the function works fine with these args undefined.

PR #364, opened 2017-09-15
disable slack notifs

mattkime:
disable slack notifs


mattkime:
I'm just going to go ahead and merge this. can be reverted if someone disagrees with this change.

PR #366, opened 2017-09-19
Suppress 400 errors

willh-meetup:
Not logging an error for 400 errors right now. They all seem related to auth, and they don't seem to break anything (other than alert us), and we are gonna be making a lot of these happen soon.

### DIFF
--- a/src/plugins/api-proxy/util/receive.js
+++ b/src/plugins/api-proxy/util/receive.js
@@ -196,14 +196,11 @@ export const makeLogResponse = request => ([response, body]) => {
 		statusCode >= 500 || // REST API had an internal error
 		(method.toLowerCase() === 'get' && statusCode >= 400) // something fishy with a GET
 	) {
-		// use console.error to highlight these cases in Stackdriver
-		if (
-			statusCode === 400 &&
-			(pathname === '/members/self' || pathname === '/members/home')
-		) {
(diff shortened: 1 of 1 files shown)


src/plugins/api-proxy/util/receive.js
```diff
@@ -196,6 +196,10 @@ export const makeLogResponse = request => ([response, body]) => {
 		statusCode >= 500 || // REST API had an internal error
 		(method.toLowerCase() === 'get' && statusCode >= 400) // something fishy with a GET
 	) {
+		if (statusCode === 400) {
```
mmcgahan: should we also qualify this by checking the `pathname` so that we only swallow `/member/self` errors? I can imagine this auto-swallowing-400 hiding a 'real' error down the road.

willh-meetup: @mmcgahan That's what I did at first, but it turns out that there are a number of requests that 400 (that we don't need to load if you aren't logged in).

Agreed that I would prefer to not swallow them, but until we fix the underlying issues, I think it's better to treat 400 specifically as a non-error.

willh-meetup: I put a thing on our agenda tomorrow so I can explain the whole issue and so we can make a better fix.

PR #367, opened 2017-09-19
WP-487: Convert detected timezone to NY timezone

eilinora:
converts the detected timezone to ny for logging

### DIFF
--- a/src/plugins/tracking/activity.js
+++ b/src/plugins/tracking/activity.js
@@ -17,9 +17,18 @@ export const getLogger: string => (Object, Object) => mixed = (
 	agent: string
 ) => (request: Object, trackInfo: Object) => {
 	const requestHeaders = request.headers;
-	const now: Date = new Date();
+	
+	// Takes in desired time to convert and applies offset
+	const offset: String = new Date().getTimezoneOffset() * 60000; // gets detected timezone + converts to milliseconds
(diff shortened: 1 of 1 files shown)


eilinora:
@willh-meetup validated this is what Tom and team are looking for

mmcgahan:
I think I'd like to file a followup ticket to implement this using [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat), which would be more performant and avoid manually managing offsets and instantiating more than one `Date` object.

src/plugins/tracking/activity.js
```diff
@@ -17,9 +17,18 @@ export const getLogger: string => (Object, Object) => mixed = (
 	agent: string
 ) => (request: Object, trackInfo: Object) => {
 	const requestHeaders = request.headers;
-	const now: Date = new Date();
+	
+	// Takes in desired time to convert and applies offset
+	const offset: String = new Date().getTimezoneOffset() * 60000; // gets detected timezone + converts to milliseconds
```
mmcgahan: Looks like our PR builder isn't catching Flow errors - this and `NY_OFFSET` are not strings. You also shouldn't have to declare types on variables with types that can be easily inferred (e.g. this Number and the Date vars).

PR #370, opened 2017-09-22
Don't log 4XX errors as error-level in our logs for right now

willh-meetup:


src/plugins/api-proxy/util/receive.js
```diff
@@ -196,31 +200,32 @@ export const makeLogResponse = request => ([response, body]) => {
...
+				id,
+				statusCode,
+				time: elapsedTime,
+				originRequestId: request.id,
+				body,
+			},
+		});
+
+		// Only want to warn about 400s for now (maybe forever?)
+		if (statusCode < 500) {
```
willh-meetup: Truthfully, I'm not sure if new relic picks these up, but I think better to err on the side of caution.

src/plugins/api-proxy/util/receive.js
```diff
@@ -122,9 +122,13 @@ export const parseApiValue = ([response, body]) => {
...
 	try {
 		const value = parseBody(body);
-		if (response.statusCode < 200 || response.statusCode > 399) {
+		if (response.statusCode < 200 || response.statusCode >= 500) {
 			return formatApiError(new Error(response.statusMessage), value);
 		}
+		if (response.statusCode >= 400) {
+			// Don't throw an error for 400s for now (maybe forever?)
+			return formatApiError({ message: response.statusMessage }, value);
+		}
```
mmcgahan: this has the same effect of the previous implementation - `formatApiError` expects an `Error` instance and will take the `message` prop from it without `throw`ing.

src/plugins/api-proxy/util/receive.js
```diff
@@ -196,31 +200,32 @@ export const makeLogResponse = request => ([response, body]) => {
...
-				},
-			})
-		);
 	}
 	const logger = request.server.app.logger;
 	// production logs will automatically be JSON-parsed in Stackdriver
-	const log = ((statusCode >= 400 && logger.error) ||
-		(statusCode >= 300 && logger.warn) ||
+	const log = ((statusCode >= 500 && logger.error) ||
+		(statusCode >= 400 && logger.warn) ||
```
willh-meetup: I'm betting on this *definitely* making a difference.

src/plugins/api-proxy/util/receive.js
```diff
@@ -196,31 +200,32 @@ export const makeLogResponse = request => ([response, body]) => {
...
+				body,
+			},
+		});
+
+		// Only want to warn about 400s for now (maybe forever?)
+		if (statusCode < 500) {
+			console.warn(errorMessageString);
+		} else {
+			console.error(errorMessageString);
 		}
```
mmcgahan: trivially DRYer alternative

```js
const logError = statusCode < 500 ? console.warn : console.error;
logError(errorMessageString);
```

src/plugins/api-proxy/util/receive.js
mmcgahan: yeah i'm guessing it watches stderr

src/plugins/api-proxy/util/receive.js
willh-meetup: Yeah, I did this first and it I don't think it's helpful. Removed.

src/plugins/api-proxy/util/receive.js
willh-meetup: 👍

willh-meetup:
OK, comments addressed.

PR #376, opened 2017-09-28
support for _set_geoip querystring param

willh-meetup:
I think this is the simplest way to do this.

Prelim so I can test in mup-web.

### DIFF
--- a/packages/mwp-api-proxy-plugin/src/util/send.js
+++ b/packages/mwp-api-proxy-plugin/src/util/send.js
@@ -173,7 +173,8 @@ export function getLanguageHeader(request) {
 }
 
 export function getClientIpHeader(request) {
-	const clientIP = request.headers['fastly-client-ip'];
+	const clientIP =
+		request.params['_set_geoip'] || request.headers['fastly-client-ip'] || null;
 	if (clientIP) {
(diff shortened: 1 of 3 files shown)


packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -173,7 +173,8 @@ export function getLanguageHeader(request) {
 }
 
 export function getClientIpHeader(request) {
-	const clientIP = request.headers['fastly-client-ip'];
+	const clientIP =
+		request.params['_set_geoip'] || request.headers['fastly-client-ip'] || null;
```
mmcgahan: `request.params` refers to named wildcard matches in the `path`. `request.query` corresponds to parsed querystring params

https://hapijs.com/api#request-properties

mmcgahan:
Sorry about not having the integration tests working - that would be the 'right' way to test this using a 'real' (simulated) Hapi request...

packages/mwp-api-proxy-plugin/src/util/send.js
willh-meetup: ooh, yep, I totally misread the docs.

willh-meetup:
@mmcgahan I think this is good to go.

PR #378, opened 2017-09-29
logging improvements across the board

mmcgahan:
This addresses a number of logging updates/improvements:

1. https://meetup.atlassian.net/browse/WP-518
   - adding an `httpRequest` property to the log so that response logs will highlight
      - request method
      - statusCode
      - agent
      - payload size

screen shot 2017-09-29 at 2 34 22 pm

2. https://meetup.atlassian.net/browse/WP-525
   - don't log requests to `/ping`
      - can be viewed in the `nginx.request` log
   - don't log requests to `/_ah/health`
      - can be viewed in the `nginx.health_check` log

3. remove the 'incoming request' log
   - this is mainly noise because we are much more interested in response logs

4. Make sure errors are logged correctly using an `err` key, along with the corresponding `req`/`res` info when available. This ensures that errors are captured correctly by the GAE Error Reporting tool and are clearly visible in the regular log viewer

5. Don't log status code 3xx responses with a 'warning' - just 'info'.

6. Integrate a pretty-print stream directly into the logger plugin rather than requiring output to be piped through the `bunyan` CLI in consumer apps (this fixes a bug that prevents testing multiple languages in dev in mup-web currently)

### DIFF
--- a/packages/mwp-app-server/src/routes/index.js
+++ b/packages/mwp-app-server/src/routes/index.js
@@ -3,7 +3,14 @@ export default function getRoutes() {
 		path: '/ping',
 		method: 'GET',
 		handler: (request, reply) => reply('pong!'),
-		config: { auth: false },
+		config: {
+			auth: false,
+			plugins: {
+				'mwp-logger-plugin': {
+					enabled: false,
+				},
+			},
+		},
 	};
 
 	// simple 200 response for all lifecycle requests
@@ -11,7 +18,14 @@ export default function getRoutes() {
 	const appEngineLifecycleRoutes = {
 		method: 'GET',
 		path: '/_ah/{param*}',
-		config: { auth: false },
+		config: {
+			auth: false,
+			plugins: {
+				'mwp-logger-plugin': {
+					enabled: false,
+				},
+			},
+		},
 		handler: (request, reply) => {
 			if (request.params.param === 'error') {
 				throw new Error('Simulated error via url');
--- a/packages/mwp-app-server/src/util/index.js
+++ b/packages/mwp-app-server/src/util/index.js
@@ -23,51 +23,8 @@ export function checkForDevUrl(value) {
 }
 
 export function onRequestExtension(request, reply) {
-	request.id = uuid.v4();
-
-	request.server.app.logger.info(
-		`Incoming request ${request.method.toUpperCase()} ${request.url.href}`
-	);
-
+	request.id = uuid.v4(); // provide uuid for request instead of default Hapi id
 	return reply.continue();
-}
-
-export function logResponse(request) {
-	const {
-		method,
-		response,
-		id,
-		info,
-		server: { app: { logger } },
-		url,
-	} = request;
-
-	if (response.isBoom) {
-		// response is an Error object
-		logger.error(`Internal error ${response.message} ${url.pathname}`, {
-			error: response.stack,
-		});
-	}
-
-	const log = ((response.statusCode >= 400 && logger.error) ||
-		(response.statusCode >= 300 && logger.warn) ||
-		logger.info)
-		.bind(logger);
-
-	log(
-		{
-			headers: response.headers,
-			id,
-			method,
-			referrer: info.referrer,
-			remoteAddress: info.remoteAddress,
-			elapsedTime: info.responded - info.received,
-			href: url.href,
-		},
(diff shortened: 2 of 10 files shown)


packages/mwp-app-server/src/util/index.js
```diff
@@ -23,53 +23,10 @@ export function checkForDevUrl(value) {
...
-
-	request.server.app.logger.info(
-		`Incoming request ${request.method.toUpperCase()} ${request.url.href}`
-	);
-
+	request.id = uuid.v4(); // provide uuid for request instead of default Hapi id
 	return reply.continue();
 }
 
-export function logResponse(request) {
```
mmcgahan: moved to the logger plugin, where it makes more sense

packages/mwp-logger-plugin/src/index.js
```diff
@@ -1,19 +1,56 @@
...
+			err: response,
 			...request.raw,
-		},
-		`500 Internal server error: ${err.message}`
-	);
+		});
+	}
+
+	const log = ((response.statusCode >= 500 && logger.error) ||
+		(response.statusCode >= 400 && logger.warn) ||
```
mmcgahan: should 4xx errors be treated as 'errors' in the logs, or just warning? Actual thrown errors will result in 5xx responses.

packages/mwp-logger-plugin/src/index.js
```diff
@@ -1,19 +1,56 @@
...
+	const routePluginSettings =
+		request.route.settings.plugins['mwp-logger-plugin'];
+	if (!routePluginSettings || routePluginSettings.enabled) {
+		log({ httpRequest: request, id });
+	}
+
+	return;
+}
+
+const onRequestError = (request, err) => {
```
mmcgahan: this _might_ result in some duplicate logs, but it was something we had before so I don't want to remove it yet

packages/mwp-logger-plugin/src/index.js
mattkime: 404s strike me as a special class of errors. in small numbers we really don't care. if a particular 404 is served in high quantity then something is likely wrong.

i guess i'm concerned that logging them as 'warm' might hide problems.

mattkime:
lgtm aside from open question about how to treat 404s

packages/mwp-app-server/src/routes/index.js
```diff
@@ -3,15 +3,29 @@ export default function getRoutes() {
 		path: '/ping',
 		method: 'GET',
 		handler: (request, reply) => reply('pong!'),
-		config: { auth: false },
+		config: {
+			auth: false,
+			plugins: {
+				'mwp-logger-plugin': {
+					enabled: false,
```
willh-meetup: Nice.

packages/mwp-logger-plugin/src/index.js
willh-meetup: Yeah, I think warnings are better. This is also a thing that we can instrument in stackdriver to provide visibility.

packages/mwp-logger-plugin/src/logger.js
```diff
@@ -1,12 +1,55 @@
 import bunyan from 'bunyan';
+import bunyanDebugStream from 'bunyan-debug-stream';
 import LoggingBunyan from '@google-cloud/logging-bunyan';
 
-const streams = [{ stream: process.stdout }]; // stdout by default
-const { GAE_INSTANCE, GCLOUD_PROJECT, GAE_VERSION, GAE_SERVICE } = process.env;
+const formatDuration = ms => {
+	const seconds = Math.floor(ms / 1000);
+	const nanos = ms % 1000 * 1000 * 1000;
```
willh-meetup: I don't really understand what this is for, but I assume we copied it from somewhere and it is what we want?

mmcgahan: yeah, I'll add comments. It's converting the millisecond response time that we get from Hapi into a `{ seconds, nanos }` object that is required by the 'latency' logger in [Stackdriver's HttpRequest](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#httprequest) param

Specifically, it is a [Duration](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration), which seems like an amazingly awkward definition of time span.

PR #379, opened 2017-09-29
remove 'send' logging, convert 'receive' logging to use the logger

mmcgahan:
This diff looks more complicated than it is. The main change is to stop logging 'send' events for requests to the API server, and improve the logging of 'receive' events from the API server - specifically, we no longer need to use `console.error`/`console.warn`, and we can format the logs so that they are more consistent with other logs and provide more contextual information WRT to the parent request and the API response properties

### DIFF
--- a/packages/mwp-tracking-plugin/src/util/avro.js
+++ b/packages/mwp-tracking-plugin/src/util/avro.js
@@ -23,14 +23,9 @@ const getPlatformAnalyticsLog = (
 		});
 		const publisher = pubsub.topic('analytics-log-json').publisher();
 		return (serializedRecord: string) => {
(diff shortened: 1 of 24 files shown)


packages/mwp-tracking-plugin/src/util/avro.js
```diff
@@ -23,14 +23,9 @@ const getPlatformAnalyticsLog = (
 		});
 		const publisher = pubsub.topic('analytics-log-json').publisher();
 		return (serializedRecord: string) => {
-			publisher.publish(new Buffer(serializedRecord)).then(
```
mmcgahan: also dropping this log, which wasn't actually very useful - just was a 'success' when an analytics log was sent. better to just log errors.

PR #381, opened 2017-10-04
Wp 503 toastmaster

mmcgahan:
**`Prelim`** - I should probably add some testing, and maybe more docs to this, but I figure the main CR can start.


sadafie:
Super dumb question -- what are toasts? or what is a toastmaster in this context? I probably missed that discussion at some point.

mattkime:
@sadafie they're bar across the screen that appears just under the header nav. they have a dismiss button. past that, i'm faiiirly certain they're just a container for other content. mostly designed for user messaging.

packages/mwp-toaster/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/meetup/meetup-web-platform/issues"
+  },
+  "homepage": "https://github.com/meetup/meetup-web-platform#readme",
+  "dependencies": {},
+  "devDependencies": {
+    "babel-jest": "21.0.0",
+    "jest": "20.0.4",
+    "meetup-web-components": "^3.0.1281",
```
mattkime: when should modules be added to meetup-web-components vs being their own module? 

do we get the benefits of an independent module if meetup-web-components is a peer dependency? is there a risk of updating meetup-web-components in such a way that it passes all its tests yet breaks module?

packages/mwp-toaster/src/ToastContainer.jsx
```diff
@@ -0,0 +1,70 @@
...
+ * 1. Creating toasts should _always_ be a Redux action, e.g. TOAST_MAKE
+ * 2. TOAST_MAKE will go through the platform reducer to populate an array in {{state.toast.ready}}
+ * 3. When the ToastMaster is instantiated, it will consume {{state.toast.ready}} and dispatch a new action TOAST_SHOW
+ * 4. TOAST_SHOW will _clear_ {{state.toast.ready}}
+ */
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import withRouter from 'react-router-dom/withRouter';
+import Toaster from 'meetup-web-components/lib/Toaster';
```
mattkime: Could all the Toast* related stuff go into the same repo?

packages/mwp-toaster/package.json
mmcgahan: components in `meetup-web-components` are intended to be dumb display units that can be composed by consumers, but the `mwp-toaster` package ties the Toast components into redux state, which we don't want to set as a `meetup-web-components` dependency. I talked with MikeP, who wrote the `<Toast>` and `<Toaster>` components, about moving at least `<Toaster>` into `mwp-toaster` because its primary purpose is to be a 'controller' of dumb toasts, so it kind of straddles the line between what meetup-web-components is responsible for and what a consumer application needs - it defines display timing logic and lays out the toasts, but stops short of depending on overall application state.

I'd like to make that move in a followup ticket, however, because it involves some refactoring and making more fine-grained decisions about what goes where. As you recommended, I think I would _like_ all the Toast* related stuff to go into this repo, but that would make it more awkward to display it in storybook with the rest of the `meetup-web-components`, so I'm not sure what the best call is in the short term.

Long term, I'm expecting that `meetup-web-components` can be moved into `meetup-web-platform` as a `mwp-components` package or something, although that will take a fair amount of planning as well.

packages/mwp-toaster/src/ToastContainer.jsx
```diff
@@ -0,0 +1,70 @@
...
+import { getReadyToasts } from './selectors';
+
+const mapStateToProps = state => ({ ready: getReadyToasts(state) });
+const mapDispatchToProps = { makeToast, showToasts };
+
+/*
+ * This component does *not* support server-rendered toasts. It must mount before
+ * toasts are added to application state through `makeToast`.
+ */
+export class ToastContainer extends React.Component {
```
eilinora: shouldn't this be PureComponent if you are using `shouldComponentUpdate`

packages/mwp-toaster/src/ToastContainer.jsx
```diff
@@ -0,0 +1,70 @@
...
+	// Custom 'shouldComponentUpdate' in order to avoid re-rendering when there are
+	// no new toasts
+	shouldComponentUpdate(nextProps) {
+		return nextProps.ready.length > 0 && nextProps.ready !== this.props.ready;
```
eilinora: If your not storing new toasts in local state, wouldn't this clear rendered toasts as `ready` updated?

packages/mwp-toaster/src/ToastContainer.jsx
mmcgahan: `PureComponent` is basically a shortcut method of implementing a [simple `shouldComponentUpdate` automatically](https://medium.com/@kylpo/all-about-reacts-shouldcomponentupdate-cc3b1e497e97#ad24) - I want to manually specify it for this component. Basically, if you are specifying your own `shouldComponentUpdate`, you probably should _not_ use `PureComponent`

packages/mwp-toaster/src/ToastContainer.jsx
mmcgahan: the `<Toaster>` stores the toasts in its local state

packages/mwp-toaster/package.json
mattkime: >>Long term, I'm expecting that meetup-web-components can be moved into meetup-web-platform as a mwp-components package or something

I think I'm still unclear of the benefit of this. Can you explain?

mmcgahan: - fewer repos to keep track of/configure - this is the big win for simplicity because we only need one travis configuration for publishing any package, and things like eslint, flow, and storybook config can be centralized
- simpler dependency management
   - peer deps: all MWP packages increment together and are tested together, avoiding the risk that you highlighted about updating meetup-web-components in such a way that it passes all its unit tests yet breaks integration with this package
   - common dev dependencies - they can all be consolidated at the root of the repo so we don't have update them in many different places

eilinora:
looking good than, just needs tests :)

eilinora:
@mmcgahan following up on this PR, what is the expected ETA for this to be merge?

mmcgahan:
I need to fix up the build error, but then i think it's ready for final review. We'll need to do some beta testing with it in mup-web, but I think the principle is relatively simple so hopefully it will behave as expected

mattkime:
I might be staring right at it, but I'm unclear about how this will be used.

mmcgahan:
@mattkime essentially it's a way of triggering [this UI](https://meetup.github.io/meetup-web-components/?selectedKind=Toast&selectedStory=multiple) using Redux actions in a way that re-uses a single 'controller' component to orchestrate show/hide. Currently, the only way to render a toast is to add them manually to each container that might want to use them, and it could get messy.

The added feature is that the controller component will also look for querystring params in the initial URL and activate some corresponding pre-defined 'system message' toasts - this will be used to support handoff from chapstick to the web platform in a way that keeps important messages visible.

PR #382, opened 2017-10-06
force ints to be ints

mmcgahan:
fixes https://console.cloud.google.com/errors/3479501470273157020?service=default&time=PT6H&filter&project=meetup-prod-east4


mmcgahan:
SV

PR #385, opened 2017-10-12
Location changes generate ref whitelist to suppress flash of empty content

mmcgahan:
When moving around the app, we can prevent some flashes of 'loading' content by maintaining 'parent route' state when navigation occurs.

Currently, _every_ nav action clears the `ref`s in `state.api` that are expected to be returned by the API. We get 'fast loading' behavior by relying on _cached_ values to immediately replace the cleared data. However, the delay between clearing the data and replacing it with cached data is long enough to be noticeable, and _can_ result in some subtle state errors (e.g. the `self` value gets cleared, which means it's not possible to tell whether a user is logged in or not during this split-second re-render).

This PR adds another metadata option to API request actions, allowing the request to carry information about which `ref`s should be _retained_ in state when the request is made. This will prevent the flash of empty content, although it may also allow stale data to remain in `state` longer than it should if 'very outdated' data is marked in `retainRefs`

### DIFF
--- a/packages/mwp-api-state/src/sync/index.js
+++ b/packages/mwp-api-state/src/sync/index.js
@@ -98,9 +98,6 @@ export const getNavEpic = (routes, baseUrl) => {
 						return newQueries;
 					})
 				).map(q => api.requestAll(q, requestMetadata));
-				// if requestAll included instructions about which refs to retain,
-				// the reducer could ignore those refs when it would otherwise
-				// clear them in the platform reducer line 49
 
 				const clickAction$ = Observable.of(clickActions.clear());
 
--- a/packages/mwp-api-state/src/sync/index.js
+++ b/packages/mwp-api-state/src/sync/index.js
@@ -8,7 +8,7 @@ import 'rxjs/add/operator/catch';
 import 'rxjs/add/operator/map';
 import 'rxjs/add/operator/mergeMap';
 import 'rxjs/add/operator/takeUntil';
-import { ActionsObservable, combineEpics } from 'redux-observable';
+import { combineEpics } from 'redux-observable';
 
 import { LOCATION_CHANGE, SERVER_RENDER } from 'mwp-router';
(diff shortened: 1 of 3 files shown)


packages/mwp-api-state/src/sync/index.js
```diff
@@ -74,19 +74,29 @@ export const getNavEpic = (routes, baseUrl) => {
...
+					retainRefs: [],
 				};
 
 				const cacheAction$ = requestMetadata.logout
 					? Observable.of({ type: 'CACHE_CLEAR' })
 					: Observable.empty();
 
 				const apiAction$ = Observable.fromPromise(
-					resolveRoutes(location, baseUrl).then(getMatchedQueries(location))
+					Promise.all([
```
eilinora: So if i'm reading this right, if i navigate 
group home w/ `retainref` for group query <- makes query request
navigate to child route group discussion page <- no new request
go back group home <- makes query request

right? :)

mmcgahan: code-wise all of this should be completely invisible to consumer apps - it will work automagically without you needing to specify anything in your queries

specifically, this routine programmatically checks what refs were populated by the queries on the previous page, diffs them with the refs that will be populated by the current page, and _only_ clears the _new_ queries.

It does _not_ prevent a new request from being made to the API - all refs will still be updated, but the 'stable' ones won't be cleared out the way they are now. When the API returns, the 'stable' values will be updated, but it is expected that there will be minimal changes - the app will re-render, but visually not much should change

mmcgahan: But yes, dynamic refs should definitely prevent stale event data from ever showing up :)

eilinora: > It does not prevent a new request from being made to the API - all refs will still be updated, but the 'stable' ones won't be cleared out the way they are now. When the API returns, the 'stable' values will be updated, but it is expected that there will be minimal changes - the app will re-render, but visually not much should change

giphy

PR #386, opened 2017-10-12
Wp 549 rest api error message

mmcgahan:
"Error: REST API error REST API error" isn't that informative. From the logs, this usually happens when we get an HTML response, and the REST API HTML responses usually includes better error info in the `<title>` tag. This PR logs _that_ for better Error Reporting and Stackdriver logs

### DIFF
--- a/packages/mwp-api-proxy-plugin/src/util/receive.js
+++ b/packages/mwp-api-proxy-plugin/src/util/receive.js
@@ -1,5 +1,5 @@
 import querystring from 'qs';
-import { logger, serializers } from 'mwp-logger-plugin';
+import { logger } from 'mwp-logger-plugin';
 
 import { API_PROXY_PLUGIN_NAME } from '../config';
 import { coerceBool, toCamelCase } from './stringUtils';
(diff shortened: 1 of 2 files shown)


packages/mwp-api-proxy-plugin/src/util/receive.js
```diff
@@ -203,17 +202,21 @@ export const makeLogResponse = request => ([response, body]) => {
 			errorMessage = JSON.stringify(info.errors[0]) || body;
 		} catch (err) {
 			// probably not JSON, could be an HTML response
-			errorMessage = 'REST API error';
+			const titleContent = /<title>(.+?)<\/title>/.exec(body);
```
willh-meetup: Is the descriptive error message in the `title` something we are doing, or is it something fastly does?

Mostly, just curious what sort of errors we are capturing with title tags.

mmcgahan: It's something the REST API does - e.g. for those 503 errors, the body of the response is:

```
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
 "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html>
  <head>
    <title>503 Request body sent more than once</title>
  </head>
  <body>
    <h1>Error ..."   
```

mmcgahan: p.s. I think this specific 503 error _might_ appear if we're sending a request body with a GET request - I need to follow up on that

willh-meetup: Cool.

PR #389, opened 2017-10-18
throw out members/self for logged-out members

mmcgahan:
`members/self` fails for `id=0`, so we shouldn't allow that query to be processed.

This is a breaking change - consumer applications will need to update their code to account for cases where `state.api.self` is empty. This will make the applications more predictable because they will explicitly have to define how 'logged out' state is determined and represented in the UI

### DIFF
--- a/packages/mwp-store/src/util/fetchUtils.js
+++ b/packages/mwp-store/src/util/fetchUtils.js
@@ -4,15 +4,14 @@ export const MEMBER_COOKIE_NAME = isProd
 	? 'MEETUP_MEMBER'
 	: 'MEETUP_MEMBER_DEV';
 const getIsLoggedIn = (memberCookie: ?string) => {
-	return memberCookie && !/id=0&/.test(memberCookie);
+	return Boolean(memberCookie) && !/id=0&/.test(memberCookie);
 };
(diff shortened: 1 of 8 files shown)


packages/mwp-store/src/util/fetchUtils.js
```diff
@@ -1,5 +1,19 @@
 // @flow
-
+const isProd = process.env.NODE_ENV === 'production';
+export const MEMBER_COOKIE_NAME = isProd
+	? 'MEETUP_MEMBER'
+	: 'MEETUP_MEMBER_DEV';
+const getIsLoggedIn = (memberCookie: ?string) => {
+	return memberCookie && !/id=0&/.test(memberCookie);
+};
+export const getValidQueries = memberCookie => (
```
willh-meetup: nitpick: `getNonAuthedQueries` or something maybe?

mmcgahan: I refactored so that this function returns a true/false filtering test function rather than a function that takes an array and returns the filtered array, which makes the usage in `fetchQueries` more expressive (the `.filter` is explicit) and will make it easier to add in any other query filtering that we might want to do later on.

PR #393, opened 2017-10-23
use redux state self to determine logged-in status

mmcgahan:
the MEETUP_MEMBER cookie is not readable in the browser (it's set to HTTP-Only), so we need to consume member info/status from Redux state.

### DIFF
--- a/packages/mwp-store/src/server/fetchQueries.js
+++ b/packages/mwp-store/src/server/fetchQueries.js
@@ -1,21 +1,16 @@
 // @flow
 import 'rxjs/add/operator/map';
 import 'rxjs/add/operator/toPromise';
-import {
-	parseQueryResponse,
-	MEMBER_COOKIE_NAME,
-	getAuthedQueryFilter,
-} from '../util/fetchUtils';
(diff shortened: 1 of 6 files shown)


packages/mwp-store/src/server/fetchQueries.js
```diff
@@ -1,21 +1,16 @@
...
-	getAuthedQueryFilter,
-} from '../util/fetchUtils';
+import { parseQueryResponse, getAuthedQueryFilter } from '../util/fetchUtils';
 
 /**
  * on the server, we can proxy the API requests directly without making a
  * request to the server's own API proxy endpoint
  */
-export default (request: HapiRequest) => () => (
+export default (request: HapiRequest) => (apiUrl, member) => (
```
eilinora: Not seeing where `apiUrl` is being used

mmcgahan: it's not used on the server, but it's part of the `fetchQueries` signature, so I've explicitly named it here

mmcgahan:
timeout in the integration test - i've bumped up the test timeout to take care of it and restart the PR build. will merge when that passes

PR #397, opened 2017-10-31
implemented individual API request timeouts and err objects

mmcgahan:
Instead of failing the entire parent request when one REST API endpoint fails (either through timeout or some other 'Internal Server Error'), this PR updates the proxy logic to apply timeouts to each individual API request, and will inject a 'mocked' error object when a request fails - other responses will be collected normally and the overall batched response to the parent request from the client should response normally.

REST API request errors will still be logged, but we should _not_ see as many 500 errors on our end, and I'm hoping that this setup will bypass the error logging in New Relic, which shouldn't send us PagerDuty alerts for API timeouts, which our team can't do anything about anyway.

### DIFF
--- a/packages/mwp-api-proxy-plugin/src/util/send.js
+++ b/packages/mwp-api-proxy-plugin/src/util/send.js
@@ -25,26 +25,32 @@ const MOCK_RESPONSE_OK = {
 	// minimal representation of http.IncomingMessage
 	statusCode: 200,
 	statusMessage: 'OK',
-	headers: {
-		'x-meetup-request-id': 'mock request',
-	},
 };
+const API_TIMEOUT_RESPONSE = {
+	statusCode: 408,
+	statusMessage: 'API Request Timeout',
+};
+const makeAPIErrorResponse = err => ({
+	statusCode: 500,
+	statusMessage: err.message,
+});
 
 /*
  * When an API response is being mocked (dev only), this response will be used
  */
-function makeMockResponseOk(requestOpts) {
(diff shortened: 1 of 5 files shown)


packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -269,34 +279,33 @@ export function getExternalRequestOpts(request) {
...
 /**
  * Make a real external API request, return response body string
  */
 export const makeExternalApiRequest = request => requestOpts => {
 	return externalRequest$(requestOpts)
-		.do(
-			// log errors
-			null,
-			err => {
+		.catch(err => {
```
mmcgahan: this is the API error response 'catcher' - instead of bubbling the error up to the parent request handler, it catches and replaces the error with a valid JS object - the application will then have to deal with error data when rendering, but the platform itself will not crash the whole request.

willh-meetup: Right, so very possible we'll still see errors if the request isn't being handled properly  in components and containers right now?

willh-meetup: I think this is OK (since things are already broken), just want to confirm.

mmcgahan: correct - I'm sort of hoping that this will expose some incorrect error-case handling in the apps so that we can find out better ways of managing that. I think there are a _lot_ of places in the app where errors are either ignored or automatically replaced with 'default' values, which isn't ideal.

If I were to re-create the query system, I would structure the error responses differently. Relay has some interesting patterns.

PR #398, opened 2017-11-01
use `Intl.DateTimeFormat` to automatically track NYC timestamp

mmcgahan:
This refactor will automatically account for daylight savings, and doesn't require any hard-coded knowledge about timezones or offsets.

Still annoying that we are sending an ISO 8601 string pretending that NYC is UTC, but whatevs

### DIFF
--- a/packages/mwp-tracking-plugin/src/activity.js
+++ b/packages/mwp-tracking-plugin/src/activity.js
@@ -6,6 +6,37 @@ import { ACTIVITY_PLUGIN_NAME } from './config';
 const YEAR_IN_MS: number = 1000 * 60 * 60 * 24 * 365;
 
 /*
+ * Get a date object that is shifted by the offset of the supplied timezone
+ */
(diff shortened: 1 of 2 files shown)


packages/mwp-tracking-plugin/src/activity.js
```diff
@@ -15,34 +46,32 @@ const YEAR_IN_MS: number = 1000 * 60 * 60 * 24 * 365;
...
-) => (request: Object, trackInfo: Object) => {
-	const requestHeaders = request.headers;
-
-	// Takes in desired time to convert and applies offset
-	const offset = new Date().getTimezoneOffset() * 60000; // gets detected timezone + converts to milliseconds
-	const nowUTC = new Date(Date.now() + offset);
-
-	// generates new date object taking the time in milliseconds and
-	// adds the runtime environment's timezone offset
-	const NY_OFFSET = -14400000;
```
mmcgahan: this line won't change with daylight savings, so can't be used next week :)

PR #399, opened 2017-11-03
WP 589 internal endpoint auth

mmcgahan:
I had a hard time replicating the prod behavior in dev, even with sending what I think were identical requests to the ones that logged errors in the prod logs.

However, part of the challenge is that the error logs don't contain the header information for the failing API requests - this data is critical for debugging, so this PR mainly adds that in.

It also adds more cautious CSRF setting behavior - instead of _always_ writing a new UUID for MEETUP_CSRF, it instead will re-use a CSRF cookie that is already set. This may make 'internal' errors less common, since the check for 'internal' requests relies primarily on accurate MEETUP_CSRF and handling.

### DIFF
--- a/packages/mwp-api-proxy-plugin/src/util/send.js
+++ b/packages/mwp-api-proxy-plugin/src/util/send.js
@@ -222,7 +222,6 @@ export function parseRequestHeaders(request) {
 	delete externalRequestHeaders['accept-encoding']; // let app server set 'accept'
 	delete externalRequestHeaders['content-length']; // original request content-length is irrelevant
 	delete externalRequestHeaders['content-type']; // the content type will be set in buildRequestArgs
-	console.log(JSON.stringify(externalRequestHeaders, null, 2));
 
 	return externalRequestHeaders;
 }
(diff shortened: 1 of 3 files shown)


packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -165,7 +165,18 @@ export function getAuthHeaders(request) {
...
 		};
 	}
+
+	// Cookie + CSRF auth: need have matching UUID in MEETUP_CSRF cookie and 'csrf-token' header
 	const cookies = { ...request.state };
+	const csrfCookie =
+		process.env.NODE_ENV === 'production'
+			? cookies.MEETUP_CSRF
+			: cookies.MEETUP_CSRF_DEV;
+	if (csrfCookie) {
```
willh-meetup: I didn't look too closely at this... but is it possible that the cookie can expire before this happens?

mmcgahan: in this case, if the cookie expires, it won't be sent in the `request` - this is just a relay of what the browser sends, so we shouldn't have to check any other metadata

PR #400, opened 2017-11-08
Match PropType

cranberyxl:
Start of exporting proptypes for properties that are injected by mwp code


cranberyxl:
Even better, I'll see if I can just forward that proptype

PR #403, opened 2017-11-13
better behaved API proxy and activity tracking logs

mmcgahan:


packages/mwp-api-state/src/sync/index.js
```diff
@@ -78,7 +78,7 @@ export const getNavEpic = (routes, baseUrl) => {
 				const { referrer = {} } = state.routing;
 				// inject request metadata from context, including `store.getState()`
 				const requestMetadata = {
-					referrer: referrer.pathname || '',
+					referrer: referrer.pathname || state.config.entryPath || '',
```
eilinora: whats entry path?

mmcgahan: added a code comment, but it's the path that the user entered the app on - i.e. the server-rendered path, not necessarily the current path determined by navigating within the SPA

PR #404, opened 2017-11-13
mwp-cli 1 or 2 peerDependency

cranberyxl:


package.json
```diff
@@ -87,11 +87,14 @@
     "swarm-icons": "1.3.259",
     "tough-cookie": "2.3.2"
   },
+  "peerDependencies": {
+    "mwp-cli": "^1.1.268 || ^2.0.0"
+  },
```
mmcgahan: this peerDep doesn't need to be here because there is no top-level package that can be installed

PR #408, opened 2017-11-29
Start of refactor from redux-observable to a Promise-based API middleware

mmcgahan:
This PR provides the first refactor of an observable-based 'epic' into a Promise-based 'epic'. There are a number of other epics that will need to be refactored, but this is the first step.

I'm using an 'epic' pattern similar to what redux-observable defines, where instead of creating an observable stream that takes a stream of actions as an input and emits a stream of actions, (actions in, actions out), the new epics create a new Promise for each action that comes in, which returns an array of actions (could be empty, a single action, or multiple actions, but it always resolves to a single array).

This conversion allows the core functional blocks of the API middleware and the tests to remain largely the same.

redux-observable also provides utilities for composing epics and turning them into proper Redux middleware functions. For this first refactor, I've written a couple of parallel helpers that work with Promises: `createEpicMiddleware` and `composeMiddleware` that allows the top-level `getApiMiddleware` export to remain the same - consumers do not have to be aware of the implementation details since the behavior of dispatched Redux actions is the same.

### DIFF
--- a/packages/mwp-api-state/src/index.js
+++ b/packages/mwp-api-state/src/index.js
@@ -1,6 +1,10 @@
-import { combineEpics, createEpicMiddleware } from 'redux-observable';
+import {
+	combineEpics,
+	createEpicMiddleware as createEpicMiddlewareRO,
+} from 'redux-observable';
+import { createEpicMiddleware } from './redux-promise-epic';
 
-import getSyncEpic from './sync';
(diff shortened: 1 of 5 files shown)


mmcgahan:
full Flow typing and docs to come in a separate ticket - I'm eager to get this out

PR #410, opened 2017-12-01
Convert CacheEpic to Promises

mmcgahan:
This is the next epic in the conversion sequence, one more PR to come

### DIFF
--- a/packages/mwp-api-state/src/cache/index.test.js
+++ b/packages/mwp-api-state/src/cache/index.test.js
@@ -7,6 +7,7 @@ import {
 import * as api from '../sync/apiActionCreators';
 
 import { makeCache } from './util';
+import { CACHE_CLEAR } from './cacheActionCreators';
 import getCacheEpic from './';
 
 const MOCK_QUERY = mockQuery(MOCK_RENDERPROPS);
@@ -26,7 +27,7 @@ function populateCacheEpic(CacheEpic) {
 
 function clearCacheEpic(CacheEpic) {
 	// clear the cache with CACHE_CLEAR
-	return CacheEpic({ type: 'CACHE_CLEAR' }).then(() => CacheEpic);
+	return CacheEpic({ type: CACHE_CLEAR }).then(() => CacheEpic);
 }
 
 const testForEmptyCache = (action = apiRequestAction) => CacheEpic =>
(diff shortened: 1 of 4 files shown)


packages/mwp-api-state/src/cache/index.test.js
```diff
@@ -25,31 +21,27 @@ function makeCacheEpic() {
...
-	const apiSuccessAction$ = ActionsObservable.of(MOCK_SUCCESS_ACTION);
-	return CacheEpic(apiSuccessAction$).toPromise().then(() => CacheEpic);
+	return CacheEpic(MOCK_SUCCESS_ACTION).then(() => CacheEpic);
 }
 
 function clearCacheEpic(CacheEpic) {
 	// clear the cache with CACHE_CLEAR
-	const clearAction$ = ActionsObservable.of({ type: 'CACHE_CLEAR' });
-	return CacheEpic(clearAction$).toPromise().then(() => CacheEpic);
+	return CacheEpic({ type: 'CACHE_CLEAR' }).then(() => CacheEpic);
```
eilinora: isn't this a const?

packages/mwp-api-state/src/sync/index.js
```diff
@@ -64,85 +52,74 @@ export function getDeprecatedSuccessPayload(successes, errors) {
...
+		if (![LOCATION_CHANGE, SERVER_RENDER].some(type => type === action.type)) {
+			return Promise.resolve([]);
+		}
+		const { payload: location } = action;
+		const state = store.getState();
+		const { referrer = {} } = state.routing;
+		// inject request metadata from context, including `store.getState()`
+		const requestMetadata = {
+			referrer: referrer.pathname || state.config.entryPath || '',
+			logout: location.pathname.endsWith('logout'), // assume logout route ends with logout
```
eilinora: would be nice to have more notes here about why/what this is for or a link to those notes

packages/mwp-api-state/src/cache/index.test.js
mmcgahan: so it would appear!

packages/mwp-api-state/src/sync/index.js
mmcgahan: okay - this is actually sort of legacy code when we were thinking about how the app should handle logout, but it hasn't been implemented in any application so I'll make that clear

PR #411, opened 2017-12-01
Convert Mutate epics to Promises

mmcgahan:
Last in the conversion PRs to remove redux-observable

This PR includes a version bump (10.1) and the removal of redux-observable from the platform


packages/mwp-api-state/src/mutate/index.js
```diff
@@ -35,32 +35,30 @@ import 'rxjs/add/operator/filter';
...
-			if (process.env.NODE_ENV !== 'production') {
-				// using a dev endpoint, render a deprecation warning
-				console.warn(
-					`This application is using Post/Delete middleware through ${type}.
-See the platform Queries Recipes docs for refactoring options:
-https://github.com/meetup/meetup-web-platform/blob/master/docs/Queries.md#recipes`
-				);
-			}
-		})
-		.map(({ payload: { query, onSuccess, onError } }) => ({
```
eilinora: only thing i'm unsure of is why you don't need to map through actions anymore  as i'm assuming we are still getting an array of actions but perhaps that is false now

mmcgahan: this is an Observable `.map`, which in this case means "Each time a matching action is dispatched, do this thing" - so it's a one-at-a-time thing rather than acting on a complete array. The new code just extracts `onSucces`, `onError` from each action and assigns them to `meta` directly on line 58.

PR #412, opened 2017-12-05
simplify auth plugin - `memberCookie` + `csrfToken` only

mmcgahan:
This PR rips out all of the Oauth processing that was implemented in the mwp-auth-plugin package and replaces it with a simple function that supplies a fallback `memberCookie` and `csrfToken` value when one or both do not exist in an incoming request. It does _not_ set any cookies in the response to client. It _only_ sets `request.auth.credentials`, which can then be consumed by other server code (primarily mwp-api-proxy-plugin when constructing REST API requests).

Less code, fewer observables, less dependence on network requests, faster processing. Win.


packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -20,6 +19,10 @@ import config from 'mwp-cli/src/config';
 import { API_PROXY_PLUGIN_NAME } from '../config';
 
 export const API_META_HEADER = 'X-Meta-Request-Headers';
+const MEMBER_COOKIE_NAME =
+	process.env.NODE_ENV === 'production' ? 'MEETUP_MEMBER' : 'MEETUP_MEMBER_DEV';
```
eilinora: obnoxious nitpick would be nice if these would more const based but nice to have

eilinora: would be nice if some of these were more consts but nice to have

packages/mwp-auth-plugin/src/index.js
```diff
@@ -1,342 +1,57 @@
...
-			request => {
-				const credentials =
-					request.state[getAuthType(request)] ||
-					request.plugins.requestAuth.oauth_token;
-				reply.continue({ credentials, artifacts: credentials });
-			},
-			err => reply(err, null, { credentials: null })
-		);
+const mockCookies = {
+	MEETUP_MEMBER:
```
eilinora: waaaaa :) but really we should describe this :)

packages/mwp-api-proxy-plugin/src/util/send.js
mmcgahan: Yeah I had them as consts for awhile, but then needed a common place to import them from, and it made the package dependencies reeeeally messy because this name is needed in a few different packages that don't directly relate to each other. I have a Jira ticket to make a `mwp-config` package that would have zero MWP dependencies and therefore would be simple to import anywhere that needed 'shared' MWP config/const info, so I think this concern can be resolved there

packages/mwp-auth-plugin/src/index.js
mmcgahan: It's described here https://github.com/meetup/meetup-web-platform/pull/412/files#diff-174f25f23863d6b2b6ba3aa79cfcc151R19

But I can add another comment here

PR #414, opened 2017-12-09
Optimize avro encoder

mmcgahan:
For the `mwp-consumer` dummy application, this code change improves response latency and throughput by 50%, and reduces the rate of garbage collection by half.

happy owl

### DIFF
--- a/packages/mwp-api-state/src/sync/index.js
+++ b/packages/mwp-api-state/src/sync/index.js
@@ -88,8 +88,8 @@ export const getNavEpic = (routes, baseUrl) => {
 			resolvePrevQueries,
 		]).then(([newQueries, previousQueries]) => {
 			if (newQueries.filter(q => q).length === 0) {
-				// no valid queries - ignore
-				return [];
+				// no valid queries - jump straight to 'complete'
+				return [api.complete([])];
 			}
 			// perform a fast comparison of previous route's serialized queries
 			// with the new route's serialized queries. All state refs for
--- a/packages/mwp-tracking-plugin/package.json
+++ b/packages/mwp-tracking-plugin/package.json
@@ -25,7 +25,7 @@
   "homepage": "https://github.com/meetup/meetup-web-platform#readme",
   "dependencies": {
     "@google-cloud/pubsub": "0.14.2",
-    "avsc": "5.0.5",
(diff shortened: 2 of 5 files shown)


packages/mwp-api-state/src/sync/index.js
```diff
@@ -88,8 +88,8 @@ export const getNavEpic = (routes, baseUrl) => {
 			resolvePrevQueries,
 		]).then(([newQueries, previousQueries]) => {
 			if (newQueries.filter(q => q).length === 0) {
-				// no valid queries - ignore
-				return [];
+				// no valid queries - jump straight to 'complete'
+				return [api.complete([])];
```
mmcgahan: this is sort of unrelated to the code optimization, but is useful regardless - it makes sure the server can render even if there are no queries matched in the route.

packages/mwp-tracking-plugin/package.json
```diff
@@ -25,7 +25,7 @@
   "homepage": "https://github.com/meetup/meetup-web-platform#readme",
   "dependencies": {
     "@google-cloud/pubsub": "0.14.2",
-    "avsc": "5.0.5",
+    "avsc": "5.1.1",
```
mmcgahan: this package upgrade didn't make a practical difference, but it's good to stay up to date.

packages/mwp-tracking-plugin/src/util/avro.js
```diff
@@ -101,22 +101,29 @@ const activity = {
...
-const avroSerializer: Object => Serializer = schema => data => {
-	const record = avro.parse(schema).toBuffer(data);
-	// data.timestamp _must_ be ISOString if it exists
-	const timestamp = data.timestamp || new Date().toISOString();
-	const analytics = {
-		record: record.toString('base64'),
-		schema: `gs://meetup-logs/avro_schemas/${schema.name}_${schema.doc}.avsc`,
-		date: timestamp.substr(0, 10), // YYYY-MM-DD
+const avroSerializer: Object => Serializer = schema => {
+	const codec = avro.parse(schema);
```
mmcgahan: this is the main optimization - the codec created by `avro.parse(schema)` is now instantiated once when the module is first loaded rather than once-per-request, which turns out to be expensive

PR #416, opened 2017-12-11
mwp-config

mmcgahan:
Moving the config code from mwp-cli into its own package in this repo. Once this is merged, I can update mwp-cli with a breaking change that removes the config code there and imports config from the published `mwp-config`

There's not actually much to review here - the brand-new files are copied directly from mwp-cli and were in pretty good shape in terms of being standalone CommonJS modules, so there aren't really any new edits to check out there. The remaining changes update the dependencies of the other mwp-* packages that were formally import config from mwp-cli but can now reference mwp-config directly.


package.json
```diff
@@ -65,7 +65,7 @@
     "lint-staged": "4.0.1",
     "meetup-web-components": "3.2.1592",
     "meetup-web-mocks": "1.0.217",
-    "mwp-cli": "3.0.327",
+    "memwatch-next": "0.3.0",
```
mmcgahan: this is just moving a devDependency from mwp-consumer into the repo root, as recommended by Lerna (our monorepo management tool)

packages/mwp-config/locales.js
```diff
@@ -0,0 +1,16 @@
+module.exports = [
```
mmcgahan: @willh-meetup this and the `localesSecondary` file below are where the locale config will be coming from going forward, although I'd like to organize things a little better at some point - I'm wondering if we might want have a module that exports `{ primary: [], secondary: [], beta: [] }` or something.

packages/mwp-config/README.md
```diff
@@ -0,0 +1,48 @@
+# Config
```
mmcgahan: all the new files from here on down in this PR are direct copies from mwp-cli

packages/mwp-config/locales.js
willh-meetup: Yeah, that's a good call. Would be nice to export as a single module, since they are often used in relation to eachother.

mmcgahan:
I'll hold off on merging this until the ru-RU language support is established in mwp-cli and running in production - then I'll pull that `locales.js` update and get it set up in mup-web and pro-web

mmcgahan:
@willh-meetup is there anything that needs to change here in order for the Russian support to remain intact?

PR #418, opened 2017-12-12
Enforce the use of method-specific action creators

mmcgahan:
This PR addresses two sources of confusion/bugs in using the API action creators to make API requests

1. The platform cannot batch API requests with different methods, e.g. sending a POST query along with a GET query to `requestAll` won't work, although the current interface allows you to manually call `requestAll` in that way. This PR no longer `export`s `requestAll` as an action creator - you may only dispatch method-specific API request actions (`api.get/post/patch/del`).
2. The current code doesn't provide a method-specific way of batching mutation queries (POST/PATCH/DELETE), so this PR updates the interface for the single-query method-specific action creators to allow _arrays_ of queries to be passed in to `api.get/post/patch/del`. All the contained queries will be assigned the correct `meta.method` internally.

The net effect is that consumer applications should no longer have to be aware of the `query.meta.method` property - it should not ever be managed manually.


packages/mwp-api-state/src/sync/apiActionCreators.js
```diff
@@ -34,19 +47,29 @@ export function requestAll(queries: Array<Query>, meta: ?Object) {
...
-	// delegate to `requestAll`
-	const requestAction = requestAll([query], meta);
-	// modify the promise to automatically pull out the single response
-	// corresponding to the single query
-	requestAction.meta.promise = requestAction.meta.request.then(
-		responses => responses[0]
-	);
-	return requestAction;
+};
+const _applyMethod = method => (query: Query | Array<Query>, meta: ?Object) => {
```
eilinora: 🎉

packages/mwp-api-state/src/mutate/index.test.js
```diff
@@ -11,7 +11,7 @@ describe('postEpic', () => {
 	it('returns an api.post(query) action', () => {
 		return postEpic(MOCK_POST_ACTION).then(([action]) => {
 			const { query, onSuccess, onError } = MOCK_POST_ACTION.payload;
-			const expectedAction = api.del(query, { onSuccess, onError });
+			const expectedAction = api.post(query, { onSuccess, onError });
```
eilinora: Just want to validate this is intentional. Makes sense based on test description though

mmcgahan: yeah, I'm sort of surprised that this wasn't throwing an error previously, but maybe there was something else not working right

PR #420, opened 2017-12-18
remove temporary file creation for uploads

mmcgahan:
Hapi has a few different settings for how the payload for multipart requests (i.e. file uploads) can be parsed. It turns out that the default option, which parses the file contents in a Buffer, can be used in the forwarded request to the REST API with only a little tweaking (adding a 'filename' descriptor) - we do not need to write the file to disk and then read from that in order to make the upload to the REST API.

This PR rolls back the write-to-disk file handling and simplifies the logic related to passing the uploaded file to the REST API as a Buffer instead of a file Stream.

I'm really not sure if this will affect memory usage in prod, but it can't hurt. It should also make file uploads marginally faster.

In the future, I'd like to explore the possibility of setting up a proper proxy handler for file uploads using https://github.com/hapijs/h2o2, which _should_ allow files to be sent immediately to the REST API as they are being uploaded rather than the current handling which reads file uploads fully into memory before sending them on to the REST API.

### DIFF
--- a/packages/mwp-api-proxy-plugin/src/util/send.js
+++ b/packages/mwp-api-proxy-plugin/src/util/send.js
@@ -1,5 +1,4 @@
 import http from 'http';
-import fs from 'fs';
 
 import querystring from 'qs';
 import url from 'url';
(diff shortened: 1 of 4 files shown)


packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -225,24 +226,15 @@ export function parseRequestHeaders(request) {
  * 
  * References to the uploaded files are stored here for cleanup later on.
```
willh-meetup: Might as well remove this comment since it's not true anymore

willh-meetup:
👇 Tests are failing, though.

PR #423, opened 2017-12-26
Mw 2907 uas

jaclynj:
Jira ticket: https://meetup.atlassian.net/browse/MW-2907

**meetup-web-platform**
- Add mobile-detect library for detecting user agent

**mwp-core**
- Use mobile-detect library to detect the device type and add to redux state.config
- later when we update fastly to re-write the ua header we can remove mobile-detect library, but the code is compatible with either way.

**mwp-app-render**
- create `connectedWithMatchMedia` HOC that sets a default media based on state.config.device ( added in state.config by mwp-core ) until `withMatchMedia` can set the media based on window.matchMedia
- Add test for `connectedWithMatchMedia`

**mwp-app-route-plugin**
- Add 'User-Agent' to vary header

**mwp-tracking-plugin**
- check `x-original-user-agent` header first, if not check user-agent header. We will be making an update for fastly to update the user-agent header to either 'mobile' or 'desktop', and copying the original user agent string to `x-original-user-agent`

**mwp-consumer**
- small update to readme file


packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
```diff
@@ -0,0 +1,80 @@
...
+	const wrappedComponentName =
+		WrappedComponent.displayName || WrappedComponent.name || 'Component';
+
+	ConnectedWithMatchMedia.WrappedComponent = WrappedComponent;
+	ConnectedWithMatchMedia.displayName = `ConnectedWithMatchMedia(${wrappedComponentName})`;
+	
+	/* connect to redux state */
+	const ConnectedComponent = connect(mapStateToProps)(ConnectedWithMatchMedia);
+	/* return wrapped component withMatchMedia */
+	return withMatchMedia(ConnectedComponent);
```
jaclynj: `withMatchMedia` wraps the result of connecting the wrapped component. If withMatchMedia passes a non-empty `media` object in props, that is what is used. Otherwise we use a default `media` object based on device type.

Not sure if it matters whether withMatchMedia wraps the connected + wrapped component, which is what's happening here, or if we should wrap the wrapped component withMatchMedia and then wrap *that* with the connected component. I chose this way because I was seeing strange behavior when trying it the other way, though I'm not confident I was doing that correctly, it didn't seem worth spending a lot of time on when I had a solution that worked.

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
```diff
@@ -0,0 +1,80 @@
...
+	const wrappedComponentName =
+		WrappedComponent.displayName || WrappedComponent.name || 'Component';
+
+	ConnectedWithMatchMedia.WrappedComponent = WrappedComponent;
+	ConnectedWithMatchMedia.displayName = `ConnectedWithMatchMedia(${wrappedComponentName})`;
+	
+	/* connect to redux state */
+	const ConnectedComponent = connect(mapStateToProps)(ConnectedWithMatchMedia);
+	/* return connected wrapped component withMatchMedia */
+	return withMatchMedia(ConnectedComponent);
```
jaclynj: withMatchMedia wraps the result of connecting the wrapped component. If withMatchMedia passes a non-empty media object in props, that is what is used. Otherwise we use a default media object based on device type.

Not sure if it matters whether withMatchMedia wraps the connected + wrapped component, which is what's happening here, or if we should wrap the wrapped component withMatchMedia and then wrap that with the connected component. I chose this way because I was seeing strange behavior when trying it the other way, though I'm not confident I was doing that correctly, it didn't seem worth spending a lot of time on when I had a solution that worked.

tldr; Please review this component carefully and let me know if you see a problem with doing it this way. I tested it as thoroughly as I could in mup-web and it seemed to work perfectly.

jaclynj: `withMatchMedia` wraps the result of connecting the wrapped component. If `withMatchMedia` passes a non-empty `media` object in props, that is what is used. Otherwise it creates a default `media` object based on device type.

Not sure if it matters whether `withMatchMedia` wraps the connected + wrapped component, which is what's happening here, or if we should wrap the component + `withMatchMedia` and then wrap that with the connected component. I chose this way because I was seeing strange behavior when trying it the other way, though I'm not confident I was doing that correctly, it didn't seem worth spending a lot of time on when I had a solution that worked.

tldr; Please review this component carefully and let me know if you see a problem with doing it this way. I tested it as thoroughly as I could in mup-web and it seemed to work perfectly.

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -53,9 +54,34 @@ const resolveSideEffects = () => ({
...
+		}
+	}
+	// parses user agent string to determine if user is on mobile / tablet / desktop device
+	const md = new MobileDetect(uas);
+	const isMobile = md.mobile() !== null && md.phone() !== null;
+	const isTablet = md.tablet() !== null;
+	return {
+		isMobile,
+		isTablet,
+		isDesktop: !isMobile && !isTablet
```
mmcgahan: I'm not sure we need to have a map of `is<Device>` keys that exactly parallels the `isAt<Size>Up` map defined by `withMatchMedia` - why not use the same map in `state.config` that will be applied by `connectedWithMatchMedia`? It keeps the 'data types' much simpler and reduces the indirection.

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -53,9 +54,34 @@ const resolveSideEffects = () => ({
...
+	notFound: NotFound.rewind()
 });
 
+/**
+ * Using the User agent string + an external library called Mobile-Device we detect if the user
+ * is attempting to render the page on the server using a mobile, tablet, or desktop device
+ * @param  {String} uas User agent string detected from headers
+ * @return {Object}     Object of which category the devices falls within
+ */
+const getDeviceType = uas => {
```
mmcgahan: I'm hoping that Flowtype will error on the missing type definitions for this function, although I just realized that Flow type checking isn't part of the PR build currently - try running `yarn flow` from inside `packages/mwp-core/`

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -240,7 +266,9 @@ const makeRenderer = (
 			initialNow: new Date().getTime(),
 			variants: getVariants(state),
 			entryPath: url.pathname, // the path that the user entered the app on
-		},
+			uas: headers["user-agent"],
```
mmcgahan: what is this used for? Also, just as an aside, `UA` is the conventional acronym for user agent string, rather than `uas`, although I think I would tend toward something more explicit like `userAgent` in this case because `state.config` doesn't offer any clues about how to interpret acronym keynames.

packages/mwp-app-route-plugin/src/route.js
```diff
@@ -12,6 +12,11 @@ export const onPreResponse = {
 	 */
 	method: (request: HapiRequest, reply: HapiReply) => {
 		const response = request.response;
+
+		if (response && response.header) {
```
mmcgahan: `response` is guaranteed to be a valid object with a `header` method here - no need to check for a falsey value

packages/mwp-app-render/src/components/connectedWithMatchMedia.test.jsx
```diff
@@ -0,0 +1,38 @@
...
+describe('connectedWithMatchMedia', () => {
+	const mockStore = createFakeStore({config: {device: {isMobile: true} } })
+	const connectedWithMatchMedia = mount(
+		<Provider store={mockStore}>
+			<TestComponentConnectedWithMatchMedia />
+		</Provider>
+	);
+	it('exists', () => {
+		expect(connectedWithMatchMedia).toMatchSnapshot();
+	});
```
mmcgahan: just to be clever you could write this as

```js
it('exists', expect(connectedWithMatchMedia).toMatchSnapshot);
```

packages/mwp-app-render/src/components/connectedWithMatchMedia.test.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+const MATCH_MEDIA_FN_MOCK = mq => ({
+	matches: false,
+	addListener: jest.fn(),
+	removeListener: jest.fn(),
+});
+
+describe('connectedWithMatchMedia', () => {
+	const mockStore = createFakeStore({config: {device: {isMobile: true} } })
+	const connectedWithMatchMedia = mount(
```
mmcgahan: can this be `shallow`? You can use [`dive`](https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/dive.md) to get the child of a shallow-mounted component

packages/mwp-app-render/src/components/connectedWithMatchMedia.test.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const createFakeStore = fakeData => ({
+	getState() {
+		return fakeData;
+	},
+	dispatch() {},
+	subscribe() {},
+});
+
+const TestComponent = () => (<div>Hello World</div>);
```
mmcgahan: Prettier should remove the redundant parens

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
```diff
@@ -0,0 +1,80 @@
...
+ * @return {Object} the config from state
+ */
+const getConfig = state => state.config || {};
+
+/**
+ * Checks if object is empty
+ * @param  {Object} object
+ * @return {Boolean} whether object is empty
+ */
+const isEmpty = object => object === undefined || Object.keys(object).length === 0;
```
mmcgahan: any reason not to check for any falsey value? `null` would throw an error here, but it seems like it should just return `false`

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
```
mmcgahan: I'd love for this to be Flow typed instead of being tagged with JSDoc comments

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
```diff
@@ -0,0 +1,80 @@
...
+import { connect } from 'react-redux';
+import { createStructuredSelector, createSelector } from 'reselect';
+import withMatchMedia from 'meetup-web-components/lib/utils/components/withMatchMedia';
+
+/**
+ * getConfig
+ * @param  {Object} state
+ * @return {Object} the config from state
+ */
+const getConfig = state => state.config || {};
```
mmcgahan: it's subjective, but this seems like over-abstraction to me - we can assume that `state.config` always exists in any MWP app

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
```diff
@@ -0,0 +1,80 @@
...
+			isAtMediumUp: Boolean(isTabletOrDesktop),
+			isAtLargeUp: Boolean(isDesktop),
+		}
+	}
+	return {};
+};
+
+const mapStateToProps = createStructuredSelector({
+	device: createSelector(getConfig, config => (config.device || {})),
+});
```
mmcgahan: the reselect-enhanced selector is overkill IMO - we can avoid the dependency and the complexity by just implementing this as

```js
const mapStateToProps = state => ({ device: state.config.device });
```

(although I'm recommending `const mapStateToProps = state => ({ media: state.config.media });` in this review!)

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * @param {Object} device - object containing device information
+ * @returns {Object} - default media object for `withMatchMedia`
+ */
+const getMediaOrDefault = (media, device) => {
+	if (!isEmpty(media)) {
+		return media;
+	}
+	if ( !isEmpty(device) ) {
```
mmcgahan: MWP has total control over the existence of this value - we can assume `state.config.device` is never empty and can enforce it with Flow types

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
```diff
@@ -0,0 +1,80 @@
...
+		const {isMobile, isTablet, isDesktop} = device;
+		const isTabletOrDesktop = isTablet || isDesktop;
+		return {
+			isAtSmallUp: Boolean(isMobile || isTabletOrDesktop),
+			isAtMediumUp: Boolean(isTabletOrDesktop),
+			isAtLargeUp: Boolean(isDesktop),
+		}
```
mmcgahan: Very subjective, but I'd recommend assigning these to `consts` in the reverse order so that the dependencies are easier to follow, e.g.

```js
const isAtLargeUp = isDesktop;
const isAtMediumUp = isAtLargeUp || isTablet;
const isAtSmallUp = isAtMediumUp || isMobile;
return { isAtSmallUp, isAtMediumUp, isAtLargeUp };
```

Although again, I think this logic should be moved to `server-render.jsx` so that you can assign `state.config.media` directly

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
```diff
@@ -0,0 +1,80 @@
...
+		render() {
+			const {device, media: currentMedia } = this.props;
+			const media = getMediaOrDefault(currentMedia, device);
+			return <WrappedComponent {...this.props} media={media}/>;
+		}
+	};
+	const wrappedComponentName =
+		WrappedComponent.displayName || WrappedComponent.name || 'Component';
+
+	ConnectedWithMatchMedia.WrappedComponent = WrappedComponent;
```
mmcgahan: What is this class prop used for?

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
mmcgahan: The wrapping order makes sense to me and seems reasonable. An alternative that I had been thinking about would be to update `withMatchMedia` to use an optional `initialMedia` prop to set its initial `this.state` in its constructor, which would make this HOC quite simple - it would do something like

```js
const MediaWrappedComponent = withMatchMedia(WrappedComponent);
const ConnectedWithMatchMedia = props => <MediaWrappedComponent initialMedia={props.media} />
```

I think that might be slightly simpler, but functionally it's the same

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * @returns {React.element}
+ */
+const connectedWithMatchMedia = WrappedComponent => {
+
+	/**
+	 * @module ConnectedWithMatchMedia
+	 */
+
+	class ConnectedWithMatchMedia extends React.Component {
```
mmcgahan: this should be a pure functional component

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
jaclynj: I'm not using it. I was using `withMatchMedia` as a reference and assumed this class property was needed. Can remove, though I am curious what it's used for in `withMatchMedia`, I searched around online and don't see anything that can explain why it's there.

packages/mwp-core/src/renderers/server-render.jsx
jaclynj: It's not currently being used. Lori added it when we were pairing, I believe her reasoning was better to have more info just in case. Generally I prefer to avoid abbreviating, when we added it it didn't seem ambiguous but in hindsight I can see why it is. Tagging @eilinora so she can see the feedback/info.

packages/mwp-core/src/renderers/server-render.jsx
jaclynj: Yeah that's definitely a better way to go.

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
mmcgahan: I think `connect` from react-redux also assigns it, presumably in order to maintain a reference to the wrapped component for debugging and such, which is fine. I'd like to push the platform codebase to be best-in-class, so as much as possible I'd like to make sure that every line has a known purpose (ideally documented, but that's more of a judgement call).

Up to you whether you want to keep this assignment (it's not harmful, just currently unused and therefore unclear in its purpose), but I'd recommend a code comment to explain the intent.

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
```diff
@@ -0,0 +1,80 @@
...
+			return <WrappedComponent {...this.props} media={media}/>;
+		}
+	};
+	const wrappedComponentName =
+		WrappedComponent.displayName || WrappedComponent.name || 'Component';
+
+	ConnectedWithMatchMedia.WrappedComponent = WrappedComponent;
+	ConnectedWithMatchMedia.displayName = `ConnectedWithMatchMedia(${wrappedComponentName})`;
+	
+	/* connect to redux state */
```
mmcgahan: just a little thing, but since `connect` is a HOC with well-understood behavior, I don't think it needs a comment, and `return withMatchMedia(ConnectedComponent)` is also pretty legible as a HOC, so `/* return connected wrapped component withMatchMedia */` is probably unnecessary.

Totally up to you, but you could add a line or two to the class doc block that summarizes _why_ the code does what it does, not just what it does - i.e. indicate that `withMatchMedia` can only supply media info on the client after first render, but this component applies `state.config.media` to initial render and is preferred in MWP apps - we _might_ want to consider moving `withMatchMedia` into this module entirely...

packages/mwp-app-render/src/components/connectedWithMatchMedia.test.jsx
```diff
@@ -0,0 +1,38 @@
+import jsdom from 'jsdom';
```
mmcgahan: just as a heads up, as we move more React modules into the platform, I've been thinking we might need to set up a separate test run for things that are designed to be run in the browser - currently Jest is [configured to run](https://github.com/meetup/meetup-web-platform/blob/master/jest.config.json#L15) in a `node` environment, without the jsdom overhead, but we could have a parallel Jest config that targeted test modules with a `.test.dom.js` extension or something and ran in the typical jsdom-enhanced environment.

package.json
```diff
@@ -33,7 +33,8 @@
   "homepage": "https://github.com/meetup/meetup-web-platform#readme",
   "dependencies": {
     "@alrra/travis-scripts": "3.0.1",
-    "lerna": "2.1.2"
+    "lerna": "2.1.2",
+    "mobile-detect": "false1.4.0"
```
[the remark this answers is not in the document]
jaclynj: This is actually not a typo. When you do `yarn add mobile-detect` this is what gets added to package.json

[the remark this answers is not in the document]
mmcgahan: I think there might be something funky with your Yarn install - double check the version you're running and update if you can https://github.com/yarnpkg/yarn/issues/4104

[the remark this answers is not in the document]
mmcgahan: Actually, I apologize - reading that GitHub issue a little more closely, it looks like we can now use `save-exact true` in our `.yarnrc` rather than `save-prefix false`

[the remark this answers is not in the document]
jaclynj: ah your change worked! Thanks

packages/mwp-app-render/src/components/connectedWithMatchMedia.test.jsx
jaclynj: hmm prettier hasn't been running on commit for these new files

packages/mwp-core/yarn.lock
```diff
@@ -2,7 +2,21 @@
 # yarn lockfile v1
 
 
-"@google-cloud/common@^0.13.0":
+"@google-cloud/common-grpc@^0.4.0":
```
jaclynj: Not sure why all this has been added but it happens when I run `yarn`. I can remove my changes from this pr but it's strange it's being added right?

mmcgahan: hmm we might need to reset the yarn.lock files for this repo. Feel free to ignore for now

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
```diff
@@ -0,0 +1,35 @@
...
+});
+
+/**
+ * connectedWithMatchMediaComponent
+ * HOC that provides an `initialMedia` prop based on `media` in state.config which is determined by user agent
+ * because withMatchMedia cannot detect the media until after initial render.
+ *
+ * `media` prop is provided from redux state in config.
+ */
+const connectedWithMatchMedia = <Props: {}>(
```
mmcgahan: I'm not sure what the `<Props: {}>` does on this line - is it a way of declaring a type inline? I've only seen something like `type Props = Object;` before.

I've mainly used this as a reference https://flow.org/en/docs/react/components/#toc-stateless-functional-components

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
```diff
@@ -0,0 +1,35 @@
...
+	WrappedComponent: React.ComponentType<Props>
+): ConnectedComponentClass<*, *> => {
+	const MediaWrappedComponent = withMatchMedia(WrappedComponent);
+	const ConnectedWithMatchMedia = props => (
+		<MediaWrappedComponent {...props} initialMedia={props.media} />
+	);
+	//
```
mmcgahan: 🔪

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
```diff
@@ -0,0 +1,35 @@
...
+	const wrappedComponentName =
+		WrappedComponent.displayName || WrappedComponent.name || 'Component';
+
+	ConnectedWithMatchMedia.displayName = `ConnectedWithMatchMedia(${wrappedComponentName})`;
+
+	return connect(mapStateToProps)(ConnectedWithMatchMedia);
+};
+
+export default connectedWithMatchMedia;
```
mmcgahan: totally fine to skip this, but what do you think about calling this HOC `connectWithMatchMedia` instead of `connected...` - mainly to keep the same grammar as `connect` from Redux

packages/mwp-app-render/src/components/connectedWithMatchMedia.test.jsx
```diff
@@ -0,0 +1,42 @@
...
+	matches: false,
+	addListener: jest.fn(),
+	removeListener: jest.fn(),
+});
+
+describe('connectedWithMatchMedia', () => {
+	const mockStore = createFakeStore({
+		config: { media: { isAtSmallUp: true } },
+	});
+	console.log('hi');
```
mmcgahan: hi

packages/mwp-app-route-plugin/src/route.js
```diff
@@ -12,6 +12,9 @@ export const onPreResponse = {
 	 */
 	method: (request: HapiRequest, reply: HapiReply) => {
 		const response = request.response;
+
+		response.header('vary', 'User-Agent')
```
mmcgahan: code comment? e.g. 

```
// Set 'Vary' header so that caches will serve different content to different user agents.
// Currently, we use Fastly to rewrite the user agent string to 'desktop' or 'mobile' in order
// to minimize cache fragmentation
```

it will be useful to know the reason for this setting in 10 years when it doesn't make sense any more

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -53,6 +54,34 @@ const resolveSideEffects = () => ({
 	notFound: NotFound.rewind(),
 });
 
+/**
+ * Using the User agent string + an external library called Mobile-Device we detect if the user
+ * is attempting to render the page on the server using a mobile, tablet, or desktop device
+ * @param  {String} userAgent User agent string detected from headers
+ * @return {Object} Object of which category the devices falls within
```
mmcgahan: you can drop the JSDoc tags now that you have Flowtypes - don't want to have to keep track of redundant type annotations

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -53,6 +54,34 @@ const resolveSideEffects = () => ({
...
+	if (userAgent === 'mobile' || userAgent === 'desktop') {
+		return {
+			isAtSmallUp,
+			isAtMediumUp: userAgent !== 'mobile',
+			isAtLargeUp: userAgent === 'desktop'
+		};
+	}
+	// parses user agent string to determine if user is on mobile / tablet / desktop device
+	const device = new MobileDetect(userAgent);
+	const isMobile = device.mobile() !== null && device.phone() !== null;
```
mmcgahan: is this sufficient?
```js
const isMobile = Boolean(device.phone())
```

if `device.phone()` returns a value, then do you _also_ need to know that `device.mobile()` returns a value?

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -53,6 +54,34 @@ const resolveSideEffects = () => ({
...
+	const isTablet = device.tablet() !== null;
```
mmcgahan: nitpick, but `!== null` sort of reads like a double negative to me. IMO `Boolean(device.tablet())` is more clear

packages/mwp-app-render/src/components/connectedWithMatchMedia.jsx
jaclynj: yes it's to declare a type inline. I was using this as a reference: https://flow.org/en/docs/react/hoc/

mmcgahan: Interesting. I'm finding that doc a little hard to follow, but it looks like maybe the typedef for the result would be

```
ConnectedComponentClass<$Diff<Props, { initialMedia: MatchMedia }>, *>
```

and `ConnectedWithMatchMedia` would be implemented with something like `const ConnectedWithMatchMedia = (props: Props) => (` although I'm not 100% sure

willh-meetup:
Big picture, I think this looks good 👍 .

packages/mwp-app-render/src/components/connectWithMatchMedia.test.jsx
```diff
@@ -0,0 +1,41 @@
...
+	removeListener: jest.fn(),
+});
+
+describe('connectWithMatchMedia', () => {
+	const mockStore = createFakeStore({
+		config: { media: { isAtSmallUp: true } },
+	});
+	const connectWithMatchMedia = shallow(
+		<TestComponentConnectWithMatchMedia />,
+		{ context: { store: mockStore } }
```
mmcgahan: nice

packages/mwp-app-route-plugin/src/route.test.js
```diff
@@ -5,10 +5,14 @@ import getRoute, { onPreResponse } from './route';
...
 	it('returns html containing error message', () => {
+
 		const errorMessage = 'foobar';
 		const errorCode = 432;
+		const response = Boom.create(errorCode, errorMessage);
+		response.header = ((key,val) =>val)
+
 		const request = {
-			response: Boom.create(errorCode, errorMessage),
+			response: response,
```
mmcgahan: nitpick, use object shorthand `{ response, ... }`

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -212,6 +239,10 @@ const makeRenderer = (
 	const domain: string =
 		headers['x-forwarded-host'] || headers['x-meetup-host'] || info.host;
 	const host = `${requestProtocol}://${domain}`;
+	const userAgent =
+		process.env.NODE_ENV === 'production'
+			? headers['x-ua-device']
+			: headers['user-agent'];
```
mmcgahan: inline comment on these two lines would be very useful
```js
  ? headers['x-ua-device'] // set by fastly
  : headers['user-agent']; // fallback to the real UA
```

PR #424, opened 2017-12-27
Update peer dependencies for toaster

mperrotti:


packages/mwp-toaster/package.json
```diff
@@ -26,10 +26,10 @@
   "dependencies": {},
   "devDependencies": {},
   "peerDependencies": {
-    "meetup-web-components": "^3.0.1281",
+    "meetup-web-components": "^4.3.1787",
```
mmcgahan: I think `"^3 || ^4"` might avoid more pinned dependency inconsistencies, unless `mwp-toaster` is strictly _incompatible_ with MWC v3.x

PR #428, opened 2018-01-29
WP-629: update duotone home endpoint logic to support most_popular row

sadafie:
We're launching a guest homepage split test later this week, and we want to duotone the images that get returned in the `explore/home` api request as part of the most popular events nearby. 

These changes continue the use of type to identify the endpoint, however, since we no longer have "rows" of groups, i've replaced the existing logic with this new logic.

### DIFF
--- a/packages/mwp-api-proxy-plugin/src/util/duotone.js
+++ b/packages/mwp-api-proxy-plugin/src/util/duotone.js
@@ -169,13 +169,7 @@ export const apiResponseDuotoneSetter = duotoneUrls => {
 				groups.forEach(setGroupDuotone);
 				break;
 			case 'home':
-				(value.rows || [])
-					.map(({ items }) => items)
-					.forEach(items =>
-						items
-							.filter(({ type }) => type === 'group')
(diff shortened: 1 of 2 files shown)


packages/mwp-api-proxy-plugin/src/util/duotone.js
```diff
@@ -169,13 +169,7 @@ export const apiResponseDuotoneSetter = duotoneUrls => {
...
 				break;
 			case 'home':
-				(value.rows || [])
-					.map(({ items }) => items)
-					.forEach(items =>
-						items
-							.filter(({ type }) => type === 'group')
-							.forEach(({ group }) => setGroupDuotone(group))
-					);
+				(value.most_popular || []).map(event => setGroupDuotone(event.group || {}));
```
mmcgahan: nitpick, but semantically this should be `forEach` since you're just relying on side effects, not using the new array

sadafie: Thanks, will update!

PR #429, opened 2018-01-30
refactor withIntl to key TRN source using state.config.requestLangage

mmcgahan:
This PR enhances `withIntl` with Redux `connect` so that it can pass along the correct language's `messages` to `IntlProvider` from the new TRN source format defined in mwp-cli v5+

I've decided to call this a minor update since it doesn't actually require downstream code changes, just a MWP-CLI peer dependency update - it's sort of a breaking change, but sort of not, and `yarn check` will ensure that the consumer apps update to v5+ because of the update to `peerDependencies` in `mwp-core/package.json`


packages/mwp-i18n/src/withIntl.jsx
```diff
@@ -2,9 +2,18 @@
...
 import { IntlProvider, injectIntl } from 'react-intl';
+import { connect } from 'react-redux';
+import type { MapStateToProps } from 'react-redux';
 
-type Messages = { [string]: string };
-type Props = { __locale?: string, [string]: any };
+const DEFAULT_LOCALE = 'en-US';
+type TRNSource = { [string]: string };
+// Messages must support 'en-US'
+type Messages = { 'en-US': TRNSource, [string]: TRNSource };
```
eilinora: maybe i'm slightly confused about the changes, but this seems to always expect an object organized by locale of trn messages... 
```js
{ LOCALE: [... MESSAGES OBJECT...] }
```

mmcgahan: yes indeed - that's the new format produced by the mwp-cli PR https://github.com/meetup/mwp-cli/pull/49

packages/mwp-app-render/src/components/__snapshots__/connectWithMatchMedia.test.jsx.snap
```diff
@@ -1,60 +1,12 @@
...
-      "media": Object {
-        "isAtSmallUp": true,
-      },
-    },
-    "ref": null,
-    "rendered": null,
-    "type": [Function],
-  },
-  Symbol(enzyme.__nodes__): Array [
+<ConnectWithMatchMedia(TestComponent)
```
jaclynj: Why were all these other snapshots affected?

mmcgahan: This repo wasn't using the 'pretty' enzyme snapshot serializer that mup-web is using, so the snapshots were accurate but not very legible - `snapshotSerializer` added [here](https://github.com/meetup/meetup-web-platform/pull/429/files#diff-04e7f2718acefccd3077079b98af4b79R16)

packages/mwp-i18n/src/withIntl.jsx
eilinora: Ahhhhh! I missed that detail in the previous PR.

package.json
```diff
@@ -13,11 +13,12 @@
...
   },
   "scripts": {
-    "build": "babel src --out-dir lib --ignore test.js,test.jsx --source-maps",
+    "bootstrap": "lerna bootstrap",
+    "build": "lerna run build",
     "lint": "eslint . --ext .js,.jsx --fix",
     "precommit": "lint-staged",
     "typecheck": "flow",
-    "test": "lerna bootstrap && lerna run build && yarn test:unit",
+    "test": "yarn bootstrap && yarn build && yarn test:unit",
```
eilinora: Curious why you moved away from lerna here

packages/mwp-i18n/src/withIntl.jsx
```diff
@@ -25,11 +34,16 @@ export default (messages: Messages, doInjectIntl?: boolean) => (
 	}
 
 	const WithIntl = (props: Props) => {
-		const { __locale, ...other } = props;
+		const {
```
eilinora: Shouldn't requestedLanguage and __locale be the same?

eilinora: Wouldn't be good to understand why these would be different. :)

package.json
mmcgahan: lerna is a local dependency, so `yarn` can always access it - if you call lerna directly it has to be globally installed

packages/mwp-i18n/src/withIntl.jsx
mmcgahan: `__locale` is a little hack that can be used to test a `withIntl`-enhanced component without using the `renderIntl` helper - I think Niko used it awhile back but I'm not sure if it's still in use. In the actual application, the `locale` prop is set exclusively by the top-level `<IntlProvider>` - it uses the same source data as `requestLanguage`, but doesn't need to be set anywhere else.

I'll go ahead and document that more fully

eilinora: Ahhh yeah, i've been eliminating uses of `__locale` as we moved tests off to enzyme. There's only 1 component left using it which we can kill the usage of :D.

mmcgahan: in that case, i won't document and we can clean up eventually

PR #432, opened 2018-02-02
Add lang attribute to html tag

jmhammons:
Adds `lang` attribute to `<html />` tag as requested here: https://meetup.atlassian.net/browse/WP-550

### DIFF
--- a/packages/mwp-app-render/src/components/Dom.test.jsx
+++ b/packages/mwp-app-render/src/components/Dom.test.jsx
@@ -1,0 +1,41 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import Dom from './Dom';
+
+const MOCK_BASE_URL = 'https://www.mock-base-url.com';  
+
+const MOCK_PROPS = {
+	appMarkup: '<div>mock app markup</div>',
+	assetPublicPath: '/mock/asset/public/path',
+	baseUrl: MOCK_BASE_URL,
+	head: {
+		title: { toComponent: () => null },
+		meta: { toComponent: () => null },
+		link: { toComponent: () => null },
+		script: { toComponent: () => null },
+	},
+	initialState: {
+		config: {
+			apiUrl: 'https://www.mock-api-url.com',
+			baseUrl: MOCK_BASE_URL,
+			enableServiceWorker: true,
+			requestLanguage: 'fr-FR',
+			supportedLangs: ['en-US', 'de-De', 'es-ES'],
+			initialNow: '1517589465414',
+			variants: 'mock-variants',
+			entryPath: '/ny-tech',
+			media: 'mobile',
+		},
+	}, 
+	scripts: ['script.js'],
+	cssLinks: ['style.css'],
+};
+
+describe('Dom', function() {
+	const domComponent = shallow(
+		<Dom {...MOCK_PROPS} />
+	);
+	it('renders with mock props', function() {
+		expect(domComponent).toMatchSnapshot();
+	});
+});
--- a/packages/mwp-app-render/src/components/Dom.test.jsx
+++ b/packages/mwp-app-render/src/components/Dom.test.jsx
@@ -16,15 +16,7 @@ const MOCK_PROPS = {
 	},
 	initialState: {
 		config: {
-			apiUrl: 'https://www.mock-api-url.com',
-			baseUrl: MOCK_BASE_URL,
-			enableServiceWorker: true,
 			requestLanguage: 'fr-FR',
-			supportedLangs: ['en-US', 'de-De', 'es-ES'],
-			initialNow: '1517589465414',
-			variants: 'mock-variants',
-			entryPath: '/ny-tech',
-			media: 'mobile',
 		},
 	}, 
 	scripts: ['script.js'],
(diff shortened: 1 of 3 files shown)


packages/mwp-app-render/src/components/Dom.test.jsx
```diff
@@ -0,0 +1,41 @@
...
+			apiUrl: 'https://www.mock-api-url.com',
+			baseUrl: MOCK_BASE_URL,
+			enableServiceWorker: true,
+			requestLanguage: 'fr-FR',
+			supportedLangs: ['en-US', 'de-De', 'es-ES'],
+			initialNow: '1517589465414',
+			variants: 'mock-variants',
+			entryPath: '/ny-tech',
+			media: 'mobile',
+		},
```
mmcgahan: `initialState` can be an empty object

packages/mwp-app-render/src/components/Dom.test.jsx
```diff
@@ -0,0 +1,41 @@
...
+	}, 
+	scripts: ['script.js'],
+	cssLinks: ['style.css'],
+};
+
+describe('Dom', function() {
+	const domComponent = shallow(
```
mmcgahan: nitpick, but in general it's helpful to make unit tests (each `it()` call) as self-contained as possible - i.e. avoid relying on parent scope so that maintainers don't have to search around the file to determine what is actually being tested. You don't have to optimize for DRYness. In this case, prefer calling `shallow` inside the `it()` - you can even do it inline in the `expect()` call without losing much clarity

```js
expect(shallow(<Dom {...MOCK_PROPS} />)).toMatchSnapshot();
```

packages/mwp-app-render/src/components/__snapshots__/Dom.test.jsx.snap
```diff
@@ -0,0 +1,37 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Dom renders with mock props 1`] = `
+<html
+  lang="fr"
```
mmcgahan: hooray!

packages/mwp-app-render/src/components/Dom.test.jsx
jmhammons: Ah, I did want to have `requestLanguage` in there so we could see the `html` tag in the snapshot. Could I simplify this to be `initialState: { config: requestLanguage: 'fr-FR' } }`?

mmcgahan: ah right, yes

packages/mwp-app-render/src/components/Dom.test.jsx
jmhammons: First, I'm happy to change this!

---

As a more general matter, I'm wondering if this feedback (regarding being self-contained) is applicable to this nearby file as well?

https://github.com/meetup/meetup-web-platform/blob/master/packages/mwp-app-render/src/components/ServerApp.test.jsx

FWIW, I happened to take inspiration from this file before submitting. I know it's a nitpick to begin with and my test file isn't exactly the same (`MOCK_PROPS`, for instance). However, I think this might be an okay opportunity to say that I have experienced this before -- where a particular feedback in PR _feels_ somewhat at odds with pre-existing code.

On the one hand -- I don't see this as a big deal. It's trivial for me to make this change and move on. On the other, I wonder if this is worth mentioning in case we can make improvements somewhere along the line. For me, I think it would be great if newcomers could feel like they could reliably follow existing patterns from the codebase.

Anyway, just some unfiltered feedback that came to mind. Happy to change this.

packages/mwp-app-render/src/components/Dom.test.jsx
jmhammons: One more question -- do you think it's worth leaving this as-is so `window.APP_RUNTIME` in the snap is more complete?

packages/mwp-app-render/src/components/Dom.test.jsx
mmcgahan: Fair point. Given the size of our codebase and how quickly it grows, we won't ever be able to have 100% consistency on things like this - best practices evolve and we can't spend time refactoring old code to match, so although it's good to use existing code as a reference, it shouldn't ever be hallowed as an 'ideal'.

Usually when I preface with 'nitpick', it's because I'm going to make a non-blocking style/structure suggestion that you may also find value in - if you think it's _not_ useful, feel free to ignore it. It's mainly a way of sharing ideas about how to do things better, but in a subjective way.

The existing `ServerApp.test.jsx` test could definitely have been structured better, but it's not a stop-the-world sort of concern.

jmhammons: Yep yep @mmcgahan, I'm with you on all that. And I definitely don't mind the nitpicks, it's more that I wanted to take this opportunity to be reflective about something I've witnessed.

packages/mwp-app-render/src/components/Dom.test.jsx
mmcgahan: a lot of the `config` spec/content could change over time, but there's no need for this test to try to track all of it unless it has a direct impact on DOM.jsx

PR #433, opened 2018-02-02
Clean up API middleware signatures

mmcgahan:
instead of passing `routes` and `baseUrl` to `getApiMiddleware`, we can instead just pass a route-resolving function. This reduces the `mwp-router` dependency in `mwp-api-state` (although it doesn't eliminate it) and makes the middleware function signatures a little simpler. It also brings the power of the `routeResolver` function to `server-render`, which will be useful for WP-547

### DIFF
--- a/packages/mwp-api-state/src/index.js
+++ b/packages/mwp-api-state/src/index.js
@@ -27,11 +27,11 @@ export { api, app, DEFAULT_API_STATE } from './reducer';
  * order to render the application. We may want to write a server-specific
  * middleware that doesn't include the other epics if performance is an issue
  */
-export const getApiMiddleware = (routes, fetchQueriesFn, baseUrl) =>
+export const getApiMiddleware = (routeResolver, fetchQueriesFn) =>
 	createEpicMiddleware(
 		combineEpics(
 			getCacheEpic(),
-			getSyncEpic(routes, fetchQueriesFn, baseUrl),
+			getSyncEpic(routeResolver, fetchQueriesFn),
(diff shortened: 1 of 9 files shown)


packages/mwp-api-state/src/index.js
```diff
@@ -27,11 +27,11 @@ export { api, app, DEFAULT_API_STATE } from './reducer';
  * order to render the application. We may want to write a server-specific
  * middleware that doesn't include the other epics if performance is an issue
  */
-export const getApiMiddleware = (routes, fetchQueriesFn, baseUrl) =>
+export const getApiMiddleware = (resolveRoutes, fetchQueriesFn) =>
```
mmcgahan: this summarizes the effect of the refactor - instead of `routes` + `baseUrl`, the function just gets a `resolveRoutes` function

packages/mwp-api-state/src/sync/index.js
```diff
@@ -58,56 +58,53 @@ export function getDeprecatedSuccessPayload(successes, errors) {
  * @param {Object} routes The application's React Router routes
  * @returns {Function} an Epic function that emits an API_REQUEST action
  */
-export const getNavEpic = (routes, baseUrl) => {
-	const resolveRoutes = getRouteResolver(routes, baseUrl);
-	return (action, store) => {
```
mmcgahan: slight signature change, but mostly whitespace change - see https://github.com/meetup/meetup-web-platform/pull/433/files?w=1 to view without whitespace changes

PR #434, opened 2018-02-02
WP-547 send mu_api requests to location-based endpoint

mmcgahan:
https://meetup.atlassian.net/browse/WP-547

*Summary*

Instead of calling `/mu_api` for _every_ fetch, this PR updates the fetch logic to append a pathname that matches the currently-visible route. This allows our analytics tools to 'bucket' analytics data based on application route, rather than putting all request timing data into the same `/mu_api` URL. See the Jira ticket for more detail/implications #

### DIFF
--- a/packages/mwp-api-proxy-plugin/src/index.js
+++ b/packages/mwp-api-proxy-plugin/src/index.js
@@ -31,7 +31,7 @@ export default function register(
 	server.ext('onRequest', setPluginState);
 
 	// add a route that will receive query requests as querystring params
-	const routes = getApiProxyRoutes(API_ROUTE_PATH);
+	const routes = getApiProxyRoutes(`${API_ROUTE_PATH}{page*}`);
 	server.route(routes);
 
 	next();
--- a/packages/mwp-api-proxy-plugin/src/index.js
+++ b/packages/mwp-api-proxy-plugin/src/index.js
@@ -31,7 +31,7 @@ export default function register(
 	server.ext('onRequest', setPluginState);
 
 	// add a route that will receive query requests as querystring params
-	const routes = getApiProxyRoutes(`${API_ROUTE_PATH}{page*}`);
+	const routes = getApiProxyRoutes(`${API_ROUTE_PATH}/{page*}`);
 	server.route(routes);
 
 	next();
--- a/packages/mwp-api-state/src/sync/index.js
+++ b/packages/mwp-api-state/src/sync/index.js
@@ -208,7 +208,7 @@ export const getFetchQueriesEpic = (resolveRoutes, fetchQueriesFn) => {
 			.then(actions => [...actions, api.complete(queries)]);
 	};
 };
-export default (fetchQueriesFn, resolveRoutes) =>
+export default (resolveRoutes, fetchQueriesFn) =>
 	combineEpics(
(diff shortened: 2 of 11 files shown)


packages/mwp-api-state/src/sync/index.js
```diff
@@ -134,70 +133,84 @@ export const apiRequestToApiReq = action =>
  * - API_ERROR  // deprecated
  * - API_COMPLETE
  */
-export const getFetchQueriesEpic = fetchQueriesFn => {
-	let locationIndex = 0; // keep track of location changes
+export const getFetchQueriesEpic = (resolveRoutes, fetchQueriesFn) => {
+	// keep track of location changes - will first be set by SERVER_RENDER
+	let currentLocation = {};
```
mmcgahan: instead of keeping track of an incrementing integer, we keep track of a `location` object - this will update any time `LOCATION_CHANGE` is fired, and we can use the value with `resolveRoutes` to find the matching route, from which we can construct the `/mu_api/` path

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -62,17 +63,17 @@ const getMedia = (userAgent: string) => {
 	let isMobile = true;
 	let isTablet = false;
 	// In development, parse user agent string to determine media by device
-	if (process.env.NODE_ENV !== 'production') {
-		const device = new MobileDetect(userAgent);
-		isMobile = Boolean(device.phone());
-		isTablet = Boolean(device.tablet());
-	} else {
+	if (process.env.NODE_ENV === 'production') {
```
mmcgahan: just a linting fix - `if/else` should start with positive (non-negated) condition

packages/mwp-api-state/src/sync/index.js
```diff
@@ -133,74 +133,84 @@ export const apiRequestToApiReq = action =>
...
+		return resolveRoutes(location)
+			.then(matched => matched.pop().match.path.replace(/:/g, ''))
+			.then(apiPath => {
+				// construct the fetch call using match.path
+				const fetchUrl = `${config.apiUrl}${apiPath}`;
+				const fetchQueries = fetchQueriesFn(
+					fetchUrl,
+					(self || {}).value
+				);
+				return fetchQueries(queries, meta); // call fetch
```
mmcgahan: lines 165-175 have changed - the rest is just an indentation change

packages/mwp-api-proxy-plugin/src/index.js
```diff
@@ -31,7 +31,7 @@ export default function register(
 	server.ext('onRequest', setPluginState);
 
 	// add a route that will receive query requests as querystring params
-	const routes = getApiProxyRoutes(API_ROUTE_PATH);
+	const routes = getApiProxyRoutes(`${API_ROUTE_PATH}{page*}`);
```
mmcgahan: this lets the api proxy listen on _any_ route that starts with `/mu_api`

packages/mwp-core/src/renderers/server-render.jsx
jaclynj: maybe move the comment on line 65 down to the `else`?

PR #435, opened 2018-02-09
Wp 532 introduce seo head

jmhammons:
Moves `mup-web`'s `SEOHead` & `seoHelper` components to `mwp-app-render` (with some light refactoring).


packages/mwp-app-render/src/util/socialHelper.js
```diff
@@ -0,0 +1,40 @@
+export const getSocialLinks = locale => {
```
jmhammons: Not sure if this is the right spot for this. Could also use some docs.

packages/mwp-app-render/src/util/seoHelper.jsx
```diff
@@ -0,0 +1,331 @@
+import React from 'react';
```
jmhammons: This was copied with next to no refactoring, though `convertToLocalTime` was added.

packages/mwp-app-render/src/components/SEOHead.jsx
```diff
@@ -0,0 +1,114 @@
...
+	pageMeta,
+	pageTitle,
+	robots,
+	route,
+}) => {
+
+	const metaData = generateMetaData({
+		appPath: `meetup:/${route}`,
+		baseUrl,
+		description: pageDescription,
```
jmhammons: I'm not sure renaming `pageDescription` to `description` is beneficial. It might be nice to keep the same name  throughout and we can simplify the code to `{ appPath, baseUrl, pageDescription }`, etc.?

packages/mwp-app-render/src/util/seoHelper.jsx
mmcgahan: I like that idea

packages/mwp-app-render/src/util/seoHelper.jsx
```diff
@@ -0,0 +1,331 @@
...
+import locales from 'mwp-config/locales';
+
+import { getSocialLinks } from 'mwp-app-render/src/util/socialHelper';
+
+export const FB_APP_ID =
```
jmhammons: I could see this living somewhere else.

packages/mwp-app-render/src/util/seo/links.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Generates array of React.element's of <link />'s containing
+ * canonical + locale urls for the path provided
+ * @param  {String} baseUrl base url of the page (protocol + hostname)
+ * @param  {String} localeCode   locale of user
+ * @param  {String} route   redux'd route to the current page
+ * @return {Array}  array of React.element's
+ */
+export const generateCanonicalUrlMetaTags = (baseUrl, localeCode, route) => {
```
jmhammons: Probably going to rename this since we're producing `<link />` tags.

packages/mwp-app-render/src/util/seo/metaTags.jsx
```diff
@@ -0,0 +1,150 @@
...
+		.replace(/<(?:.|\n)*?>/gm, '')
+		.replace(/\r?\n|\r\t/g, '')
+		.substr(0, charLen);
+
+/**
+ * Generates array of just topic names based on array of topic objects passed in
+ * @param {Array} topics array of topics
+ * @return {Array} simplified array of topic names
+ */
+export const getTopicNames = topics => topics.map(topic => topic.name);
```
mmcgahan: not sure this is worth breaking out into a function, but i guess this just part of copy-paste?

packages/mwp-app-render/src/util/seo/metaTags.jsx
```diff
@@ -0,0 +1,150 @@
...
+		{ property: 'twitter:site', content: '@meetup' },
+		{ property: 'twitter:image', content: imageUrl },
+		{ property: 'twitter:description', content: desc },
+		{ property: 'twitter:app:id:iphone', content: TWITTER_APP_ID },
+		{ property: 'twitter:app:name:iphone', content: 'Meetup' },
+		{ property: 'twitter:app:id:ipad', content: TWITTER_APP_ID },
+		{ property: 'twitter:app:name:ipad', content: 'Meetup' },
+		{ property: 'twitter:app:id:googleplay', content: 'com.meetup' },
+		{ property: 'twitter:app:name:googleplay', content: 'Meetup' },
+	].filter(metaEntry => metaEntry !== undefined && metaEntry !== '');
```
mmcgahan: is this the same as `].filter(x => x);` ? You want any truthy value to pass?

packages/mwp-app-render/src/util/seo/metaTags.jsx
```diff
@@ -0,0 +1,150 @@
...
+ * @param {Number} options.lat lat of content on page (generally group or event location)
+ * @param {Number} options.lon lon of content on page (generally group or event location)
+ * @param {String} options.city city of content on page (generally group or event location)
+ * @param {String} options.state state of content on page (generally group or event location)
+ * @param {String} options.country country of content on page (generally group or event location)
+ * @return {Array} array of meta objects for use by Helmet
+ */
+export const generateGeoMetaData = ({ lat, lon, city, state, country }) => {
+	const place = [city, state, country]
+		.filter(k => k !== '' && k !== undefined)
```
mmcgahan: `.filter(x => x)`?

packages/mwp-app-render/src/util/seo/metaTags.jsx
```diff
@@ -0,0 +1,150 @@
...
+		.join(', ');
+	return [
+		lat && lon && { property: 'geo.position', content: `${lat};${lon}` }, // MUP coordinates
+		{ property: 'geo.placename', content: place },
+		{ property: 'geo.region', content: country },
+	].filter(metaEntry => metaEntry !== undefined && metaEntry !== '');
```
mmcgahan: `.filter(x => x)`?

packages/mwp-app-render/src/util/seo/metaTags.test.js
```diff
@@ -0,0 +1,168 @@
...
+		const alUrlMeta = metaData.filter(
+			obj => obj.content === expectedUrl && obj.property === 'al:web:url'
+		);
+		expect(ogUrlMeta).toHaveLength(1);
+		expect(alUrlMeta).toHaveLength(1);
+	});
+
+	describe('open graph meta data', () => {
+		it('should set og:site_name to "Meetup"', () => {
+			const ogSiteName = obj =>
```
mmcgahan: is this missing a `metaData.filter` call? the `expect(ogSiteName).toHaveLength(1);` is just checking the function, not the data.

packages/mwp-app-render/src/util/seo/metaTags.test.js
```diff
@@ -0,0 +1,168 @@
...
+	});
+	describe('twitter meta data', () => {
+		it('should set twitter:image', () => {
+			const twitterImage = metaData.filter(
+				obj =>
+					obj.content === MOCK_META.imageUrl && obj.property === 'twitter:image'
+			);
+			expect(twitterImage).toHaveLength(1);
+		});
+	});
```
mmcgahan: _all_ of these tests that use `metaData.filter` could be replaced with a single `expect(metaData).toMatchSnapshot()` test, which would be faster and more robust

packages/mwp-app-render/src/util/seo/metaTags.test.js
```diff
@@ -0,0 +1,168 @@
...
+		const lat = 'hello world';
+		const lon = 'hello world';
+		const processedObj = generateGeoMetaData({ lat, lon });
+		const filteredObj = processedObj.filter(
+			obj => obj.content === `${lat};${lon}` && obj.property === 'geo.position'
+		);
+		expect(filteredObj).toHaveLength(1);
+	});
+
+	it('should NOT geo position', () => {
```
mmcgahan: usually good to specify the conditions under which a certain output is expected unless the output is _always_ expected, e.g. the test above could be `'should set position when provided lat/lon'` and this one could be `'should NOT set geo position when NOT provided lat/lon'` - otherwise the test output is ambiguous

packages/mwp-app-render/src/components/SEOHead.jsx
mmcgahan: yes

packages/mwp-app-render/src/components/SEOHead.jsx
```diff
@@ -0,0 +1,114 @@
...
+	return (
+		<Helmet defaultTitle="Meetup" titleTemplate="%s | Meetup">
+			<title>{pageTitle}</title>
+			<link rel="image_src" href={imageUrl} />
+
+			{robots ? (
+				<meta name="robots" content="index, follow" />
+			) : (
+				<meta name="robots" content="noindex, nofollow" />
+			)}
```
mmcgahan: nitpick, but I think I would make this a little DRYer with

```js
<meta name="robots" content={robots ? 'index, follow' : 'noindex, nofollow'} />
```

packages/mwp-app-render/src/util/socialHelper.js
emilyeng: looks like this exists in meetup-web-components [here](https://github.com/meetup/meetup-web-components/blob/master/src/utils/getSocialLinks.js)

packages/mwp-app-render/src/components/SEOHead.jsx
```diff
@@ -0,0 +1,114 @@
...
+
+	const ldJsonTags = ldJson.map((jsonObj, index) => (
+		// eslint-disable-next-line react/no-array-index-key
+		<script type="application/ld+json" key={`ldjson-${index}`}>
+			{JSON.stringify(jsonObj)}
+		</script>
+	));
+
+	return (
+		<Helmet defaultTitle="Meetup" titleTemplate="%s | Meetup">
```
emilyeng: might be nice to make this flexible so that it could be either `Meetup` or `Meetup Pro`

packages/mwp-app-render/src/util/seo/metaTags.jsx
```diff
@@ -0,0 +1,150 @@
...
+	localeCode,
+}) => {
+	const desc = stripAndTruncate(description);
+	const ogDesc = stripAndTruncate(ogDescription);
+
+	return [
+		{ name: 'description', content: desc },
+		{ name: 'keywords', content: keywords },
+		{ property: 'fb:app_id', content: FB_APP_ID },
+		{ property: 'og:site_name', content: 'Meetup' },
```
emilyeng: maybe also here -- `Meetup` or `Meetup Pro`

packages/mwp-app-render/src/util/seo/metaTags.jsx
jmhammons: Yeah, part of the copy / paste. Happy to change though.

packages/mwp-app-render/src/util/seo/ldJson.jsx
```diff
@@ -0,0 +1,134 @@
...
+};
+
+/**
+ * Generated Object of LD-JSON if event has a fee associated
+ * @see https://json-ld.org/ - to learn more about ld-json
+ * @param {Object} fee Fee data associated with event
+ * @return {Object} accumulated json object
+ */
+export const generateFeeLdJson = fee => {
+	return {
```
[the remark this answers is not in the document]
jmhammons: Yeah, totally.

packages/mwp-app-render/src/util/socialHelper.js
jmhammons: Ah! It's also in `mup-web` here - https://github.com/meetup/mup-web/blob/master/src/util/socialHelper.js

I think I'll get a ticket in to deal with this.

packages/mwp-app-render/src/util/seo/metaTags.jsx
jmhammons: Pretty sure I didn't write this bit but I think its aim is to not render any 'empty meta tags like - `<meta property="description" content="" />`. So I think what you're suggesting will work. I'll test.

packages/mwp-app-render/src/util/seo/ldJson.test.js
```diff
@@ -0,0 +1,82 @@
...
+			MOCK_VENUE,
+			MOCK_FEE,
+		};
+		const processedObj = generateEventLdJson(event);
+		expect(processedObj).toMatchSnapshot();
+	});
+});
+
+describe('generateOrganizationLdJson', () => {
+	it('should generate json with all the correct fields', () => {
```
jmhammons: I may do snapshots for this instead.

packages/mwp-app-render/src/util/seo/ldJson.jsx
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Generated Object of LD-JSON used for search engines to read data on page
+ * @see https://json-ld.org/ - to learn more about ld-json
+ * @param {Object} eventInfo event object for the page
+ * @return {Object} accumulated json object
+ */
+export const generateEventLdJson = eventInfo => {
+	const startDate =
+		eventInfo.time &&
+		convertToLocalTime(eventInfo.time, eventInfo.utc_offset).toISOString();
```
jmhammons: TODO: Need apply final fix for issue described here -- https://meetup.atlassian.net/browse/MW-3284

packages/mwp-app-render/src/components/seoHead.test.jsx
```diff
@@ -0,0 +1,63 @@
...
+		});
+		it('turns robots "off"', () => {
+			const props = {...MOCK_PROPS, robots: false};
+			expect(renderComponent(props).find(robotsSelector)).toMatchSnapshot();
+		});
+	});
+	describe('Custom Open Graph data', () => {
+		const ogDescriptionSelector = 'meta[property="og:description"]';
+		it('renders pageDescription as open graph description', () => {
+			expect(renderComponent().find(ogDescriptionSelector)).toMatchSnapshot();
```
mmcgahan: this test is already part of the full default snapshot

packages/mwp-app-render/src/components/seoHead.test.jsx
```diff
@@ -0,0 +1,63 @@
...
+		});
+		it('renders custom open graph description', () => {
+			const props = {...MOCK_PROPS, ogDescription: 'mock custom open graph description' };
+			expect(renderComponent(props).find(ogDescriptionSelector)).toMatchSnapshot();
+		});
+	});
+	describe('imageUrl', () => {
+		it('applies provided imageUrl to og:image, twitter:image and img_src tags', () => {
+			expect(renderComponent()).toMatchSnapshot();
```
mmcgahan: this is the same as the full snapshot test

packages/mwp-app-render/src/util/seo/metaTags.test.js
```diff
@@ -0,0 +1,71 @@
...
+	};
+	it('matches snap', () => {
+		const geoMetaData = generateGeoMetaData(MOCK_GEO);
+		expect(geoMetaData).toMatchSnapshot();
+	});
+	it('should *not* set geo.position when lat or lon are missing', () => {
+		const mockGeo = { ...MOCK_GEO };
+		delete mockGeo.lat;
+		const geoMetaData = generateGeoMetaData(mockGeo);
+		expect(geoMetaData).toMatchSnapshot();
```
mmcgahan: since this is a test for the existence of a specific property, a snapshot is probably not the best tool. You can try this to get a very specific failure message:

```js
expect(geoMetaData).not.toContainEqual(expect.objectContaining({ property: "geo.position" }))
```

Jest has a lot of great tools that ensure your tests are as specific as necessary and provide very readable error messages when you use the right matchers: https://facebook.github.io/jest/docs/en/expect.html

packages/mwp-app-render/src/components/seoHead.test.jsx
jmhammons: That's true, but I feel like there's a useful distinction, in that this explicitly calls this out with a snap that only looks at `og:description`. And where the 'full default snapshot' feels like it just snaps if things simply render, this delves specifically into its content a little more.

mmcgahan: it's useful to think about what the test would show if something changed unexpectedly - i.e. someone made an update that removed the `og:description` the snapshot test failure would highlight _just that markup_ (the diff would be specific), and the developer would have to figure out why it changed, knowing that it was an expected part of the original snapshot. That is nearly the same behavior as this test, only you get a more specific test description.

You should definitely think of snapshot as more than just "did this render", but "did the rendered object look exactly like this" so that you don't have to write tests for nested markup. By analogy, you currently have a check for `if (a && b)` and then you're checking for `if (a)` inside that block in order to be more explicit, which you generally wouldn't do.

jmhammons:
Hey @mmcgahan -- how is this looking?

PR #436, opened 2018-02-09
add datadog metric logging

burrata:
taken from node-service after https://github.com/meetup/node-service/pull/23.


eilinora:
@burrata any reason we havent merged?

PR #441, opened 2018-03-01
Wp 649 ldjson event dates

jmhammons:
Attempts to port recent changes in `mup-web`'s seo tools.


packages/mwp-app-render/src/util/dateTimeUtils.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Pads numbers less than two digits with an additional
+ * leading zero. Used for date and time formatting.
+ * @param {Number} num  Number to be padded
+ * @returns {String}    Returns a padded string
+ */
+export const padNumber = num => (num < 10 ? `0${num}` : `${num}`);
```
amandakievet: I'm wondering if we should create a story to change the `mup-web` codebase to use these functions from the platform and remove the duplicates there?

packages/mwp-app-render/src/util/dateTimeUtils.js
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Returns a formatted time string from the local value of the datetime
+ * object passed in for usage with the `find/upcoming_events` endpoint.
+ * The formatted date must follow the format: `HH:MM:SS`. Javascript's
+ * `.toISOString` method returns the iso value while we need the local value.
+ * @param {Object} timeObj  Datetime object
+ * @returns {String}        Formatted local date string in format `HH:MM:SS`
+ */
+export const getFormattedTime = timeObj =>
+        timeObj.toLocaleTimeString('UTC', { hour12: false });
```
amandakievet: a nitpick, but since I'm new -- it looks like this is a lot of space for the indent. Is this the convention for the platform codebase?

packages/mwp-app-render/src/util/dateTimeUtils.js
jmhammons: Yeah, in fact I need to do a similar exercise with `getSocialLinks()` that's like in every repo we have now.

packages/mwp-app-render/src/util/dateTimeUtils.js
jmhammons: Ah, I think this is just my bad.

packages/mwp-app-render/src/util/dateTimeUtils.test.js
```diff
@@ -0,0 +1,44 @@
...
+const runtimeOffset = runtimeDate.getTimezoneOffset() * 60 * 1000; // runtime-dependent millisecond offset
+
+describe('padNumber', () => {
+	it('should pad numbers less than 10', () => {
+		for (let i = 0; i < 10; i += 1) {
+			expect(padNumber(i)).toBe(`0${i}`);
+		}
+	});
+	it('should not pad numbers greater or equal to 10', () => {
+		for (let i = 10; i < 60; i += 1) {
```
mmcgahan: I think the range of 10-60 wouldn't be as meaningful as `10, 100, 1000, 10000` because then you're testing the string formatting of longer strings - I can't think of reasonable implementations that would handle 11 differently than 12, but I can definitely think of implementations that would fail to account for 3-digit numbers even if it handled 2-digit numbers beautifully

```js
for (let i = 10; i < 10001; i *= 10) {
```

or a little more explicit:

```js
for (pow = 1; pow < 5; pow += 1) {
  const test = Math.pow(10, pow);
  expect(padNumber(test).toBe(test.toString())
}
```

packages/mwp-app-render/src/util/dateTimeUtils.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Returns a Date object that offsets the supplied UTC time, correcting for the runtime environment's current UTC offset.
```
mmcgahan: Given how confusing this behavior of this function is and the specific caveats for how it should be used, I feel like it needs a much more comprehensive description - what use cases does it apply to? Why would it be useful to manually apply 2 timezone shifts on a Date? How is daylight savings time accounted for?

Without docs, the best case scenario is that developers spend a lot of time discussing the implementation each time it is used - worst case, it is used in 'inappropriate' contexts and bugs appear.

packages/mwp-app-render/src/util/dateTimeUtils.test.js
```diff
@@ -0,0 +1,44 @@
...
+			expect(padNumber(i)).toBe(String(i));
+		}
+	});
+});
+
+describe('getFormattedTime', () => {
+	it('should return a local time string in the format: `HH:MM:SS`', () => {
+		// Regex for formatted string `HH:MM:SS`
+		const regExp = new RegExp(/\d{2}:\d{2}:\d{2}/);
```
mmcgahan: no need to call `new RegExp` when using a regex literal - `new RegExp` is generally only needed when you have to construct a regex from a string representation. This is equivalent to `const regExp = /\d{2}:\d{2}:\d{2}/;`

packages/mwp-app-render/src/util/dateTimeUtils.test.js
```diff
@@ -0,0 +1,44 @@
...
+		expect(getFormattedTime(date)).toMatch(regExp);
+	});
+});
+
+describe('convertToLocalTime', () => {
+	it('should return a date with UTC offset by eventOffset _and_ runtimeOffset', () => {
+		const expectedTime = timeUTC + eventOffset + runtimeOffset;
+		expect(convertToLocalTime(timeUTC, eventOffset).getTime()).toBe(expectedTime);
+	});
+	it('should return expected date value based on time is provided with no offset', () => {
```
mmcgahan: Could you rephrase? or maybe it's just a typo and `is` should be `if`

packages/mwp-app-render/src/util/seo/__snapshots__/ldJson.test.js.snap
```diff
@@ -7,6 +7,7 @@ Object {
   "description": "The coolest event in the world during which we will run and dance and sing
 	<script>alert(\\"bad time\\")</script>, 😊, &lt;blink&gt;what what&lt;blink&gt; this is getting
 	longer than it needs to be why am I still typing omg",
+  "endDate": "1970-01-18T10:01+00:00",
```
mmcgahan: is this the right time? I think the event object mock used seconds rather than milliseconds, but we should find out what events _usually_ provide and make sure the test matches up with typical usage

jmhammons:
I'll have a swing at this feedback @mmcgahan. All this code was not authored by me originally and was copied from `mup-web`, so bear with...

mmcgahan:
@jmhammons sounds good - as soon as it's in this repo, we own it for the next decade, so it's important to get it as 'right' as possible

packages/mwp-app-render/src/util/dateTimeUtils.js
amandakievet: @jmhammons I tried a stab at some of this explanation here, if this helps at all: https://meetup.atlassian.net/wiki/spaces/INTL/pages/353338155/Engineering+SEO

@mmcgahan I think you have about the best understanding of anyone I've talked to about what this does

mmcgahan: maybe we should start a 'how to work with dates' document somewhere...

I'm fine for this code comment to be a confluence link, (although we only have weak guarantees of keeping those URLs consistent), so if anyone wants to start a page like that in the Web Eng Collective space, that would be great, but I'd like to avoid the necessarily-wandering discussions about when/how to use this in the future :)

packages/mwp-app-render/src/util/seo/__snapshots__/ldJson.test.js.snap
jmhammons: Yeah, my understanding here is `MOCK_EVENT` has a `0` `utc_offset` but we can mock this with a valid `utc_offset` in milliseconds.

packages/mwp-app-render/src/util/dateTimeUtils.js
```diff
@@ -0,0 +1,38 @@
...
+ * Returns a formatted time string from the local value of the datetime
+ * object passed in for usage with the `find/upcoming_events` endpoint.
+ * The formatted date must follow the format: `HH:MM:SS`. Javascript's
+ * `.toISOString` method returns the iso value while we need the local value.
+ */
+export const getFormattedTime = (date: Date): string =>
+	date.toLocaleTimeString('UTC', { hour12: false });
+
+/**
+ * Shifts a supplied time to match time of another timezone
```
mmcgahan: I recommend changing 'another timezone' to 'timezone indicated by offset'

PR #442, opened 2018-03-06
add loaders and config for scss style modules

akdetrick:
Adds configuration and loader deps for handling `component.module.scss` files.

```
    component.module.scss[1]                            Component.jsx
    +---------+---------+                               +-----+------+
              |                                               ^
+-------------+                                               |
|                                                             |
|    +----------------+      +------------------+      +------+-------+       +--------------+
|    |                |      |                  |      |              |       |              |
+--> |  sass-loader   +----> |  postcss-loader  +----> |  css-loader  +-----> | simple-universal-style-loader[2] |
     |                |      |                  |      |              |       |              |
     +----------------+      +------------------+      +--------------+       +--------------+

     precompile Sass         1) CSSNext                creates importable     injects CSS into
                             2) custom properties      style module           the document

```

[1] `component.module.scss` files will live next to `Component.jsx` files in the `src` dir, like test files
[2] `universal-style-loader` differs from `style-loader` only in that it can inject styles into the document on the server side by collecting styles in node's `global` object under the `__styles__` key.


akdetrick:
Marking "prelim" to investigate populating CSS custom properties as a loader option rather than manually importing them in app code.

akdetrick:
Marking "Prelim" to populate CSS custom properties in this config (rather than running a separate CSS file through these loaders).

akdetrick:
Loading custom properties into the loader is complete and working, but leaving this as `Prelim` so we can attempt to get server side style rendering working before merging this.

packages/mwp-app-render/src/components/Dom.jsx
```diff
@@ -57,6 +57,24 @@ const DOM = props => {
 		escapedState,
 	};
 
+	// simple-universal-style-loader accumulates CSS Module style rules
+	// in node's `global` object under the `__styles__` key.
+	//
+	// If this key is populated, we render a single style tag with all
+	// module styles needed for the request.
+	let moduleStyleTag;
+	if ( global.__styles__ ) {
```
mmcgahan: because referencing `global` is dirty, let's use `getStyles` https://github.com/janoist1/simple-universal-style-loader#server-environment

packages/mwp-app-render/src/components/Dom.jsx
```diff
@@ -57,6 +57,24 @@ const DOM = props => {
 		escapedState,
 	};
 
+	// simple-universal-style-loader accumulates CSS Module style rules
+	// in node's `global` object under the `__styles__` key.
+	//
+	// If this key is populated, we render a single style tag with all
+	// module styles needed for the request.
+	let moduleStyleTag;
```
mmcgahan: even better, define a function called `getModuleStyleTags` outside of the `Dom` class to avoid the `let` and encapsulate this behavior - I would potentially even break this out into a standalone utility module, but that's probably overkill in terms of separating concerns.

Another alternative is to follow the same pattern that `cssLinks` is using and pass in an array of `style` strings as a `prop` to `Dom` that can be mapped to `<style>` tags internally - that might be nice to separate the _ability_ to inject `style` tags from the specific way that `simple-universal-style-loader` _defines_ the styles - make it the responsibility of the function that calls `<Dom>` rather than `Dom`'s direct responsibility.

```jsx
const { styles = [] } = this.props;

...
<head>
  {styles.map((body, key) => <style type="text/css" key={key}>body</style>}
</head>
```

```jsx
const { getStyles } = require('simple-universal-style-loader');
...
<Dom styles=`getStyles()` ... />
```

packages/mwp-core/src/util/inlineStyle.js
```diff
@@ -0,0 +1,23 @@
...
+import { getStyles } from 'simple-universal-style-loader';
+
+/**
+ * Returns inline style tags from CSS accumulated by
+ * `simple-universal-style-loader`. This allows us to
+ * render styles from CSS modules on the server.
+ *
+ * @returns {Array} - array of style tags
+ */
+export function getInlineStyleTags(): array {
```
sadafie: Probably needs a corresponding unit test

sadafie:
@mmcgahan PRs like this that touch both mwp-core and mwp-app-render make me wonder why they're not a single package? I'm sure there a clear nuances for the separation, but i'm not remembering what those are.

mmcgahan:
mwp-core was sort of a catch-all for things that i didn't have time to better organize when I was separating out packages - at that time `mwp-app-render` was mostly just React components. You could argue that almost all of the code in this PR could be contained in `mwp-core` except for the import into PageWrap, but really we just need to make another pass at how we want to organize things - `mwp-core` was never supposed to last for more than a couple months, but... priorities.

Organizing more code into a more general 'rendering' package that would contain all changes in this PR makes sense to me.

packages/mwp-core/src/util/inlineStyle.js
mmcgahan: Not sure what Flow says about the `: array` return type but I _think_ it's recommended to use `Array<type>` to specify the type of the elements an array contains, e.g. `Array<React$Element>` or something for this (not 100% confident of the React type, but it should be easy to confirm).

That will also show you that unless you know the return type of `getStyles`, you could be returning a falsy non-array value rather than guaranteeing an Array return type.

packages/mwp-core/src/util/inlineStyle.test.js
```diff
@@ -0,0 +1,22 @@
+import { getInlineStyleTags } from './inlineStyle.js';
+
+// `simple-universal-style-loader` reades from `global`,
+// so we must mock the data directly in `global.__styles__`
+global.__styles__ = [
```
mmcgahan: might be nicer to do this before the `it` and then clear it with `global.__styles__ = null` after the test so you don't leak

packages/mwp-core/src/util/inlineStyle.test.js
```diff
@@ -0,0 +1,22 @@
...
+			}
+		]
+	}
+];
+
+describe('getInlineStyleTags', () => {
+	it('matches snapshot', () => {
+		const result = getInlineStyleTags()[0];
+		expect(result).toMatchSnapshot();
+	});
```
mmcgahan: Probably good to test that it works when there are _no_ global styles?

packages/mwp-core/src/util/inlineStyle.js
akdetrick: It _seems_ like the right annotation is `: Array<React.Element<'style'>>`, but typecheck is failing:

```
Cannot get React.Element because property Element is missing in object
type [1].
```

fwiw, the docs only mention a [primitive annotation for function returns](https://flow.org/en/docs/types/functions/#toc-function-returns).

mmcgahan: I suspect a Flow version thing - I think the syntax might actually be `React$Element<'style'>` or `React$Element`

akdetrick: Success. Pushing the updated type annotation; otherwise this looks ready for a final pass.

PR #444, opened 2018-03-09
GoogleTagManager component

jmhammons:
Draft version of what a bespoke `<GoogleTagManager />` comp could look like.

https://developers.google.com/tag-manager/quickstart

The basic idea.
```
<head>
    <!-- Part 1: GTM script (because Google wants this close to the top of the head tag) -->
    <GoogleTagManagerScript />
</head>
<body>
    <!-- Part 2: GTM noscript (because Google wants this right after opening body tag) -->
    <GoogleTagManagerNoscript />
    <div>more app html</div>
</body>
```

To do / discuss:
- ~~Integration touch point.~~ Spoke to Mike M. and we think `Dom.jsx`
- What about a package? (Pro web used this)
-- https://github.com/holidaycheck/react-google-tag-manager
- Performance concerns. 
-- _Note:_ this draft version implements GTM as per Google's suggestions. However, we should probably discuss if we can deviate.
- Supporting data:
-- RTF: I think we can pass a flag to `Dom` from `server-render`?
- `dataLayerName` -- not sure if this is dynamic

### DIFF
--- a/packages/mwp-app-render/src/components/GoogleTagManager.jsx
+++ b/packages/mwp-app-render/src/components/GoogleTagManager.jsx
@@ -1,0 +1,36 @@
+// @flow
+
+import React from 'react';
+import PropTypes from 'prop-types';
+
+// Production and dev keys
+const GTM_KEY = process.env.NODE_ENV === 'production' ? 'GTM-T2LNGD' : 'GTM-W9W847';
+
+/*
+ * @description Gets google tag manager script tag
+ * @see {@link https://developers.google.com/tag-manager/quickstart}
+*/
+export const GoogleTagManagerScript = (key : string = GTM_KEY) => {
+	const script = `
+		(function(w,d,s,l,i){
+			w[l]=w[l]||[];
+			w[l].push({'gtm.start':new Date().getTime(),event:'gtm.js'});
+			var f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';
+			j.async=true;j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;
+			f.parentNode.insertBefore(j,f);
+		})(window,document,'script','dataLayer','${GTM_KEY}');`
+	return <script dangerouslySetInnerHtml={{ __html: script }}></script>
+};
+
+/*
+ * @description Gets google tag manager noscript tag
+ * @see {@link https://developers.google.com/tag-manager/quickstart}
+*/
+export const GoogleTagManagerNoscript = (key : string  = GTM_KEY) => {
+	const iframe = `
+		<iframe
+			src="https://www.googletagmanager.com/ns.html?id=${GTM_KEY}"
+			height="0" width="0" style="display:none;visibility:hidden">
+		</iframe>`;
+	return <noscript dangerouslySetInnerHtml={{ __html: iframe }}></noscript>
+};
--- a/packages/mwp-app-render/src/components/GoogleTagManager.test.jsx
+++ b/packages/mwp-app-render/src/components/GoogleTagManager.test.jsx
@@ -1,0 +1,15 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { GoogleTagManagerScript, GoogleTagManagerNoscript } from './GoogleTagManager';
+
+describe('<GoogleTagManagerScript />', () => {
+	it('matches snap', () => {
+		expect(shallow(<GoogleTagManagerScript />)).toMatchSnapshot();
+	});
+});
+
+describe('<GoogleTagManagerNoscript />', () => {
+	it('matches snap', () => {
+		expect(shallow(<GoogleTagManagerNoscript />)).toMatchSnapshot();
+	});
+});
--- a/packages/mwp-app-render/src/components/__snapshots__/GoogleTagManager.test.jsx.snap
+++ b/packages/mwp-app-render/src/components/__snapshots__/GoogleTagManager.test.jsx.snap
@@ -1,0 +1,32 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`<GoogleTagManagerNoscript /> matches snap 1`] = `
+<noscript
+  dangerouslySetInnerHtml={
+    Object {
+      "__html": "
+    		<iframe
+    			src=\\"https://www.googletagmanager.com/ns.html?id=GTM-W9W847\\"
+    			height=\\"0\\" width=\\"0\\" style=\\"display:none;visibility:hidden\\">
+    		</iframe>",
+    }
+  }
+/>
+`;
+
+exports[`<GoogleTagManagerScript /> matches snap 1`] = `
+<script
+  dangerouslySetInnerHtml={
+    Object {
+      "__html": "
+    		(function(w,d,s,l,i){
+    			w[l]=w[l]||[];
+    			w[l].push({'gtm.start':new Date().getTime(),event:'gtm.js'});
+    			var f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';
+    			j.async=true;j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;
+    			f.parentNode.insertBefore(j,f);
+    		})(window,document,'script','dataLayer','GTM-W9W847');",
+    }
+  }
+/>
+`;


packages/mwp-app-render/src/components/GoogleTagManager.jsx
```diff
@@ -0,0 +1,70 @@
...
+ * <head>
+ * 	<!-- GTM script -->
+ * 	<GoogleTagManager item="script" rtf={true} isProd={true} />
+ * </head>
+ * <body>
+ * 	<!-- GTM noscript -->
+ * 	<GoogleTagManager item="noscript" rtf={true} isProd={true} />
+ * </body>
+*/
+const GoogleTagManager = ({item, rtf = false, isProd = false}) => {
```
jmhammons: Fwiw, I'm not actually sure how I'm going to get `rtf` and `isProd` data here yet. Maybe it's wishful thinking.

packages/mwp-app-render/src/components/GoogleTagManager.jsx
```diff
@@ -0,0 +1,70 @@
...
+const getNoscript = (key: string) => { 
+	const src = `https://www.googletagmanager.com/ns.html?id=${key}`;
+	return <noscript><iframe src={src} height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+};
+
+/*
+ * @description Gets google tag manager content (script or noscript)
+ * @param {String} 'script' or 'noscript', indicating which GTM chunk to return
+ * @param {Boolean} rtf Whether or not the runtime flag is on
+ * @param {Boolean} isProd Whether or not to production environment
```
jmhammons: This makes no sense.

packages/mwp-app-render/src/components/GoogleTagManager.jsx
```diff
@@ -0,0 +1,76 @@
...
+	if(tag === 'noscript') {
+		return getNoscript(key);
+	}
+
+};
+
+GoogleTagManager.propTypes = {
+	rtf: PropTypes.bool,
+	isProd: PropTypes.bool,
+	tag: PropTypes.string.isRequired,
```
mmcgahan:
```js
PropTypes.oneOf(['script', 'noscript']),
```

packages/mwp-app-render/src/components/GoogleTagManager.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+	if(tag === 'noscript') {
+		return getNoscript(key);
+	}
+
+};
+
+GoogleTagManager.propTypes = {
+	rtf: PropTypes.bool,
+	isProd: PropTypes.bool,
```
mmcgahan: you can check for prod env statically with `const isProd = process.env.NODE_ENV === 'production'` - then you don't have to pass it in as a prop, and you can assign `key` when the module is first loaded

packages/mwp-app-render/src/components/GoogleTagManager.jsx
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	const key = GTM_KEYS[isProd ? 'prod' : 'dev'];
+
+	if(tag === 'script') {
+		return getScript(key);
+	}
+
+	if(tag === 'noscript') {
+		return getNoscript(key);
```
mmcgahan: why not export 2 components - one noscript, one regular? Doesn't look like there's any other need for the prop.

packages/mwp-app-render/src/components/GoogleTagManager.jsx
```diff
@@ -0,0 +1,76 @@
...
+ * <body>
+ * 	<!-- GTM noscript -->
+ * 	<GoogleTagManager tag="noscript" rtf={true} isProd={true} />
+ * </body>
+*/
+const GoogleTagManager = ({tag, rtf = false, isProd = false}) => {
+
+	const validTagProp = ['script','noscript'].indexOf(tag) >= 0;
+	
+	if(!validTagProp || !rtf) {
```
mmcgahan: why pass in `rtf`? Since the parent component is responsible for determining the RTF state, it seems reasonable to let it also decide what to do with it? i.e. don't try to render `<GoogleTagManager />` at all if the rtf if false?

packages/mwp-app-render/src/components/GoogleTagManager.jsx
```diff
@@ -0,0 +1,76 @@
...
+	'dev': 'GTM-W9W847',
+};
+
+/*
+ * @description Gets google tag manager script with key applied
+ * @see {@link https://developers.google.com/tag-manager/quickstart}
+*/
+const getScript = (key: string) => {
+	const script = `(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);})(window,document,'script','dataLayer','${key}');` 
+	return <script>{script}</script>
```
mmcgahan: 2 things to watch out for:

1. This might need to be `<script dangerouslySetInnerHtml={ { _html: script } } />` in order to avoid HTML escaping the contents
2. I like the line breaks that are used in https://github.com/holidaycheck/react-google-tag-manager/blob/master/src/build_parts.js - even though it's minified, you can at least see the whole script without side scrolling

packages/mwp-app-render/src/components/GoogleTagManager.jsx
jmhammons: Yeah, I had considered this but opted to pass it in, because it seemed I could more easy create 'dev' and 'prod' snapshots in `GoogleTagManager.test.jsx` this was way.

I was planning to do, in `Dom.jsx`...
```
const isProd = process.env.NODE_ENV === 'production';
<GoogleTagManager isProd={isProd} />
```

If I do this check in `<GoogleTagManager />` can I keep the 'dev' and 'prod' snaps?

packages/mwp-app-render/src/components/GoogleTagManager.jsx
jmhammons: Yeah, I had considered both approaches.

I like something like `{gtmRtf && <GoogleTagManager />}` as well, because it doesn’t try to render the component if the flag is `false`.

But my thinking behind passing `rtf` instead was something like…
- Keeps the runtime flag controlled rendering behavior with the component
- Testable in `GoogleTagManager.test.jsx`
- Hopefully aids in making it clear that a rtf exists and is necessary to render, just by looking <GoogleTagManager /> docs
- It allows the component to not render by default (which I preferred in this case since it’s a 3rd party script) and requires the implementer to explicitly pass a `true` rtf to get it to render. Otherwise, though probably not likely, checking the runtime flag could be more easily forgotten.

packages/mwp-app-render/src/components/GoogleTagManager.jsx
jmhammons: Oh, awesome.

packages/mwp-app-render/src/components/GoogleTagManager.jsx
jmhammons: I think mostly because of the `key` and `rtf` overlap that would exist between the two. And this setup also just came to me and made sense.

packages/mwp-app-render/src/components/GoogleTagManager.jsx
jmhammons: re: 1. Gotcha. Thanks.
re: 2. Yeah, I also like that. Will update.

packages/mwp-app-render/src/components/GoogleTagManager.jsx
mmcgahan:
```js
const GTM_KEY = process.env.NODE_ENV === 'production' ? 'foo': 'bar'
const GoogleTagManager = ({tag, rtf = false, key = GTM_KEY}) => {
  ...
}
```

packages/mwp-app-render/src/components/GoogleTagManager.jsx
```diff
@@ -0,0 +1,87 @@
...
+ * <head>
+ * 	<!-- GTM script -->
+ * 	<GoogleTagManager tag="script" rtf={true} isProd={true} />
+ * </head>
+ * <body>
+ * 	<!-- GTM noscript -->
+ * 	<GoogleTagManager tag="noscript" rtf={true} isProd={true} />
+ * </body>
+*/
+const GoogleTagManager = ({tag, rtf = false, isProd = false}) => {
```
mmcgahan: Flow will probably complain that you haven't added type defs to this - you might want to add https://github.com/w0rp/ale as a vim plugin to get line-by-line Flow errors while you edit

packages/mwp-app-render/src/components/GoogleTagManager.jsx
```diff
@@ -0,0 +1,44 @@
...
+		<iframe
+			src="https://www.googletagmanager.com/ns.html?id=${GTM_KEY}"
+			height="0" width="0" style="display:none;visibility:hidden">
+		</iframe>`;
+	return <noscript dangerouslySetInnerHtml={{ __html: iframe }}></noscript>
+};
+
+GoogleTagManagerNoscript.propTypes = {
+	key: PropTypes.string,
+};
```
mmcgahan: I generally leave out propTypes when using flow types - they're redundant for most applications

jmhammons: Ah, fair point.

PR #445, opened 2018-03-13
inject member id for cache key

mmcgahan:
By adding the current `memberId` into the cache key, we can avoid stale state errors coming from changes in logged-in/out status.

It also provides a minor additional privacy layer for cached data

### DIFF
--- a/packages/mwp-api-state/src/cache/util.js
+++ b/packages/mwp-api-state/src/cache/util.js
@@ -40,14 +40,17 @@ export function makeCache(): Cache {
 	return require('idb-keyval');
 }
 
+const makeKey = (memberId: string, query: Query) =>
+	`${memberId}${JSON.stringify(query)}`;
+
 /**
  * Generates a function that can read queries and return hits in the supplied cache
  */
-export const cacheReader = (cache: Cache) => (
+export const cacheReader = (cache: Cache, memberId: string) => (
 	query: Query
 ): Promise<QueryState> =>
 	cache
-		.get(JSON.stringify(query))
+		.get(makeKey(memberId, query))
 		.then((response: QueryResponse) => ({ query, response }))
 		.catch(err => ({ query, response: null })); // errors don't matter - just return null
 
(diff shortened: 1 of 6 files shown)


packages/mwp-api-state/src/cache/util.js
```diff
@@ -66,5 +69,5 @@ export const cacheWriter = (cache: Cache) => (
 		// skip cache writing
 		return Promise.resolve(true);
 	}
-	return cache.set(JSON.stringify(query), response);
+	return cache.set(makeKey(memberId, query), response);
```
eilinora: Just to make sure I'm 💯 understanding the cache that this will be affecting. It's this IndexDB cache that will be key'd, correct? Jusstttt want to make sure there isn't another caching layer i'm not aware of... besides redux state. :)

screen shot 2018-03-13 at 6 26 42 pm

jaclynj: @eilinora I found the docs :) https://meetup.atlassian.net/wiki/spaces/WEG/pages/243499188/Caching

```The cache middleware implementation uses IndexedDB for local storage of API response data - it caches the JSON response rather than the full HTTP response in order to avoid complexity arising from the batched API request/response routine.```

Comment (eilinora) [reply packages/mwp-api-state/src/cache/util.js]: Excellent! Thank you!

File: packages/mwp-api-state/src/cache/util.test.js [hunk 2]
```diff
@@ -34,8 +34,8 @@ describe('cache utils', () => {
 		const cache = makeCache();
 		const query = { foo: 'baz' };
 		const response = 'bar';
-		return cacheWriter(cache)(query, response)
-			.then(() => cache.get(JSON.stringify(query)))
+		return cacheWriter(cache, 'memberId')(query, response)
+			.then(() => cache.get(`memberId${JSON.stringify(query)}`))
```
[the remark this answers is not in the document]
mmcgahan: mainly because it's a test script so i wasn't sure that I wanted to hide the 'expected' string format... can argue for either option, but I figure it's probably ok to break this test when `makeKey` changes

PR #450, opened 2018-03-15
GTM: remove the noscript

jmhammons:
The [Google Tag Manager quick start guide](https://developers.google.com/tag-manager/quickstart) recommends a `<noscript />` tag as part of the set up.

After implementing this in `mup-web`'s `AppContainer.jsx`, React is raising a warning when a `<noscript>` tag is present.

`react.79819fe2.js:468 Warning: Expected server HTML to contain a matching <iframe> in <noscript>`

In gearing up to troubleshoot, Mike points out we can remove this all together because we aren't supporting clients without JS anyway, so we'll remove it here.


jmhammons:
Hey @mmcgahan -- given the contents of `src/components/GoogleTagManager.jsx` after this change, I'm wondering if we should move this to `src/util/googleTagManager.js` (or some other more appropriate place)?

mmcgahan:
yeah, I like that file-move recommendation

PR #453, opened 2018-03-19
MW-3351: Optimize Polyfill.io

sadafie:
Our New Relic dashboard has a chart indicating the most used browsers being Chrome 64 and 65, followed by Safari 11, Edge 16, and FF 58 and 59.

screen shot 2018-03-19 at 12 38 56 pm

The majority of our polyfills are currently for IE 11 and Safari, and possibly for partial support of older version of certain browsers.

The simplest first step fix I could think of is simply not include the polyfill script on server render for Chrome and Firefox, since we know those browsers won't need them.

After the corresponding update in mup-web, I would expect to see the TTI times in that NR chart to dip slightly for Chrome and FF, by approximately 300ms or so.

I'd propose to launch with this simple solution and figure out how much of a perf gain we get in Chrome and then decide if we want to optimize further for browsers that do need the polyfills.


packages/mwp-app-render/src/components/Dom.jsx
```diff
@@ -2,6 +2,8 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import escapeHtml from 'escape-html';
 
+import { polyfillServiceUrl } from 'mwp-app-render/lib/components/PageWrap/browserPolyfill';
```
sadafie: If we go with this approach, I think it makes sense to move the browserPolyfill out of PageWrap and next to Dom.jsx

packages/mwp-app-render/src/components/Dom.jsx
```diff
@@ -84,6 +92,12 @@ const DOM = props => {
 						`window.APP_RUNTIME=${JSON.stringify(APP_RUNTIME)};`
 					)}
 				/>
+				{doPolyfill &&
+					<script
```
mmcgahan: since we already have a `scripts` array prop to work with, I'd recommend just using the conditional to run `scripts.push(polyfillServiceUrl(initialState.config.requestLanguage))` rather than adding a lot of special markup for the polyfill script

packages/mwp-app-render/src/components/Dom.jsx
```diff
@@ -36,9 +38,15 @@ const DOM = props => {
...
+		userAgent,
 	} = props;
 
-	const htmlLang = initialState.config.requestLanguage.split('-')[0];
+	const localeCode = initialState.config.requestLanguage;
+	const htmlLang = localeCode.split('-')[0];
+
+	// Polyfill all browsers except Chrome and Firefox
+	// TODO: fine-tune this or leave is for now?
+	const doPolyfill = userAgent && userAgent.indexOf("Chrome") === -1 && userAgent.indexOf("Firefox") === -1;
```
mmcgahan: I recommend breaking this out into a function that can be called from here rather than adding specific functionality inline

e.g.

```js
const polyfillWhitelist = ['Chrome', 'Firefox'];
const getPolyfill = (userAgent, localeCode) = > // this is all nice and pure and unit-testable
  !polyfillWhitelist.some(userAgent.includes) &&
    polyfillServiceUrl(localeCode);

// ...
const polyfillSrc = getPolyfill(userAgent, localeCode);
```

mmcgahan: the `getPolyfill` function could/should actually live in the `browserPolyfill` module to reduce the amount of polyfill-related noise in `Dom.jsx`

packages/mwp-app-render/src/components/Dom.jsx
```diff
@@ -36,9 +38,19 @@ const DOM = props => {
...
-	const htmlLang = initialState.config.requestLanguage.split('-')[0];
+	const localeCode = initialState.config.requestLanguage;
+	const htmlLang = localeCode.split('-')[0];
+
+	/**
+	 * Add polyfill.io script if needed
+	 */
+	const polyfill = getPolyfill(userAgent, localeCode);
+	if(polyfill) {
+		scripts.push(polyfill);
```
sadafie: @mmcgahan I went with this approach, but we want the polyfill JS to load before the app code, and this doesn't achieve that since it sticks it to the end of the array. I think I'd prefer `unshift` here.

sadafie:
This is actually causing two side-effects in the mup-web application that I need to investigate more:

1. DateTime for other locales like `ja-JP`, render on the server as `M20` and then get replaced with the appropriate characters once the browser JS executes. 
2. In IE 11 I get several javascript errors that don't seem to happen on beta2.meetup.com. See screenshots below.

I'm trying to figure out what may have caused this. The polyfill is being loaded correctly in Safari, IE 11, and Edge. It doesn't load in Chrome and Firefox as intended. 

The only other change could be moving the script tag for the polyfill further down in the DOM, right before the app code, which is where we want it.

screen shot 2018-03-21 at 2 25 10 pm
screen shot 2018-03-21 at 2 25 28 pm

sadafie:
> DateTime for other locales like ja-JP, render on the server as M20 and then get replaced with the appropriate characters once the browser JS executes.

So, I built mup-web master and the flash of content happens on dev:master as well, so somehow it must be due to a difference between how dev and prod is built.

The URLs I'm comparing are:

https://beta2.dev.meetup.com:8000/ja-JP/Tokyo-Mums-Kids/
https://beta2.meetup.com/ja-JP/Tokyo-Mums-Kids/

packages/mwp-app-render/src/components/Dom.jsx
mmcgahan: ah, yeah, totally mean `unshift` 👍

packages/mwp-app-render/src/util/browserPolyfill.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * If the userAgent does not contain any of the blacklist strings,
+ * then return the polyfill url
+ *
+ * @param {String} userAgent  userAgent string of the current request
+ * @param {String} localeCode xx-XX locale string for current request
+ * @returns {Boolean|String} the polyfill service url or false
+ */
+export const getPolyfill = (userAgent, localeCode) => !userAgentPolyfillBlacklist.some(ua => userAgent.includes(ua)) && polyfillServiceUrl(localeCode);
```
mmcgahan: this script would be super simple to add Flowtype annotations to... just sayin'

sadafie: just pushed an update w/ flow. first time using it, lmk if there are any mistakes. it builds and tests fine.

sadafie:
> In IE 11 I get several javascript errors that don't seem to happen on beta2.meetup.com. See screenshots below.

This issue also seems to exist in IE 11 on dev:master, so assuming it's a dev environment issue and not related to prod builds of this branch.

PR #455, opened 2018-03-22
add New Relic traces for api calls, render method

mmcgahan:
I'm not 100% sure what we should expect to see in the New Relic APM transaction logs, but the `newrelic.createTracer` API is the thing that New Relic recommends using to add custom timing data to APM logs - the docs also indicate that this value can be consumed by Insights, which would be great.

So, this might need another iteration or two, but the expected behavior is:

1. A new trace that specifically corresponds to the time that the parallel Meetup REST API calls _start_ to the time that _all API calls finish_ - we don't have to infer timing info from the traces that are automatically generated for each individual call, which appear to contain some extraneous timeout data
2. new traces for each individual Meetup REST API call - the current request-specific trace values sometimes are measured as being much longer than the overall response time, which doesn't make a lot of sense, so hopefully this trace will be more useful
3. A new trace that specifically times the React rendering function - it should be only a few milliseconds, but we want to have a clear timing value for it, particularly for complex pages. We currently don't have any clear timing info about the render call independent of other aspects of request handling, so this should paint a much clearer picture about where time is spent on the server

Together, the ratio of (1) to (3) will tell us the balance of timing impact between the app server and the API server.


mmcgahan:
I've rebuilt all the yarn lockfiles to get grandchildren up to date

packages/mwp-api-proxy-plugin/src/util/receive.js
```diff
@@ -255,8 +256,10 @@ export const makeReceive = request => {
 	const injectResponseCookies = makeInjectResponseCookies(request);
 	return query => {
 		const parseApiResponse = makeParseApiResponse(query);
-		const apiResponseToQueryResponse = makeApiResponseToQueryResponse(query);
-		return response => {
+		const apiResponseToQueryResponse = makeApiResponseToQueryResponse(
+			query
+		);
+		return newrelic.createTracer('meetupApiRequest', response => {
```
eilinora: the only question i had was more around this, which looks to be tracing after the request resolves, do we have one in other PR for when it began?

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -292,16 +293,22 @@ const makeRenderer = (
...
+				getRouterRenderer({
+					routes,
+					store: initializedStore,
+					location: url,
+					basename,
+					scripts,
+					cssLinks,
+					userAgent,
+				})
+			)() // self-invoke
```
jaclynj: Why are we invoking here but in the other parts of this PR we are not?

packages/mwp-router/src/RouteLayout.jsx
```diff
@@ -25,7 +25,12 @@ class RouteLayout extends React.Component {
 							path={path}
 							exact={route.exact || false}
 							strict={route.strict || false}
-							render={props => <AsyncRoute {...props} route={route} />}
+							render={props =>
+								<AsyncRoute
+									{...props}
+									{...childProps}
```
jaclynj: could you explain why this was added? I'm not familiar with childProps, is that a react-router thing?

packages/mwp-api-proxy-plugin/src/util/receive.js
mmcgahan: yeah the lifecycle of `createTracer` is a little awkward to reason about, but I think this is correct - the trace timer starts when `createTracer` is called, and it finishes when the callback is called. In this case, `createTracer` on line 262 is called when the receiver is first initialized - which happens just _before_ the requests are sent. The callback is called when the responses are _received_. I'll have another look at this to see if there's a better way of composing send & receive to make the lifecycle more clear, but if it's a significant refactor I'll probably leave as-is for now.

packages/mwp-core/src/renderers/server-render.jsx
mmcgahan: good question! I've outlined what I _think_ will happen in the comments above where `createTracer` is called, and I'm eager to see how this works in practice, but I think this is the correct way of putting a timer around a synchronous function call.

`createTracer` is designed primarily for async callback-based patterns, but in this case I'm expecting the synchronous, blocking function to take a non-trivial amount of time, so the 'callback' is called immediately. According to the docs, the timer for `createTracer` starts when `createTracer` is first called and finishes _after the callback returns_, so this should effectively measure the execution time of the callback itself.

1. call `createTracer`
2. invoke callback
3. callback complete

packages/mwp-router/src/RouteLayout.jsx
mmcgahan: ah, this was actually something I was starting to sketch out in order to support injecting props into components attached to routes, but it wasn't working out and I didn't clean up - will fix!

PR #456, opened 2018-03-24
use browser-detect to better parse UA strings

mmcgahan:
Edge was masquerading as Chrome


packages/mwp-app-render/src/util/browserPolyfill.js
```diff
@@ -1,9 +1,10 @@
 // @flow
-
+// use mobile-detect to check browser agent name because we are using it in server-render as well
+import detect from 'browser-detect';
```
eilinora: ♥️ this solution!

PR #457, opened 2018-03-26
MG-719: Support click tracking headers

osermay:
In order to track urls with `_xtd` query params in mup-web
- check if query param exists
- add `X-Meetup-External-Track` and `X-Meetup-External-Track-Url ` onPreResponse
- add tests

### DIFF
--- a/packages/mwp-app-route-plugin/src/route.js
+++ b/packages/mwp-app-route-plugin/src/route.js
@@ -12,7 +12,11 @@ export const onPreResponse = {
 	 */
 	method: (request: HapiRequest, reply: HapiReply) => {
 		const response = request.response;
-
+		const trackingParam = request.query['_xtd'];
+		if (trackingParam) {
+			response.header('X-Meetup-External-Track', trackingParam);
+			response.header('X-Meetup-External-Track-Url', request.response.referer);
+		}
 		if (!response.isBoom || process.env.NODE_ENV === 'production') {
 			return reply.continue();
 		}
--- a/packages/mwp-app-route-plugin/src/route.js
+++ b/packages/mwp-app-route-plugin/src/route.js
@@ -3,6 +3,9 @@ import React from 'react';
 import ReactDOMServer from 'react-dom/server';
 import getHandler from './handler';
 
+export const EXTERNAL_TRACK_HEADER = 'x-meetup-external-track';
+export const EXTERNAL_TRACK_URL_HEADER = 'x-meetup-external-track-url';
+
 export const onPreResponse = {
 	/*
 	 * This function processes the route response before it is sent to the client.
@@ -12,10 +15,12 @@ export const onPreResponse = {
 	 */
 	method: (request: HapiRequest, reply: HapiReply) => {
 		const response = request.response;
-		const trackingParam = request.query['_xtd'];
-		if (trackingParam) {
-			response.header('X-Meetup-External-Track', trackingParam);
-			response.header('X-Meetup-External-Track-Url', request.response.referer);
+		if (request.query) {
+			const trackingParam = request.query['_xtd']
+			if (trackingParam) {
+				response.header(EXTERNAL_TRACK_HEADER, trackingParam);
+				response.header(EXTERNAL_TRACK_URL_HEADER, request.url.href);
+			}
 		}
 		if (!response.isBoom || process.env.NODE_ENV === 'production') {
 			return reply.continue();
--- a/packages/mwp-app-route-plugin/src/route.test.js
+++ b/packages/mwp-app-route-plugin/src/route.test.js
@@ -9,7 +9,7 @@ describe('onPreResponse.method', () => {
 		const errorMessage = 'foobar';
 		const errorCode = 432;
 		const response = Boom.create(errorCode, errorMessage);
-		response.header = ((key,val) =>val)
+		response.header = ((key, val) => val)
 
 		const request = {
 			response,
@@ -17,7 +17,7 @@ describe('onPreResponse.method', () => {
 			server: getServer(),
 		};
 		const replyObj = {
-			code() {},
+			code() { },
 		};
 		const spyable = {
 			reply: () => replyObj,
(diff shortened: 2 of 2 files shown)


packages/mwp-app-route-plugin/src/route.js
```diff
@@ -12,7 +15,13 @@ export const onPreResponse = {
 	 */
 	method: (request: HapiRequest, reply: HapiReply) => {
 		const response = request.response;
-
+		if (request.query) {
```
mmcgahan: I _think_ you can always count on `request.query` existing, even if it's an empty object.

packages/mwp-app-route-plugin/src/route.js
```diff
@@ -12,7 +15,13 @@ export const onPreResponse = {
...
+			const trackingParam = request.query['_xtd']
```
mmcgahan: linter should probably be set up to catch this, but you can use dot notation `request.query._xtd`

packages/mwp-app-route-plugin/src/route.js
```diff
@@ -12,7 +15,13 @@ export const onPreResponse = {
...
+			if (trackingParam) {
+				response.header(EXTERNAL_TRACK_HEADER, trackingParam);
```
mmcgahan: this behavior definitely needs some documentation so that people can understand the intent in the future - what is being tracked? where does the _xtd param come from and what is its expected value/format? Maybe just a link to documentation elsewhere, but otherwise it's kinda mystery meat.

packages/mwp-app-route-plugin/src/route.test.js
```diff
@@ -30,6 +30,22 @@ describe('onPreResponse.method', () => {
 		expect(errorMarkup).toContain(errorMessage);
 		expect(replyObj.code).toHaveBeenCalledWith(errorCode);
 	});
+	it('returns X-Meetup-External-Track and X-Meetup-External-Track-Url headers if _xtd query param exists', () => {
```
mmcgahan: do you want a test to make sure these headers are _not_ set when the query param is _not_ present? If someone accidentally broke the `if (trackingParam) {` check, this test would pass even though it wouldn't be working as intended.

packages/mwp-app-route-plugin/src/route.js
mmcgahan: I think this behavior might not actually be what you're hoping for - you're sending these headers back to the browser, but I would guess that the browser will just ignore them in the response? Are you intending to send them along with REST API requests instead? Is there another part of the ticket that add browser-side handling? The ticket and PR description are vague about how the headers are used.

osermay: I will ask for a more clarified ticket description but the general idea is to allow for email tracking as stated in https://meetup.atlassian.net/wiki/spaces/DAT/pages/27754630/Email+Tracking

osermay: No browser-handling

mmcgahan: ok, so `onPreResponse` modifies the response sent back to the browser, but the docs you linked indicate that the custom headers need to be sent to the REST API - the current implementation won't affect the requests to API.

Instead, you'll probably want to modify [this function](https://github.com/meetup/meetup-web-platform/blob/master/packages/mwp-api-proxy-plugin/src/util/send.js#L200) to inject custom headers into the REST API requests. One thing you'll have to find out from the data team is whether it's okay to add those headers to _all_ parallel REST API requests that correspond to opening an email or whether the headers should only be sent with one of the REST API requests - it's possible that the BE assumes that there will be only one REST request per application request.

packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -197,11 +197,23 @@ export function getClientIpHeader(request) {
 	}
 }
 
+export function getTrackingHeaders(request) {
+	// email tracking with _xtd query param: https://meetup.atlassian.net/wiki/spaces/DAT/pages/27754630/Email+Tracking
+	var trackingParam = request.query._xtd;
```
mmcgahan: prefer `const` here - there's almost never a reason to use `var` any more

packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -197,11 +197,23 @@ export function getClientIpHeader(request) {
...
+	if (trackingParam) {
+		return {
+			"X-Meetup-External-Track": trackingParam,
+			"X-Meetup-External-Track-Url": request.url.href,
+		};
+	}
```
mmcgahan: just to keep return types consistent, I'd recommend returning an empty object here - i.e. the function always returns an object rather than an `?Object`

packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -197,11 +197,23 @@ export function getClientIpHeader(request) {
 	}
 }
 
+export function getTrackingHeaders(request) {
+	// email tracking with _xtd query param: https://meetup.atlassian.net/wiki/spaces/DAT/pages/27754630/Email+Tracking
+	var trackingParam = request.query._xtd;
+	if (trackingParam) {
+		return {
+			"X-Meetup-External-Track": trackingParam,
+			"X-Meetup-External-Track-Url": request.url.href,
```
mmcgahan: Looks like Prettier might be behaving correctly - these should be single quotes

packages/mwp-api-proxy-plugin/src/util/send.js
osermay: should that be added to 
```
export function getClientIpHeader(request) {
	const clientIP =
		request.query['_set_geoip'] || request.headers['fastly-client-ip'];
	if (clientIP) {
		return { 'X-Meetup-Client-Ip': clientIP };
	}
}
```

packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -271,10 +283,10 @@ export const makeMockRequest = (
 	mockResponseContent,
 	responseMeta
 ) => requestOpts =>
-	Observable.of([
-		makeMockResponse(requestOpts, responseMeta),
-		JSON.stringify(mockResponseContent),
-	]);
+		Observable.of([
```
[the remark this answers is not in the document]
osermay: should i run `yarn lint`? if this does it automatically with eslint how do i prevent that?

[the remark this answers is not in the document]
burrata: not sure -- sometimes this happens because of a discrepancy between spaces and tabs, so I'd check that first. then, I'd maybe check when the last change was made -- it's possible that the linting config may have changed since then.

packages/mwp-api-proxy-plugin/src/util/send.js
mmcgahan: sure!

PR #460, opened 2018-03-27
WP-672: Added ability of pushing to GTM

DA-Denis:
Added dataLayer.push as shared utils method.

### DIFF
--- a/packages/mwp-app-render/src/util/googleTagManager.js
+++ b/packages/mwp-app-render/src/util/googleTagManager.js
@@ -5,14 +5,15 @@ const GTM_KEY = process.env.NODE_ENV === 'production' ? 'GTM-T2LNGD' : 'GTM-W9W8
 
 /**
  * @description Method for passing additional variables to GTM
- * @param variablesList - object which contains custom variables. example: { variableAccessibleInGTM: "foo" }
- * window.dataLayer - global reserved GTM array
+ * @see {@link https://developers.google.com/tag-manager/devguide}
  */
-export const gtmPush = (variablesList) => {
-	window.dataLayer.push(variablesList);
+export const gtmPush = (data: { [string]: string }) => {
+	if (typeof window !== 'undefined' && window.dataLayer) {
+		window.dataLayer.push(data);
+	}
 };
 
-/*
+/**
  * @description Gets google tag manager JS snippet
  * @see {@link https://developers.google.com/tag-manager/quickstart}
 */
--- a/packages/mwp-app-render/src/util/googleTagManager.js
+++ b/packages/mwp-app-render/src/util/googleTagManager.js
@@ -17,7 +17,7 @@ export const gtmPush = (data: { [string]: string }) => {
  * @description Gets google tag manager JS snippet
  * @see {@link https://developers.google.com/tag-manager/quickstart}
 */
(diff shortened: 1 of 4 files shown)


packages/mwp-app-render/src/util/googleTagManager.js
```diff
@@ -3,12 +3,22 @@
 // Production and dev keys
 const GTM_KEY = process.env.NODE_ENV === 'production' ? 'GTM-T2LNGD' : 'GTM-W9W847';
 
+/**
+ * @description Method for passing additional variables to GTM
+ * @param variablesList - object which contains custom variables. example: { variableAccessibleInGTM: "foo" }
```
emilyeng: maybe worth renaming this param to `variables` since it's technically not a "list" and list typically suggests an array

mmcgahan: we're using Flowtype in this file - no need for JSdoc `param`, but you'll need to annotate the argument with something like `{ [string]: string }` (which is a 'map', not a list, as @emilyeng mentioned)

I had to look up some documentation to understand how this feature works - it would be helpful to link to the Google dev documentation here to steer future maintainers in the right direction: https://developers.google.com/tag-manager/devguide.

The current implementation of `gtmPush` is a little fragile because it will only work in the browser - I recommend thinking about how it might be refactored to work on the server (using Node's `global` object, perhaps, or a better-namespaced location) - be careful about making sure you get a new copy for each request, however, rather than `push`ing into an array forever. It might be that `gtmPush` should only be used in browser-side event handlers, but that's an important constraint to emphasize through code or comments.

packages/mwp-app-render/src/util/googleTagManager.js
```diff
@@ -3,12 +3,22 @@
...
+ * window.dataLayer - global reserved GTM array
+ */
+export const gtmPush = (variablesList) => {
```
mmcgahan: Let's add a unit test for this as well

packages/mwp-app-render/src/util/googleTagManager.js
jmhammons: Few thoughts:
- I'm not a flow types pro yet, but wondering if this should be `export const gtmPush = (variablesList: Object) => {` or something of the sort
- Since I had to look up what this is, might be good to `@see {@link` a good resource -- maybe https://developers.google.com/tag-manager/devguide?
- Also is probably worth updating a use case in https://meetup.atlassian.net/wiki/spaces/WEG/pages/438632854/Get+started+with+Google+Tag+Manager

packages/mwp-app-render/src/util/seo/__snapshots__/ldJson.test.js.snap
```diff
@@ -31,7 +31,7 @@ Object {
     "price": 10,
     "priceCurrency": "USD",
   },
-  "startDate": "1970-01-18T04:01-05:00",
+  "startDate": "1970-01-18T05:01-04:00",
```
emilyeng: hmm i wonder why this changed in the snap?

DA-Denis: it's because of my timezone. I'll change it back

PR #461, opened 2018-03-27
Add full url to X-Meetup-External-Track-Url header

osermay:
This is the original ticket: https://meetup.atlassian.net/browse/MG-719
This is a follow up to this: https://github.com/meetup/meetup-web-platform/pull/457

Originally, the 'route' or everything after `https://meetup.com` was being logged as the track url. 
For tracking purposes, we need log the full url. This adds that.

### DIFF
--- a/packages/mwp-api-proxy-plugin/src/util/send.js
+++ b/packages/mwp-api-proxy-plugin/src/util/send.js
@@ -200,11 +200,19 @@ export function getClientIpHeader(request) {
 
 export function getTrackingHeaders(request) {
 	// email tracking with _xtd query param: https://meetup.atlassian.net/wiki/spaces/DAT/pages/27754630/Email+Tracking
+
+	// request protocol and host might be different from original request that hit proxy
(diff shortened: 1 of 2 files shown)


packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -200,11 +200,19 @@ export function getClientIpHeader(request) {
...
+	const domain =
+		request.headers['x-forwarded-host'] || request.headers['x-meetup-host'];
+	const host = `${requestProtocol}://${domain}`;
+
 	const trackingParam = request.query._xtd;
 	if (trackingParam) {
 		return {
 			'X-Meetup-External-Track': trackingParam,
-			'X-Meetup-External-Track-Url': request.url.href,
+			'X-Meetup-External-Track-Url': `${host}` + request.url.href,
```
mmcgahan: prefer template string for the whole thing `${host}${request.url.href}`

PR #463, opened 2018-03-28
New Relic Browser instrumentation with Node agent

sadafie:
Just a prelim to test with.


sadafie:
Testing this in dev on mup-web, I'm seeing the `newrelic.getBrowserTimingHeader()` method returning an html comment ```<!-- NREUM(1) -->```:

screen shot 2018-03-29 at 12 26 29 pm
screen shot 2018-03-29 at 12 28 13 pm


According to the docs [here](https://docs.newrelic.com/docs/agents/nodejs-agent/supported-features/new-relic-browser-nodejs-agent):

> If Browser is disabled, or if there is an error so that working headers cannot be generated, the newrelic module generates an innocuous html comment.

I think this is happening for one of two reasons (or both):

1. Dev environment doesn't communicate with new relic
2. We don't have a node deployed browser app set up for mup-web yet.

I'm going to go ahead and do number 2 and see if that improves things in dev.

sadafie:
I got confirmation from New Relic support that the next steps would be:

1. Deploy these changes to our application. 
2. Delete the mup-web Browser app that's deployed via copy/paste instrumentation (https://rpm.newrelic.com/accounts/1238700/browser/36183534) 
3. Deploy a new mup-web Browser app using the new relic agent by going to https://rpm.newrelic.com/accounts/1238700/browser > Add More > Enable Via New Relic APM w/ Pro + SPA instrumentation for the mup-web browser application.

If we don't see data populating in Browser after step 3, we will need to troubleshoot from there. Because I want to be available for any debugging and I'm OOO tomorrow, I'm going to hold off and do all this on Monday of next week.

mmcgahan:
@sadafie Is this ready for review? Code changes LGTM

PR #467, opened 2018-04-05
WP 680 LaunchDarkly flags on server

mmcgahan:
simplest diff view: https://github.com/meetup/meetup-web-platform/pull/467/files?diff=split&w=1

This PR sets up the LaunchDarkly NodeJS SDK in the app-route-plugin so that every _server-rendered_ request can inject all the _member-specific_ feature flag values into Redux state at `state.flags`.

This is a short term solution - the longer term solution is to use the Query system to allow queries to request specific flags, which can then be delivered every time that query is dispatched rather than blindly loading all flag values on server render. Requesting specific flags also allows LaunchDarkly to track requests by each user, which helps auditing.

```js
state.flags = {
  'this-is-my-flag-key': false,
  ...
}
```

Followup:
1. Create a new platform-supplied `withFeatureFlag` HOC that reads flag values from `state.flags` but otherwise returns the same data (i.e. an array of 'true' flag keys)
2. Augment routing definition with the ability to specify feature flags needed for a particular route, including server-side support for those requests in server-render and API proxy.

### DIFF
--- a/packages/mwp-app-route-plugin/src/index.js
+++ b/packages/mwp-app-route-plugin/src/index.js
@@ -2,7 +2,6 @@
 import LaunchDarkly from 'ldclient-node';
 import getRoute from './route';
 
-// set up launchdarkly instance before continuing
 const LAUNCH_DARKLY_SDK_KEY = 'sdk-86b4c7a9-a450-4527-a572-c80a603a200f';
 
 /*
@@ -21,6 +20,7 @@ export default function register(
 
 	const ldClient = LaunchDarkly.init(options.ldkey || LAUNCH_DARKLY_SDK_KEY);
 	server.expose('getFlags', ldClient.all_flags);
+	// set up launchdarkly instance before continuing
 	ldClient.once(`ready`, () => next());
 }
 
--- a/packages/mwp-app-route-plugin/src/index.js
+++ b/packages/mwp-app-route-plugin/src/index.js
@@ -19,7 +19,18 @@ export default function register(
 	server.route(getRoute(options.languageRenderers));
 
 	const ldClient = LaunchDarkly.init(options.ldkey || LAUNCH_DARKLY_SDK_KEY);
-	server.expose('getFlags', ldClient.all_flags);
+	server.expose('getFlags', memberId =>
+		ldClient.all_flags(memberId).then(
+			flags => flags,
+			err => {
+				server.app.logger.error({
+					err,
+					memberId,
+				});
+				return {}; // return empty flags on error
+			}
+		)
+	);
 	// set up launchdarkly instance before continuing
 	ldClient.once(`ready`, () => next());
 }
--- a/packages/mwp-app-route-plugin/src/index.js
+++ b/packages/mwp-app-route-plugin/src/index.js
@@ -13,7 +13,7 @@ export default function register(
 	server: HapiServer,
 	options: {
 		languageRenderers: { [string]: LanguageRenderer },
-		ldkey: string,
+		ldkey?: string,
 	},
 	next: () => void
 ) {
@@ -21,9 +21,7 @@ export default function register(
 
 	const ldClient = LaunchDarkly.init(options.ldkey || LAUNCH_DARKLY_SDK_KEY);
 	server.expose('getFlags', ldClient.all_flags);
-	ldClient.once(`ready`, () => {
-		next();
-	});
+	ldClient.once(`ready`, () => next());
 }
 
 register.attributes = {
--- a/packages/mwp-app-route-plugin/src/index.js
+++ b/packages/mwp-app-route-plugin/src/index.js
@@ -1,5 +1,9 @@
 // @flow
+import LaunchDarkly from 'ldclient-node';
 import getRoute from './route';
+
+// set up launchdarkly instance before continuing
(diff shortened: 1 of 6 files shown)


packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -233,80 +233,87 @@ const makeRenderer = (
...
-			enableServiceWorker,
-			requestLanguage,
-			supportedLangs,
-			initialNow: new Date().getTime(),
-			variants: getVariants(state),
-			entryPath: url.pathname, // the path that the user entered the app on
-			media: getMedia(userAgent, userAgentDevice),
-		},
-	};
+	const initializeStore = () =>
```
mmcgahan: this is a refactor to support async flag fetching from `request.server['mwp-app-route-plugin'].getFlags` before setting up the initial state of the store

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -233,80 +233,87 @@ const makeRenderer = (
 	const host = `${requestProtocol}://${domain}`;
 	const userAgent = headers['user-agent'];
 	const userAgentDevice = headers['x-ua-device'] || ''; // set by fastly
+	const memberId = parseMemberCookie(state).id;
```
eilinora: what if a member isn't logged in? should we have a fallback anonymous id

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -233,80 +233,87 @@ const makeRenderer = (
...
-	const store = createStore(reducer, initialState);
+			const createStore = getServerCreateStore(
+				getRouteResolver(routes, basename),
+				middleware,
+				request
+			);
+			return Promise.resolve(createStore(reducer, initialState));
+		});
 
 	// render skeleton if requested - the store is ready
```
eilinora: comment probably needs to move down to L297

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -233,80 +233,87 @@ const makeRenderer = (
...
-				store: initializedStore,
-				location: url,
-				basename,
-				scripts,
-				cssLinks,
-				userAgent,
-			}) // immediately invoke callback
-	);
+	return initializeStore().then(store => {
+		if ('skeleton' in request.query) {
```
eilinora: where does 'skeleton' come from?

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -233,80 +233,87 @@ const makeRenderer = (
...
-			requestLanguage,
-			supportedLangs,
-			initialNow: new Date().getTime(),
-			variants: getVariants(state),
-			entryPath: url.pathname, // the path that the user entered the app on
-			media: getMedia(userAgent, userAgentDevice),
-		},
-	};
+	const initializeStore = () =>
+		request.server['mwp-app-route-plugin'].getFlags(memberId).then(flags => {
```
eilinora: what if getFlags fails?

packages/mwp-core/src/renderers/server-render.jsx
mmcgahan: `parseMemberCookie` handles it

packages/mwp-core/src/renderers/server-render.jsx
mmcgahan: this is an old part of the offline-first work we did way back when. It's still useful, in theory, but is existing code that isn't really affected by this ticket

packages/mwp-core/src/renderers/server-render.jsx
mmcgahan: good call - will add a `catch`

packages/mwp-core/src/renderers/server-render.jsx
eilinora: cool cool

packages/mwp-core/src/renderers/server-render.jsx
mmcgahan: I've moved the error handling into the `getFlags` function itself - it is now guaranteed to return an object (even if the object is empty because the flags couldn't be read for some reason)

packages/mwp-app-route-plugin/src/index.js
```diff
@@ -1,18 +1,38 @@
 // @flow
+import LaunchDarkly from 'ldclient-node';
 import getRoute from './route';
 
+const LAUNCH_DARKLY_SDK_KEY = 'sdk-86b4c7a9-a450-4527-a572-c80a603a200f';
```
eilinora: why not make this a something travis fills in?

mmcgahan: mainly because that pushes responsibility to the apps to handle the key rather than it being something that Just Works without the apps having to do anything in particular - I've set it up so that in theory you pass in a unique `ldkey` (although the source of that option would need to be set up with a code update), but my understanding is that we'll be able to consolidate all LaunchDarkly flags under one 'account' for the time being.

eilinora: True, just more was trying to avoid keys in our code :)

mmcgahan: yeah I should probably follow up to set this differently

PR #469, opened 2018-04-05
Fix New Relic browser script injection

sadafie:
When this launched we were seeing the new relic script tag render code into the page:

group_home


sadafie:
I'm testing this by updating mup-web to point to this beta version, and then running --

```
NODE_ENV=production BUILD_ARGS='--locales en-US' make -j4 build-app; 
yarn run start:prod
```

And going to a http (not https) version of the dev site. 

I'm able to reproduce the problem we were getting in production on the master version. However, when I'm testing these changes locally, I get an error in the new relic module code specifically at the import line in PageWrap.js -- 

screen shot 2018-04-05 at 2 01 50 pm

I'm not sure why the import works from within `Dom.jsx` (master) but errors in `Pagewrap.jsx` (this branch).


Thoughts? Or alternative implementation ideas?

sadafie:
Stripping the `<script>` tags and using dangerouslySetInnerHTML seems to work ... here's the screenshot

screen shot 2018-04-05 at 4 06 21 pm

packages/mwp-app-render/src/components/Dom.jsx
```diff
@@ -69,14 +69,22 @@ const DOM = props => {
 		escapedState,
 	};
 
+	// strip <script> html tags
+	// in order to dangerouslySetInnerHTML in JSX
+	const newrelicJS = newrelic
+		.getBrowserTimingHeader()
+		.replace(/<script[^>]+>/, '')
```
mmcgahan: I guess technically you could have `/<script[^>]*>/` here (`*` instead of `+`), which would allow it to work if they ever changed the implementation to start with `<script>` instead of `<script type='...'>`

sadafie: Wondering if this should be `/<script[^>]*>/` instead to match `<script>`

PR #470, opened 2018-04-05
Use `Helmet` to manage module styles

akdetrick:
https://meetup.atlassian.net/browse/WP-674

- replaced `simple-universal-style-loader` with `isomorphic-style-loader`
- removed `inlineStyleTags` prop from `Dom.jsx`along with supporting code
- added `WriteStyles` component that allows consumer applications to inject style module CSS into `head` via `Helmet`

Thanks to `Helmet`, we do not need to use any features of `isomorphic-style-loader` that rely on React context. However, we do need to use this loader for the following reasons:
- it provides a map of class names to localized class names calculated in `css-loader`
- it provides an interface for getting CSS content from a `styles` module imported into a component.

We may be able to eliminate `isomorphic-style-loader` down the line. [A ticket to follow up has been filed](https://meetup.atlassian.net/browse/WP-681).


packages/mwp-app-render/src/components/WriteStyles.jsx
```diff
@@ -0,0 +1,44 @@
...
+const WriteStyles = props => {
+	const moduleCSS = new Set();
+	props.styles.forEach(style => {
+		moduleCSS.add(style._getCss()); // eslint-disable-line no-underscore-dangle
+	});
+
+	return (
+		<Helmet defer={false}>
+			<style type="text/css">
+				{[...moduleCSS].join('')}
```
mmcgahan: I _think_ it would be better to output this with `dangerouslySetInnerHtml` so that valid CSS characters like `>` don't get HTML-escaped.

packages/mwp-app-render/src/components/WriteStyles.jsx
```diff
@@ -0,0 +1,44 @@
...
+ *		return (
+ *			<div>
+	 *			<WriteStyles styles={styles} />
+	 *			<p className={styles.someClass}>Hello world!</p>
+ *			</div>
+ *		);
+ * };
+ * ```
+ */
+const WriteStyles = props => {
```
mmcgahan: `WriteStyles` is a fine name, but I'm wondering if it's too implementation-specific - i.e. it refers to the fact that in its current form, the styles are written into the markup, but from the developers perspective, they just want to know that the styles are being made available, not _how_ they are being made available. The difference might be academic, but what about making `WriteStyles` a wrapper that implicitly provides a 'styled context' to its children, either as a HOC function (e.g. `withStyles(styles)(WrappedComponent)`) or as a component that simply renders its children

```jsx
<StyleWrapper styles={styles}>
  <WrappedComponent />
</StyleWrapper>
```

I would actually lean toward the latter because it matches my intuition about how I would want styles to be provided to my component (even if it doesn't actually match what the implementation does), and it might give you a good excuse to play with [React Fragments](https://reactjs.org/docs/fragments.html#short-syntax) like

```jsx
class StyleWrapper extends React.Component {
  render() {
    <>
      <Helmet>...</Helmet>
      {children}
    </>
  }
}
```

If you think `WriteStyles` as a sibling element is the right way to set this up, that's cool too, no change necessary

mmcgahan: I was also thinking `StyleProvider` would be a nice name, but `Provider` has become the convention for labeling React Context providers, and we don't want to make that implication here

akdetrick: I had actually named this `StylesProvider`, and @mperrotti pointed out that it's not _actually_ providing anything in the way that other `[Thing]Provider` components in our apps do.

I've been divided on sibling vs. wrapping with `WriteStyles`; using it as a sibling seems to break a convention, so I think i'm going to change this to wrap children. Perhaps `<StyleWith styles={styles}>...children...</StyleWith>`?

mmcgahan: sounds good to me

packages/mwp-app-render/src/components/StyleWith.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+	return (
+		<div>
+			<Helmet defer={false}>
+				<style type="text/css">
+					{[...moduleCSS].join('')}
+				</style>
+			</Helmet>
+			{children}
+		</div>
```
mmcgahan: once we update to React v16.2+, we can use Fragments here

PR #472, opened 2018-04-06
try again with github_deploy_key

mmcgahan:


mmcgahan:
I manually deleted `GITHUB_DEPLOY_KEY` and then re-ran `/repo-strap` in Slack to repopulate it, and it now works... my guess is that the script did _not_ overwrite the key and therefore we were using an invalid value.

PR #473, opened 2018-04-09
check if getCss is defined in StylesWith to prevent test errors

akdetrick:
Because tests do not run style imports through our webpack loaders, the `_getCss` function will not be defined in any style objects passed into `StyleWith` when run in a unit test.

This branch adds a check to see if the function is defined before invoking it.

### DIFF
--- a/packages/mwp-app-render/src/components/StyleWith.jsx
+++ b/packages/mwp-app-render/src/components/StyleWith.jsx
@@ -28,7 +28,9 @@ const StyleWith = props => {
 
 	const moduleCSS = new Set();
 	styles.forEach(style => {
-		moduleCSS.add(style._getCss()); // eslint-disable-line no-underscore-dangle
+		if (typeof style._getCss === 'function') {
+			moduleCSS.add(style._getCss()); // eslint-disable-line no-underscore-dangle
+		}
 	});
 
 	return (


packages/mwp-app-render/src/components/StyleWith.jsx
```diff
@@ -28,7 +28,9 @@ const StyleWith = props => {
 
 	const moduleCSS = new Set();
 	styles.forEach(style => {
-		moduleCSS.add(style._getCss()); // eslint-disable-line no-underscore-dangle
+		if (typeof style._getCss === 'function') {
```
mmcgahan: might be worth a code comment to explain when this would _not_ be true. Also, is there a way to add a mock somewhere that would set this up automatically in tests? I can't remember the details of what how the loader works, but if we could use `moduleNameMapper` config in Jest to get this to work, that would be preferable.

akdetrick: Yeah, this is better handled by mocks for `module.scss` and `StyleWith` imports. Closing this PR.

akdetrick:
Instead of adding this check in `StyleWith` for the sake of tests, we will instead use `moduleNameMapper` and mocks for `.module.scss` and `StyleWith` imports.

PR #474, opened 2018-04-10
WP 669 parallel routes

mmcgahan:
This PR adds support for `getComponent`, which is a way of defining an async 'getter' for a component attached to a `route`. It **removes** support for `getNestedRoutes` and `getIndexRoute`, which forced async routes to be resolved sequentially rather than in parallel.

This new setup gets used in 3 places

1. `server-render` - when the server starts, it immediately resolves _all_ the components in the route tree (because async loading is free on the server), and just uses the full `resolvedRoutes` for all further request handling
2. `browser-render` - when the client app starts, it first determines which routes will be 'matched' by the current URL, and then downloads all of the async components defined by the matched routes - any async components defined by _unmatched_ routes will not be loaded. The downloaded components will be cached for later use.
3. `AsyncRoute` - this is the component that turns the routing tree into nested React components - when it finds a route that has a `getComponent` param, it will initially render a 'PassThrough' component and then when the real component gets resolved, it will re-render with the expected component.


packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -203,125 +203,133 @@ const makeRenderer = (
 	scripts: Array<string> = [],
 	enableServiceWorker: boolean,
 	cssLinks: ?Array<string>
-) => (request: HapiRequest, reply: HapiReply): Promise<RenderResult> => {
-	middleware = middleware || [];
+) => {
+	// set up a Promise that emits the resolved routes - this single Promise will
+	// be reused for all subsequent requests, so we're not resolving the routes repeatedly
+	// hooray performance
+	const routesPromise = resolveAllRoutes(routes);
```
mmcgahan: big indentation change because this promise gets set up once on server load (see code comment)

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -203,125 +203,133 @@ const makeRenderer = (
...
-					store,
-					location: url,
-					basename,
-					scripts,
-					cssLinks,
-					userAgent,
-				}) // immediately invoke callback
-		);
-	});
+		return routesPromise.then(resolvedRoutes => {
```
mmcgahan: this promise is 'resolved' on server startup, but we can attach a `.then` handler for each new request and it will get the resolved value 'instantly'

packages/mwp-router/src/AsyncRoute.jsx
```diff
@@ -10,90 +10,86 @@ type Props = {
...
 	history: RouterHistory,
 };
-type State = {
-	childRoutes: Array<PlatformRoute>,
-	_routesCache: { [string]: Array<PlatformRoute> },
+type ComponentState = {
+	component: React$ComponentType<*>,
+	_componentCache: { [string]: React$ComponentType<*> },
 };
+type State = ComponentState;
```
mmcgahan: This looks redundant, but it allows me to write a state-setting function that is explicitly limited to updating 'component state', while the component itself can get other state params in the future.

packages/mwp-router/src/AsyncRoute.jsx
```diff
@@ -10,90 +10,86 @@ type Props = {
...
+
+// simple pass through component to use while real component is loading
+const PassThrough = (children: React$Node) => React.Children.only(children);
```
mmcgahan: I'm interested to see how this plays in a real application - it effectively allows 'inner' routes/containers to render _before_ their parent. For the most part, this won't matter, but there may be cases where a parent provides some wrapping chrome that will pop in after a delay - should be rare, and I think that's preferable to _not_ rendering the child content (which is what the people came to see) if it's ready before the parent wrapper.

packages/mwp-router/src/util/matcher.js
```diff
@@ -0,0 +1,74 @@
+// @flow
```
mmcgahan: this module is a little refactor of some things that were in `resolve.js` - it includes _synchronous_ functions that return `route` objects that match a passed-in location/URL. The previous implementations were async functions that used `getNestedRoutes`/`getIndexRoute`, but those features are gone so we can do things more simply. There's still some recursion, but it should be more clear - please ask questions if the docs are insufficient!

packages/mwp-router/src/util/query.js
```diff
@@ -50,13 +49,5 @@ const _matchedRouteQueriesReducer = (location: URL) => (
  */
 export const getMatchedQueries = (location: URL) => (
 	matchedRoutes: Array<MatchedRoute>
-): Array<Query> => matchedRoutes.reduce(_matchedRouteQueriesReducer(location), []);
-
-/*
- * A curried interface into `_resolveRouteMatches` + `getMatchedQueries`
- */
-export const activeRouteQueries = (routes: Array<PlatformRoute>, baseUrl: string) => {
```
mmcgahan: `activeRouteQueries` wasn't being used anywhere

packages/mwp-router/src/util/resolve.js
```diff
@@ -1,119 +1,54 @@
 // @flow
```
mmcgahan: check out the new module in split view - most of the deletions are things that were moved, so you don't need to look at them much.

packages/mwp-router/src/util/resolve.js
```diff
@@ -1,119 +1,54 @@
...
- * An curried interface into `_resolveRouteMatches`, using `basename`
- * + `location` instead of `path`
- */
-export const getRouteResolver = (routes: Array<PlatformRoute>, basename: string) => (
-	location: URL
-): Promise<Array<MatchedRoute>> => {
-	const path = location.pathname.replace(basename, '');
-	return _resolveRouteMatches(routes, path);
-};
+export const resolveAllRoutes = (
```
mmcgahan: not sure if it's really worth implementing this as a standalone function since the implementation is so simple - parent could call `Promise.all(routes.map(resolveRoute));` itself pretty easily...

packages/mwp-router/src/AsyncRoute.jsx
```diff
@@ -10,90 +10,86 @@ type Props = {
...
-		this.setState(state => ({ childRoutes }));
-		if (childRoutes.length) {
-			return;
+			return; // no new route, just re-render normally
 		}
-		if (match.isExact && route.getIndexRoute) {
-			this.resolveAsyncChildRoutes(route.getIndexRoute);
+
+		if (route.component) {
+			this.setState(state => ({ component: route.component }));
```
jaclynj: is it possible that `route.component` would already equal `state.component` ?

mmcgahan: if two different routes specified the same component, yes. Since new routes will trigger a re-render regardless, however, this isn't 'expensive' in that case.

packages/mwp-api-state/src/sync/index.js
```diff
@@ -74,35 +74,31 @@ export const getNavEpic = resolveRoutes => (action, store) => {
...
-				.map(JSON.parse)
-				.map(q => q.ref);
-			requestMetadata.retainRefs = sharedRefs;
-			return [
-				cacheAction,
-				api.get(newQueries, requestMetadata),
-				clickActions.clear(),
-			].filter(a => a);
-		}
+	return Promise.resolve(
```
eilinora: I have some assumptions as to why Promise.all => Promise.resolve but would ♥️ hearing thoughts on the change.

packages/mwp-api-state/src/sync/index.js
```diff
@@ -154,17 +150,13 @@ export const getFetchQueriesEpic = (resolveRoutes, fetchQueriesFn) => {
...
-				// clean up path for use as endpoint URL
-				matched.pop().match.path.replace(/[^a-z0-9/]/gi, '')
-			)
-			.then(apiPath => {
-				// construct the fetch call using match.path
-				const fetchUrl = `${config.apiUrl}${apiPath}`;
-				const fetchQueries = fetchQueriesFn(fetchUrl, (self || {}).value);
-				return fetchQueries(queries, meta); // call fetch
-			})
+		const matched = findMatches(location);
```
eilinora: Nice, much easier to reason through

packages/mwp-core/src/renderers/browser-render.jsx
```diff
@@ -16,20 +16,28 @@ type AppProps = {
...
-	return resolveRoutes(window.location).then(() => ({
+
+	// find the currently-matched routes that have components that need to
+	// be resolved, then resolve them - this doesn't actually change the route
+	// object, but loads the app chunks into memory
+	const matchedAsyncComponentGetters = findMatches(window.location)
+		.map(({ getComponent }) => getComponent)
+		.filter(x => x);
+
+	return Promise.all(matchedAsyncComponentGetters).then(() => ({
```
eilinora: To make sure i understand, in above we are loading all the chunks into memory and then once all done, actually "resolve" the route now that we have all the pieces. Correct?

mmcgahan: mostly correct - the 'route' itself is already resolved, but the `getComponent` Promise-returning functions will still be called when the route is rendered by `AsyncRoute`. Since the component modules themselves have already been downloaded, however, the Promises should return instantly.

We actually might want to adjust this behavior to just start rendering immediately, and let the components pop into place as they download - TTI would improve significantly, but it might be a little jarring from a UX perspective. Worth checking out at some point... similar to a FOUC, but maybe not as bad?

eilinora: Yeah maybe but not sure how much its worth if the page feels odd. Our FOM (flash of mobile) already feels awkward. :) I think a change like that we should definitely launch behind a LDF so we can easily test the experience for some users internally :)

packages/mwp-router/src/AsyncRoute.jsx
mmcgahan: Yeah I want to see it in action as well. There are a few things that should mitigate the likelihood of that kind of layout thrashing, however.

1. _Most_ of the time people will be navigating from a view that has already loaded the 'wrapper' component. Not always, but most of the time.
2. When someone first loads the app on a particular page, the client will have pre-loaded all the matching components, so the layout shouldn't thrash in that case.
3. The service worker will greedily download all components for _unmatched_ routes when the app first loads, so by the time someone navigates within the app, it's very likely that the new route will already have access to all the components it needs.

Still need to see it in action to confirm, however. I'm just realizing that there might be some issue with the initial render of the client-side app not matching the server markup, which might require some tweaks to avoid warnings, but we'll see.

eilinora: sounds good to me :)

packages/mwp-core/src/renderers/browser-render.jsx
mmcgahan: I've updated this to avoid the flash - it was bad.

PR #477, opened 2018-04-12
Pure resolve routes

mmcgahan:
This thing needs tessssts


flow-typed/platform.js
```diff
@@ -121,8 +121,9 @@ declare type QueryResponse = {
 
 declare type QueryFunction = (location: { [string]: mixed }) => Query;
 
-declare type AsyncPlatformRoute = {|
-	getComponent: () => Promise<React$ComponentType<*>>,
+declare type PlatformRoute = {|
```
mmcgahan: unwinding a separation I made earlier between an 'AsyncPlatformRoute' and a 'StaticPlatformRoute' - it was messing with types too much and is a little hard to handle with so many resolvers - I'll be able to restore it once `getIndexRoute` and `getNestedRoutes` are gone.

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -203,125 +203,130 @@ const makeRenderer = (
 	scripts: Array<string> = [],
 	enableServiceWorker: boolean,
 	cssLinks: ?Array<string>
-) => (request: HapiRequest, reply: HapiReply): Promise<RenderResult> => {
-	middleware = middleware || [];
+) => {
+	const routesPromise = resolveAllRoutes(routes);
+	return (request: HapiRequest, reply: HapiReply): Promise<RenderResult> => {
```
mmcgahan: new setup created an indentation change https://github.com/meetup/meetup-web-platform/pull/477/files?w=1 to see the diff without whitespace changes

packages/mwp-router/src/util/resolve.js
```diff
@@ -42,30 +42,39 @@ export const getChildRoutes = (
...
 };
 
-const _resolveNestedRoutes: Resolver<MatchedRoute> = matchedRoute =>
-	matchedRoute.route.getNestedRoutes
-		? matchedRoute.route
-				.getNestedRoutes()
-				.then(routes => (matchedRoute.route.routes = routes))
-				.then(() => matchedRoute)
-		: Promise.resolve(matchedRoute);
+const _resolveNestedRoutes: Resolver<PlatformRoute> = ({
```
mmcgahan: these functions were refactored to be 'pure' and deal only with routes, not 'matchedroutes' - they return a _new route object_ - the original route is not modified like it was previously

packages/mwp-router/src/util/resolve.js
```diff
@@ -138,3 +135,23 @@ export const getRouteResolver = (
...
+			resolveAllRoutes(routes || []).then((routes: Array<PlatformRoute>) =>
+				Object.freeze({
+					...noRoutesRoute,
+					routes,
+				})
+			)
+		);
+};
+
+export const resolveAllRoutes = (
```
mmcgahan: TODO: document these functions. They are server-side only because they "over-resolve" - i.e. they resolve all the Promises in the routing tree, not just the ones that correspond to a supplied route like `getRouteResolver` does - going to be reworking that in a new PR

PR #479, opened 2018-04-16
upgrade react 16.3

sadafie:
`yarn bootstrap` gets stuck. investigating.


mmcgahan:
I think this could also use a bump in the MWC dependencies, but probably not a big deal - I would consider dropping support for React 15 entirely at this point, at least for mwp-app-render components, because that's the most likely package to start using React 16-only features, and now seems like a good moment to ratchet forward

sadafie:
I'll update MWC components here.

> I would consider dropping support for React 15 entirely at this point

I can bump the major version & drop support but it will be a breaking change for pro-web. They're still on React 15.

mmcgahan:
Yeah I think a breaking change might preferable here - fine to leave as-is, but it will be up to the next person who starts using 16.x APIs to make sure that the dependency list is also updated.

mmcgahan:
You might want to check for an update to `enzyme-adapter-react-16` as well - not sure if that also tends to get bumped with react updates, but worth trying.

mmcgahan:
Looks like Yarn 1.6.0 is out - might want to try that out to see if it fixes the bootstrap errors

sadafie:
> Looks like Yarn 1.6.0 is out - might want to try that out to see if it fixes the bootstrap errors

I did. Same errors sadly. The yarn workspace ticket in on my to-do so hopefully we can bump this back up soon.

PR #481, opened 2018-04-17
WP 688 patch up flowtypes, particularly for mwp-app-render

mmcgahan:
general flow type update to keep things running smoothly.

Also, update to redux 4.0 because why not


mmcgahan:
SV, will merge when PR build passes

PR #483, opened 2018-04-19
WP 679 `query.list` param to aggregate API list responses

mmcgahan:
Docs to come


flow-typed/platform.js
```diff
@@ -97,16 +97,29 @@ declare type FluxStandardAction = {
...
 	params?: Params,
 	type?: string,
+	mockResponse?: () => any,
 	meta?: {
 		flags?: Array<string>,
-		method?: string,
+		method?: 'get' | 'post' | 'delete' | 'patch',
 		noCache?: boolean,
+		metaRequestHeaders?: Array<string>,
+		variants?: {
```
mmcgahan: these are just some params that are supported but weren't part of the type def

packages/mwp-api-state/src/reducer.js
```diff
@@ -25,11 +25,48 @@ export const filterKeys: ObjectFilter = (
...
-	[resp.response.ref]: { ...resp.response, query: resp.query },
+type ResponseStateSetter = (
+	state: ApiState,
+	resp: QueryState
+) => { [string]: QueryResponse };
+
+/*
+ * Queries can specify 'list' response handling - response values will be
+ * stored in a `dyanmicRef` in state. If a `list.merge` param is specified,
+ * the response will be merged with the existing list state stored in `dynamicRef`
```
jaclynj: nice clear comment 👍

packages/mwp-api-state/src/reducer.js
```diff
@@ -25,11 +25,48 @@ export const filterKeys: ObjectFilter = (
...
+	}
+
+	// do some smart merging
+	const { idTest, sort } = merge;
+	// remove anything in old list that is part of new list
+	const oldList = (state[dynamicRef] || {}).value
+		.filter(valOld => !newList.find(valNew => idTest(valOld, valNew)));
+
+	// combine the new list and the old list and sort the results
+	return { [dynamicRef]: [...oldList, ...newList].sort(sort) };
```
jaclynj: Should we also handle removing an item from the list or should that be handled in the consumer application? 

If a user pages through items and starts deleting them - is it still the responsibility of the application to remove those items from the list? If so the consumer would need to continue to maintain its own list.

mmcgahan: this is a really good question - I haven't really put much effort into handling add/edit/delete in this PR. I talked with @eilinora a little bit about it, and we're thinking that the current implementation is not ideal but it should be sufficient because it entirely re-sets the list when a user navigates away from the page and then comes back, and we are expecting that to happen whenever a user edits or deletes an item.

If you wanted to implement 'inline delete', that would need to be handled in local React state - I'm not 100% sure what that would look like, but we'll probably just need to see it in action.

Longer term, I'm hoping to update the query system in such a way that add/edit/delete requests automatically update any corresponding object that exists in state, which will reduce the burden on the app to maintain API data in React state.

PR #488, opened 2018-05-03
check for __disable_sw in query to override service worker setting

mmcgahan:
This will allow profiling requests unencumbered by precache requests

### DIFF
--- a/packages/mwp-core/src/renderers/server-render.jsx
+++ b/packages/mwp-core/src/renderers/server-render.jsx
@@ -239,7 +239,7 @@ const makeRenderer = (
 
 		// service worker can be disabled per-request with __disable_sw querystring param
 		const enableServiceWorkerReq =
-			'__disable_sw' in request.query || enableServiceWorker;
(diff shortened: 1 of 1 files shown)


packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -237,13 +237,17 @@ const makeRenderer = (
 		const userAgent = headers['user-agent'];
 		const userAgentDevice = headers['x-ua-device'] || ''; // set by fastly
 
+		// service worker can be disabled per-request with __disable_sw querystring param
+		const enableServiceWorkerReq =
+			'__disable_sw' in request.query || enableServiceWorker;
```
sergeychernyshev: Looks to me like having `__disable_sw` in a query string is in fact enabling service worker, setting `enableServiceWorkerReq` to `true`. Should probably be negated and `&&`-ed with config parameter.

mmcgahan: :facepalm:

PR #490, opened 2018-05-08
use Array instead of Set for IE11 support

akdetrick:
fixes https://meetup.atlassian.net/browse/SDS-612

IE11 does not support `Set`. This meant that server render was working fine, but client render in IE11 resulted in empty style strings.

We _could_ polyfill `Set` instead, but we rarely use that JS feature and will likely drop support for IE11 soon anyway. This is a quick fix until we drop support for this browser.

### DIFF
--- a/packages/mwp-app-render/src/components/StyleWith.jsx
+++ b/packages/mwp-app-render/src/components/StyleWith.jsx
@@ -26,16 +26,16 @@ import Helmet from 'react-helmet';
 const StyleWith = props => {
 	const { styles, children } = props;
 
-	const moduleCSS = new Set();
+	const moduleCSS = [];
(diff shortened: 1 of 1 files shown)


mmcgahan:
@akdetrick I'm curious about this ticket - MDN indicates that IE11 should support the previous implementation in this ticket - there's no stack trace in the ticket or the PR, but I'm interested in what error was being thrown so that we can watch out for it in other places where we're using Set currently.

akdetrick:
CSS strings were not being added to the `Set` using the `add` method in IE11, which caused rendering of empty style tags. The troubling thing here is that we were not seeing any JS errors; just empty style tags.

PR #491, opened 2018-05-09
Yarn Workspaces

sadafie:
- Use yarn workspaces
- Resolve yarn check errors by upgrading dependencies across packages
- Enable yarn 1.6 in travis build

Using workspaces you no longer need to run `lerna bootstrap`. `lerna bootstrap --use-workspaces` just calls `yarn install` at the root, so the only necessary command is `yarn install`. 

See this issues for details: 
https://github.com/lerna/lerna/issues/1001#issuecomment-370612630 
https://github.com/lerna/lerna/issues/1301


packages/mwp-logger-plugin/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "homepage": "https://github.com/meetup/meetup-web-platform#readme",
   "dependencies": {
-    "@google-cloud/logging-bunyan": "0.7.0",
+    "@google-cloud/logging-bunyan": "0.8.1",
```
mmcgahan: turns out this is a breaking change - I can update the code, but I'm curious what the motivation for the version bump was

PR #492, opened 2018-05-12
adding isQLd prop to config object

eilinora:
Adds `isQLd` to config object in state, used to help with an indicator on client UI to donote that the user is QL'd  as another member 

See: https://github.com/meetup/mup-web/pull/3296
for usage

### DIFF
--- a/flow-typed/platform.js
+++ b/flow-typed/platform.js
@@ -17,6 +17,7 @@ declare type MWPState = {
 		requestLanguage: string,
 		supportedLangs: Array<string>,
 		initialNow: number,
+		isQLd: boolean,
 		variants: mixed,
 		entryPath: string,
 		media: MatchMedia,
--- a/flow-typed/platform.js
+++ b/flow-typed/platform.js
@@ -17,7 +17,7 @@ declare type MWPState = {
 		requestLanguage: string,
 		supportedLangs: Array<string>,
 		initialNow: number,
-		isQLd: boolean,
+		isQL: boolean,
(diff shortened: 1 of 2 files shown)


packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -247,6 +248,7 @@ const makeRenderer = (
 					requestLanguage,
 					supportedLangs,
 					initialNow: new Date().getTime(),
+					isQLd: parseMemberCookie(state)['ql'] === 'true',
```
mmcgahan: this should be `.ql` not `['ql']`.

flow-typed/platform.js
```diff
@@ -17,6 +17,7 @@ declare type MWPState = {
 		requestLanguage: string,
 		supportedLangs: Array<string>,
 		initialNow: number,
+		isQLd: boolean,
```
mmcgahan: The acronym-as-past-participle-verb is sort of hard to read here - I'm wondering about just `isQL` (just to eliminate confusion about what the trailing `d` means) or `isImpersonate` - "impersonate" is the more industry-general term, but unfortunately it's almost never used at Meetup so it actually might create more confusion, but "QL" is so... opaque.

eilinora: Agreed, i think anything other than QL would be confusing as its to engrained in our code / culture

eilinora:
@mmcgahan all updated :)

PR #495, opened 2018-05-23
Added Stylelint config

mperrotti:
These new configuration files reflect styling rules and best practices Adam and I discussed.

`.stylelintrc` usage in consumer apps would look like this:
```
{
	'extends': [
		'mwp-config/stylelint/stylelint-config-default', 
		'example-thridparty-config',
		'./example-local-config'
	],
	'rules': {
		...any rule overrides that repo needs...
	}
}
```


packages/mwp-config/stylelint/stylelint-config-default.js
```diff
@@ -0,0 +1,206 @@
...
+		'declaration-colon-space-after': 'always',
+		'declaration-colon-space-before': 'never',
+		'declaration-empty-line-before': 'never',
+		'declaration-property-unit-whitelist': {
+			'line-height': [],
+			'/^animation/': ['ms'],
+			'/^transition/': ['ms'],
+		},
+		'declaration-property-value-blacklist': {
+			outline: ['outline'],
```
akdetrick: should this be `outline: ['none']`?

packages/mwp-config/stylelint/stylelint-config-default.js
```diff
@@ -0,0 +1,206 @@
...
+		'property-case': 'lower',
+		'property-no-unknown': true,
+		'property-no-vendor-prefix': true,
+		'rule-empty-line-before': 'always',
+		'selector-attribute-brackets-space-inside': 'never',
+		'selector-attribute-operator-blacklist': '|=',
+		'selector-attribute-operator-space-after': 'never',
+		'selector-attribute-operator-space-before': 'never',
+		'selector-attribute-quotes': 'always',
+		'selector-class-pattern': [
```
akdetrick: Do we need to set a level for this?

akdetrick: Alternatively, we could pull this out of shared rules and export a separate, stricter set for use in SDS projects:

```
const SDSFrameworkConfig { rules: { 'selector-class-pattern': ... } }
```

packages/mwp-config/stylelint/stylelint-config-default.js
```diff
@@ -0,0 +1,206 @@
...
+		'string-quotes': 'double',
+		'time-min-milliseconds': 100,
+		'unit-case': 'lower',
+		'unit-no-unknown': true,
+		'value-keyword-case': 'lower',
+		'value-list-max-empty-lines': 0,
+		'value-no-vendor-prefix': true,
+
+		// the following rules are intentionally disabled
+		'at-rule-blacklist': null,
```
akdetrick: dumb question - do the rules included in `stylelint` or `stylelint-scss` default to `null`? If so, we may not need to explicitly declare them here.

mperrotti: They default to null, but I kind of like being explicit about what we've disabled. Happy to remove if you prefer not having them :)

packages/mwp-config/stylelint/stylelint-config-default.js
mperrotti: Severity level:
Going to come back to this after I fix all error/warnings in MWC

Only including in SDS projects:
Totally makes sense

packages/mwp-config/stylelint/stylelint-config-default.js
akdetrick: The reason I can think of to keep these around is to document their absence as being intentional, but that doesn't provide much context around _why_ we intend not to use these.

I'd vote to remove them to fall back on default `null` values for rules we don't want.

akdetrick:
Some nitpicks about the naming for `stylelint-config-default.js` and `stylelint-config-scss.js`... 

"default" implies that the `scss` config somehow overrides "default". What do you think of making it `stylelint-config-base`?

This may also be a good time to introduce `stylelint-config-swarmProject.js` which could extend `stylelint-config-base` and add some extra, strict rules (like the PCV naming convention validator)

mperrotti:
Very into those suggestions. I'll update the PR :)

packages/mwp-config/stylelint/stylelint-config-swarmProject.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+	rules: {
```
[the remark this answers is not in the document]
mperrotti: I had that at first, but thought it might be better to explicitly extend `stylelint-config-base` _and_ `stylelint-config-swarmProject` in SDS projects. Happy to update though :)

mmcgahan:
this is creating a strange `yarn check` error when being installed in mup-web:

```
error "stylelint#postcss-html#postcss-syntax@^0.10.0" doesn't satisfy found match of "postcss-syntax@0.9.1"
error "stylelint#postcss-markdown#postcss-syntax@^0.10.0" doesn't satisfy found match of "postcss-syntax@0.9.1"
```

can be fixed by installing `postcss-syntax@0.10.0`, but it's weird.

PR #500, opened 2018-06-11
Setting UserTiming mark and measure to show up in dev tools

sergeychernyshev:
This makes TTI metric that we measure visible in developer tools (Chrome DevTools, WebPageTest/SpeedCurve, etc.) through W3C UserTiming API.

Both, mark and measure with the name `meetup-tti` are set as well as custom `console.timeStamp()` supported by Chrome and Firefox.

### DIFF
--- a/packages/mwp-app-render/src/components/PageWrap/index.jsx
+++ b/packages/mwp-app-render/src/components/PageWrap/index.jsx
@@ -25,7 +25,9 @@ class PageWrap extends React.Component {
 	}
 
 	componentDidMount() {
-		// Browser has now rendered client-side application - fire the browser TTI trigger
+		// Browser has now rendered client-side application - fire the browser TTI triggers
+
+		// Fire TTI trigger to be sent to NewRelic
 		if (window.newrelic) {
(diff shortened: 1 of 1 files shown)


packages/mwp-app-render/src/components/PageWrap/index.jsx
```diff
@@ -41,6 +43,21 @@ class PageWrap extends React.Component {
...
+			window.performance &&
+			window.performance.mark &&
+			window.performance.measure
+		) {
+			window.performance.mark('meetup-tti');
+			window.performance.measure('meetup-tti', 'navigationStart', 'meetup-tti');
+		}
+
+		// Specially for Developer Tools in the browsers (Chrome & Firefox), create entry for the event so it shows up on Performance timeline
+		if (console && console.timeStamp) {
```
mmcgahan: no need to change, but we should be able to rely on `console` being globally available everywhere.

sergeychernyshev: Got it.

PR #506, opened 2018-06-20
update csrf plugin, use new header and cookie name

mmcgahan:
This will provide a clean break from the 'old' CSRF system - we will use a new csrf header name and a new csrf cookie name so that there are no conflicts.

I think this update will need to launch in pro-web and mup-web simultaneously since pro-web is using the mup-web API server


packages/mwp-app-server/src/util/getPlugins.js
```diff
@@ -17,20 +17,23 @@ import apiProxyPlugin from 'mwp-api-proxy-plugin';
  * @module ServerPlugins
  */
 
+const CSRF_COOKIE_NAME = 'x-mwp-csrf';
```
mmcgahan: followup will be to move this CSRF plugin code into a standalone plugin instead of managing it inside `mwp-app-server`

packages/mwp-app-server/src/util/getPlugins.js
```diff
@@ -142,7 +145,10 @@ export default function getPlugins({ languageRenderers }) {
 		getApiProxyPlugin(),
 		getLanguagePlugin(),
 		getLogger(),
-		getCsrfPlugin(),
+		getCsrfPlugin({
```
eilinora: why are we passing this?

eilinora: in `getCsrfPlugin` its not expecting any params unless i'm looking at the wrong function

jaclynj: good q

mmcgahan: thanks, missed that - updated.

mmcgahan:
p.s. to all reviewers - I've had on my todo list to build out integration tests for the platform but it's a reasonably large project and I haven't had time, so that's why testing is pretty weak here.

PR #507, opened 2018-06-20
`url` and `referrer` should include querystring

mmcgahan:


mmcgahan:
@eilinora yeah this won't affect any data inside the app itself - it just reads info from HTTP requests to the app server and decides which values should be treated as the 'referrer' and which should be considered the 'url' of the page being requested - purely a consumer.

eilinora:
Cool thanks for validating

PR #508, opened 2018-06-20
pass state through to getMatchedQueries, document update

mmcgahan:
lots of annoying whitespace changes in this PR - I'll highlight changes but might end up making this SV since the changes are pretty small and simple and non-breaking

### DIFF
--- a/packages/mwp-api-state/src/sync/index.js
+++ b/packages/mwp-api-state/src/sync/index.js
@@ -76,9 +76,9 @@ export const getNavEpic = findMatches => (action, store) => {
 	const cacheAction = requestMetadata.logout && { type: 'CACHE_CLEAR' };
 
 	const previousQueries = referrer.pathname
-		? getMatchedQueries(referrer)(findMatches(referrer))
+		? getMatchedQueries(referrer, state)(findMatches(referrer))
 		: [];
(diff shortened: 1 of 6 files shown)


packages/mwp-api-state/src/sync/index.js
```diff
@@ -76,9 +76,9 @@ export const getNavEpic = findMatches => (action, store) => {
 	const cacheAction = requestMetadata.logout && { type: 'CACHE_CLEAR' };
 
 	const previousQueries = referrer.pathname
-		? getMatchedQueries(referrer)(findMatches(referrer))
+		? getMatchedQueries(referrer, state)(findMatches(referrer))
```
mmcgahan: new arg!

packages/mwp-router/src/util/query.js
```diff
@@ -37,7 +37,7 @@ const _matchedRouteQueriesReducer = (location: URL) => (
 	// call the query functions with non-url-encoded params
 	const params = decodeParams(match.params);
 	const routeQueries = routeQueryFns
-		.map(queryFn => queryFn({ ...match, location, params }))
+		.map(queryFn => queryFn({ ...match, location, params }, state))
```
mmcgahan: new arg actually gets passed to `queryFn`

PR #510, opened 2018-06-28
WP-742: As a Meetup developer, I want to have access to state.flags on server load

burrata:


packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -279,15 +279,18 @@ const makeRenderer = (
...
+				// state.flags is available to query calls
 				if (checkReady(store.getState())) {
 					addFlags(store).then(() => {
 						resolve(store);
 					});
 					return;
 				}
+
+				// dispatch SERVER_RENDER to kick off API middleware
+				store.dispatch({ type: SERVER_RENDER, payload: url });
```
burrata: Isn't is possible that `checkReady` would not be true and that `SERVER_RENDER` might still dispatch before `addFlags`? I think we may just have to move this beneath `resolve(store` in the `addFlags` resolution, unless there's an explicit reason we should not or don't have to.

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -277,26 +277,29 @@ const makeRenderer = (
...
 		const checkReady = state =>
 			state.preRenderChecklist.every(isReady => isReady);
-		const populateStore = store =>
-			new Promise((resolve, reject) => {
-				// dispatch SERVER_RENDER to kick off API middleware
-				store.dispatch({ type: SERVER_RENDER, payload: url });
+		const populateStore = store => {
+			// keep this above the SERVER_RENDER so
+			// state.flags is available to query calls
+			if (checkReady(store.getState())) {
```
mmcgahan: `checkReady` is only used _after_ `SERVER_RENDER` is dispatched - the store can't be ready before any API calls are made.

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -277,26 +277,29 @@ const makeRenderer = (
...
-						resolve(store);
-					});
-					return;
-				}
-				const unsubscribe = store.subscribe(() => {
-					if (checkReady(store.getState())) {
-						addFlags(store).then(() => {
-							resolve(store);
-							unsubscribe();
+					new Promise((resolve, reject) => {
```
mmcgahan: this will have to be `return`ed in order to prevent the outer Promise from resolving

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -277,26 +277,25 @@ const makeRenderer = (
...
+			return addFlags(store).then(() => {
 				// dispatch SERVER_RENDER to kick off API middleware
 				store.dispatch({ type: SERVER_RENDER, payload: url });
 
-				if (checkReady(store.getState())) {
-					addFlags(store).then(() => {
-						resolve(store);
+				return new Promise((resolve, reject) => {
+					// dispatch SERVER_RENDER to kick off API middleware
+					store.dispatch({ type: SERVER_RENDER, payload: url });
```
mmcgahan: this will be a duplicate call to `store.dispatch` - you've already triggered the action on line 283

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -277,26 +277,25 @@ const makeRenderer = (
...
+					const unsubscribe = store.subscribe(() => {
```
mmcgahan: before setting up the `store.subscribe()`, it's important to check whether the store is already ready (synchronously) so that you can `resolve` immediately, similar to what was happening before on line 285.

If it's not ready synchronously, then you set up the `store.subscribe()` listener to `resolve` async.

packages/mwp-core/src/renderers/server-render.jsx
burrata: My lesson today is that no change is ever "simple." :)

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -277,26 +277,30 @@ const makeRenderer = (
 		};
 		const checkReady = state =>
 			state.preRenderChecklist.every(isReady => isReady);
-		const populateStore = store =>
-			new Promise((resolve, reject) => {
+		const populateStore = store => {
+			return addFlags(store).then(() => {
```
mmcgahan: nitpick, skip the curly brackets and `return`

burrata: made this change but then forgot to push 😢

PR #512, opened 2018-06-28
WP-742: Fall back on member cookie for id when api.self is not yet available

burrata:
Add fallback for member id from member cookie when api.self is not loaded yet before adding flags, otherwise we will make route queries with the wrong initial flags.

### DIFF
--- a/packages/mwp-core/src/renderers/server-render.jsx
+++ b/packages/mwp-core/src/renderers/server-render.jsx
@@ -265,7 +265,14 @@ const makeRenderer = (
 
 		// otherwise render using the API and React router
 		const addFlags = store => {
-			const memberObj = (store.getState().api.self || {}).value || {};
+			// use the api self object first if it exists,
+			// else, use the member id on the member cookie,
+			// else, this person's a guest and they'll get a
+			// default id of 0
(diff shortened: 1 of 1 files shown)


packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -265,7 +265,14 @@ const makeRenderer = (
 
 		// otherwise render using the API and React router
 		const addFlags = store => {
-			const memberObj = (store.getState().api.self || {}).value || {};
+			// use the api self object first if it exists,
+			// else, use the member id on the member cookie,
+			// else, this person's a guest and they'll get a
+			// default id of 0
+			const memberObj = (store.getState().api.self || {}).value || {
```
mmcgahan: ah, I remember this issue now and why we had to wait until the API calls had happened. Many feature flags are targeted at email addresses, so sending only the member ID won't work. We have a catch-22 here, where you need the full member object from the API in order to select the right flags, and you need to right flags in order to make the right API calls :/

burrata:
Closing this based on the issue above. Going to make a new PR that adds comments for reasoning for future devs.

PR #521, opened 2018-07-27
WP-760 handle list which is under `value` field in response

andeliv:
Adjust getListState in `mwp-api-atate` to correctly handle list which is under `value` field in response:
https://meetup.atlassian.net/wiki/spaces/CP/pages/679346519/RFC+Pageable+Resources

### DIFF
--- a/packages/mwp-api-state/src/reducer.js
+++ b/packages/mwp-api-state/src/reducer.js
@@ -45,7 +45,13 @@ export const getListState: ResponseStateSetter = (
 	}
 	const { dynamicRef, merge } = query.list;
 	// this query should be treated as a list-building query
-	const newList = response.value instanceof Array ? response.value : [];
+	// list can be either a root query result (response.value) or be under `value` field of the result (response.value.value)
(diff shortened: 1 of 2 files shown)


packages/mwp-api-state/src/reducer.js
```diff
@@ -45,7 +45,13 @@ export const getListState: ResponseStateSetter = (
 	}
 	const { dynamicRef, merge } = query.list;
 	// this query should be treated as a list-building query
-	const newList = response.value instanceof Array ? response.value : [];
+	// list can be either a root query result (response.value) or be under `value` field of the result (response.value.value)
+	let newList = [];
+	if (response.value instanceof Array) {
+		newList = response.value;
+	} else if (response.value && response.value.value instanceof Array) {
+		newList = response.value.value;
```
eilinora: For meta are we going to read from the original response body?

andeliv: Yep, I suggest to read is from original response

PR #523, opened 2018-08-03
add the member id pre store population

browne0:


packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -263,7 +264,19 @@ const makeRenderer = (
 		// addFlags _must_ be called after the store is 'ready' to ensure that
 		// there is a full member object available in state - feature flags can
 		// be selected based on member id, email, and other properties
-		const addFlags = populatedStore => {
+		const addFlags = (populatedStore, member?: { id: number }) => {
```
mmcgahan: prefer strict and explicit interfaces in platform code - the fewer 'forms' we have to support, the more better.

In this case, the second call can extract the `api.self.value` before the function is called - feel free to write a HOC for that case.

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -284,14 +285,16 @@ const makeRenderer = (
 				store.dispatch({ type: SERVER_RENDER, payload: url });
 
 				if (checkReady(store.getState())) {
-					addFlags(store).then(() => {
+					const memberObj = (store.getState().api.self || {}).value || {};
```
sadafie: First time looking at this code, but can't you just define memberObj once on line 286?

PR #525, opened 2018-08-09
Make sort callback optional for List state

andeliv:
Make sort callback optional as e.g. in members search default sort is `closest_match` which we decided not to implement on FE side

### DIFF
--- a/packages/mwp-api-state/src/reducer.js
+++ b/packages/mwp-api-state/src/reducer.js
@@ -63,12 +63,12 @@ export const getListState: ResponseStateSetter = (
 	const oldList = ((state[dynamicRef] || {}).value || [])
 		.filter(valOld => !newList.find(valNew => idTest(valOld, valNew)));
 
+	// combine the new list and the old list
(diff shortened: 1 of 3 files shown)


packages/mwp-api-state/src/reducer.js
```diff
@@ -63,8 +63,13 @@ export const getListState: ResponseStateSetter = (
 	const oldList = ((state[dynamicRef] || {}).value || [])
 		.filter(valOld => !newList.find(valNew => idTest(valOld, valNew)));
 
+	const mergedList = [...oldList, ...newList];
+	// we can omit sort callback e.g. in members search default sort is `closest_match` which we decided not to implement on FE side
```
mmcgahan: I like to keep product details out of the platform code, so a comment like `'sort callback is optional'` would be fine, although the place that should be declared is [in the type declaration itself](https://github.com/meetup/meetup-web-platform/blob/master/flow-typed/platform.js#L59)

PR #526, opened 2018-08-10
Wp 745 bad token in dev

jaclynj:
Jira ticket: https://meetup.atlassian.net/browse/WP-745

- Add dev csrf token flow so it doesn't conflict with production environment cookies
- Update comments with more detail for future understandability

### DIFF
--- a/packages/mwp-app-server/src/util/getPlugins.js
+++ b/packages/mwp-app-server/src/util/getPlugins.js
@@ -135,6 +135,7 @@ export default function getPlugins({ languageRenderers }) {
 	const { package: { agent }, getServer } = config;
 	const server = getServer();
 	const isProdApi = server.properties.api.isProd;
+	const isProdEnv = server.properties.isProd;
 
 	return [
 		getAppRoutePlugin({ languageRenderers }),
@@ -142,8 +143,10 @@ export default function getPlugins({ languageRenderers }) {
 		languagePlugin,
 		getLoggerPlugin(),
 		getCsrfPlugin({
-			headerName: isProdApi ? CSRF_HEADER_NAME : DEV_CSRF_HEADER_NAME,
-			cookieName: isProdApi ? CSRF_COOKIE_NAME : DEV_CSRF_COOKIE_NAME,
+			// Use prod cookies in production environment and dev cookies in
+			// dev environment regardless of whether you're using the prod api
+			headerName: isProdEnv ? CSRF_HEADER_NAME : DEV_CSRF_HEADER_NAME,
+			cookieName: isProdEnv ? CSRF_COOKIE_NAME : DEV_CSRF_COOKIE_NAME,
 		}),
 		requestAuthPlugin,
 		getActivityTrackingPlugin({ agent, isProdApi }),
--- a/packages/mwp-app-server/src/util/getPlugins.js
+++ b/packages/mwp-app-server/src/util/getPlugins.js
@@ -16,25 +16,16 @@ import { plugin as requestAuthPlugin } from 'mwp-auth-plugin';
  *
  * @module ServerPlugins
  */
-
-const CSRF_COOKIE_NAME = 'x-mwp-csrf';
+const CSRF_COOKIE_NAME =
+	process.env.NODE_ENV === 'production' ? 'x-mwp-csrf' : 'x-mwp-csrf_dev';
 const CSRF_HEADER_COOKIE_NAME = `${CSRF_COOKIE_NAME}-header`;
 const CSRF_HEADER_NAME = CSRF_COOKIE_NAME;
-// dev cookie names
-const DEV_CSRF_COOKIE_NAME = `${CSRF_COOKIE_NAME}-dev`;
-const DEV_CSRF_HEADER_COOKIE_NAME = `${DEV_CSRF_COOKIE_NAME}-header`;
-const DEV_CSRF_HEADER_NAME = DEV_CSRF_COOKIE_NAME;
 
 // Sets the csrf header token from plugin into a cookie
 export function setCsrfCookies(request, h) {
-	const { headers = {} } = request.response;
-	const csrfHeader = headers[CSRF_COOKIE_NAME];
-	const csrfDevHeader = headers[DEV_CSRF_COOKIE_NAME];
+	const csrfHeader = (request.response.headers || {})[CSRF_COOKIE_NAME];
 	if (csrfHeader) {
 		h.state(CSRF_HEADER_COOKIE_NAME, csrfHeader);
-	}
-	if (csrfDevHeader) {
-		h.state(DEV_CSRF_HEADER_COOKIE_NAME, csrfDevHeader);
 	}
 	return h.continue;
 }
@@ -52,8 +43,7 @@ export function setCsrfCookies(request, h) {
  * so that it syncs across browser tabs.
  *
  * We set similar but different cookie names for dev and prod environments so the prod
- * cookies are not read in the dev environment, which was causing BAD_TOKEN errors
- * because we were comparing dev and prod tokens which would never match.
+ * cookies are not read in the dev environment.
  * 
  * In order to ensure that both cookie values have parallel settings, this
  * function calls `server.state` for both cookie names before registering the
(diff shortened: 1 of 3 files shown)


packages/mwp-app-server/src/util/getPlugins.js
```diff
@@ -20,23 +20,41 @@ import { plugin as requestAuthPlugin } from 'mwp-auth-plugin';
...
+ * The CSRF plugin we use, 'electrode-csrf-jwt', generates a token for each request 
+ * that we make and sets the token in an HTTP-only cookie (CSRF_COOKIE_NAME) and in
+ * the HTTP response header (also CSRF_COOKIE_NAME). In non-GET requests we must supply
+ * the latest generated token from the response header as an HTTP request header - the
+ * plugin will compare the cookie token and the request header token and return a
+ * BAD_TOKEN error if they do not match.
+ * 
+ * We updated this flow to set the token from the response header as a cookie 
+ * (CSRF_COOKIE_NAME-header) and use the cookie value to set the request header
+ * so that it syncs across browser tabs.
```
jaclynj: I'm not sure if I was testing this right, @mmcgahan @sadafie would you have any insight? I tried opening two different tabs and submitting and it works fine, but I'm not sure if that's all it takes to test this.

sadafie: i think it's two different tabs and/or have both the prod and dev  `x-mwp-csrf-header` cookies

packages/mwp-app-server/src/util/getPlugins.js
```diff
@@ -20,23 +20,41 @@ import { plugin as requestAuthPlugin } from 'mwp-auth-plugin';
...
+const DEV_CSRF_COOKIE_NAME = `${CSRF_COOKIE_NAME}-dev`;
+const DEV_CSRF_HEADER_COOKIE_NAME = `${DEV_CSRF_COOKIE_NAME}-header`;
+const DEV_CSRF_HEADER_NAME = DEV_CSRF_COOKIE_NAME;
 
+// Sets the csrf header token from plugin into a cookie
 export function setCsrfCookies(request, h) {
-	const csrfHeader = (request.response.headers || {})[CSRF_COOKIE_NAME];
+	const { headers = {} } = request.response;
+	const csrfHeader = headers[CSRF_COOKIE_NAME];
+	const csrfDevHeader = headers[DEV_CSRF_COOKIE_NAME];
```
mmcgahan: This function should only manage the cookie that is relevant to the environment, not both prod and dev - similar to the pattern we use for handling member cookies: https://github.com/meetup/meetup-web-platform/blob/master/packages/mwp-api-proxy-plugin/src/util/send.js#L26 - read/determine the cookie(s) of interest in one place and then continue execution without side effects on the other environment.

For these cookies, you can just choose the cookie based on `NODE_ENV === 'production'` rather than `api.isProd`

EDIT: see comment about assigning the `const`s initially to avoid any special logic here.

packages/mwp-app-server/src/util/getPlugins.js
```diff
@@ -20,23 +20,41 @@ import { plugin as requestAuthPlugin } from 'mwp-auth-plugin';
...
 /**
- * The CSRF plugin we use - 'electrode-csrf-jwt' compares a cookie token to a
- * header token in non-GET requests. By default, it will set the cookie token
- * itself (CSRF_COOKIE_NAME), and supply the corresponding header token in a custom
- * header (also CSRF_COOKIE_NAME). However, we update this flow to also supply the
- * header token as a cookie (CSRF_COOKIE_NAME-header) so that it syncs across
- * browser tabs.
+ * The CSRF plugin we use, 'electrode-csrf-jwt', generates a token for each request 
+ * that we make and sets the token in an HTTP-only cookie (CSRF_COOKIE_NAME) and in
+ * the HTTP response header (also CSRF_COOKIE_NAME). In non-GET requests we must supply
```
mmcgahan: I don't think the plugin makes any determination of which request method it will apply to (GET vs other) - it's up to the routes to opt in/out of CSRF protection.

packages/mwp-app-server/src/util/getPlugins.js
mmcgahan: I think the description and your test is good.

packages/mwp-app-server/src/util/getPlugins.js
```diff
@@ -20,23 +20,41 @@ import { plugin as requestAuthPlugin } from 'mwp-auth-plugin';
...
+ * plugin will compare the cookie token and the request header token and return a
+ * BAD_TOKEN error if they do not match.
+ * 
+ * We updated this flow to set the token from the response header as a cookie 
+ * (CSRF_COOKIE_NAME-header) and use the cookie value to set the request header
+ * so that it syncs across browser tabs.
  *
+ * We set similar but different cookie names for dev and prod environments so the prod
+ * cookies are not read in the dev environment, which was causing BAD_TOKEN errors
+ * because we were comparing dev and prod tokens which would never match.
```
mmcgahan: I don't think you need to describe the defect that was fixed, just specify the current behavior so that future devs know that this is an intentional parallel implementation.

packages/mwp-app-server/src/util/getPlugins.js
```diff
@@ -58,14 +77,17 @@ export function getCsrfPlugin(electrodeOptions) {
 			{ secret: server.settings.app.csrf_secret },
 			electrodeOptions
 		);
-
+		const cookie_name = isProd ? CSRF_COOKIE_NAME : DEV_CSRF_COOKIE_NAME;
```
mmcgahan: camelCase?

packages/mwp-app-server/src/util/getPlugins.js
```diff
@@ -20,23 +20,41 @@ import { plugin as requestAuthPlugin } from 'mwp-auth-plugin';
 const CSRF_COOKIE_NAME = 'x-mwp-csrf';
 const CSRF_HEADER_COOKIE_NAME = `${CSRF_COOKIE_NAME}-header`;
 const CSRF_HEADER_NAME = CSRF_COOKIE_NAME;
+// dev cookie names
+const DEV_CSRF_COOKIE_NAME = `${CSRF_COOKIE_NAME}-dev`;
+const DEV_CSRF_HEADER_COOKIE_NAME = `${DEV_CSRF_COOKIE_NAME}-header`;
```
mmcgahan: Thinking about this a bit more - why not just keep all the implementation logic the same, and just make the initial assignment of `CSRF_COOKIE_NAME` and `CSRF_HEADER_COOKIE_NAME` match the environment? i.e.

```js
const CSRF_COOKIE_NAME = process.env.NODE_ENV === 'production'
  ? 'x-mwp-csrf'
  : 'x-mwp-csrf_DEV';
const CSRF_HEADER_COOKIE_NAME = process.env.NODE_ENV === 'production'
  ? 'x-mwp-csrf-header'
  : 'x-mwp-csrf-header_DEV';
```

I think if you set up the values that way, you don't have to do any more juggling of values in the rest of the module.

packages/mwp-app-server/src/util/getPlugins.js
jaclynj: It actually does check the request method - https://github.com/meetup/electrode-csrf-jwt/blob/master/lib/csrf.js#L60
Or you can specify a 'skipVerify' function as an option, but the default behavior when using this plugin is to protect POST/PATCH/PUT/DELETE requests

packages/mwp-app-server/src/util/getPlugins.js
```diff
@@ -16,11 +16,12 @@ import { plugin as requestAuthPlugin } from 'mwp-auth-plugin';
  *
  * @module ServerPlugins
  */
-
-const CSRF_COOKIE_NAME = 'x-mwp-csrf';
+const CSRF_COOKIE_NAME =
+	process.env.NODE_ENV === 'production' ? 'x-mwp-csrf' : 'x-mwp-csrf_dev';
```
jaclynj: @mmcgahan you suggested `_DEV` (in caps) but it compares the cookie + header in all lowercase so when this is _DEV it will always fail the csrf token comparison. I didn't dig too deeply to figure out exactly where it's changing to lowercase, but figured having `_DEV` in caps for the cookie/header name isn't a dealbreaker

mmcgahan: ah ok yeah I think Hapi converts all header names to lowercase, so that makes sense!

PR #527, opened 2018-08-13
upgrade electrode-csrf-jwt

sadafie:


sadafie:
> feel free to SV updates that are just version bumps :)

Yep! @jaclynj and I both wanted to do one more round of testing to makes sure the changes worked as expected, and they do!

PR #529, opened 2018-08-14
Upgrade dependencies to fix yarn warnings

sadafie:


package.json
```diff
@@ -64,6 +65,7 @@
     "iron": "4.0.5",
     "jest": "23.1.0",
     "joi": "13.4.0",
+    "js-joda": "^1.8.0",
```
sadafie: This is an interesting case. Currently `meetup-web-components` has several peerDependencies that cause `yarn install` warnings in `mwp-*` packages because they are missing. Example:

```
warning " > meetup-web-components@5.2.2434" has unmet peer dependency "classnames@^2.2.5".
warning " > meetup-web-components@5.2.2434" has unmet peer dependency "js-joda@>1.8.0".
warning " > meetup-web-components@5.2.2434" has unmet peer dependency "swarm-animation@^0.0.95".
warning " > meetup-web-components@5.2.2434" has incorrect peer dependency "swarm-icons@^3.4.334".
warning " > meetup-web-components@5.2.2434" has unmet peer dependency "swarm-sasstools@^5.0.412 || ^4.2.408".
```
There are two options I can see in handling this:
1) The option shown in this PR -- add all the peer dependencies to `package.json` files of all `mwp-*` packages that depend on `meetup-web-components`.
2) Another option is to remove the peerDependences in `meetup-web-components` and make them dependencies. However i'm not familiar with the decision as the why they are peerDependencies to begin with.

mmcgahan: peer dependencies help avoid having multiple versions of a particular package built into the app - e.g. if MWP uses js-joda and specifies v1.8.0 as a dependency, and then some mup-web code wants to use js-joda and so installs version 1.8.2, _both_ versions will be bundled by webpack into the client script.

Specifying a peerDependency is the same as saying "I will use whatever you have installed that matches this version range" - same thing for React and other packages that get directly imported in the platform and in the apps where we want to avoid double-bundling.

mmcgahan: that said, we might want to find a way to reduce the peer dependencies for MWC... there are a lot

sadafie: Interesting. The versioning scenario makes sense.

I always read peerDependencies are requirements as well. Is that not the case? Or do we not want that to be the case?

mmcgahan: they're requirements in order to avoid the double-bundling issue, but it's always a judgement call based on

1. how likely is it that the consumer will need to use the same dependency itself?
   - not likely: define it as a direct dependency
   - very likely: probably good to make it a peer dependency so that it will always be shared
2. how big is the dependency?
   - small: not a major problem to double-bundle
   - large: make every effort to avoid double-bundling

So, 'peerDependencies' is a "guaranteed-safe' way of sharing a dependency, but it creates some logistical/maintenance burden... I don't think there's a good bright-line test on whether/when to use it

sadafie:
We're down to only **3** warnings on a clean install: `yarn lerna clean; yarn` gives the following output:

```
sadaf: ~/git_repo/meetup-web-platform [feature/upgrade_deps_yarn_warnings] [!] $ yarn
yarn install v1.9.4
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
warning gitbook-plugin-github@2.0.0: The engine "gitbook" appears to be invalid.
[4/5] 🔗  Linking dependencies...
warning "meetup-web-components > file-loader@1.1.11" has unmet peer dependency "webpack@^2.0.0 || ^3.0.0 || ^4.0.0".
warning "workspace-aggregator-e765a262-7e71-4a0c-a418-3743fc114289 > mwp-app-route-plugin@0.0.1" has unmet peer dependency "electrode-csrf-jwt@git+https://github.com/meetup/electrode-csrf-jwt.git#793431729d8fb4ee315f28709f323f198b25d460".
[5/5] 📃  Building fresh packages...
✨  Done in 8.69s
```

The last warning about electrode-csrf-jwt can be ignored. yarn complains because of the hash commit.

PR #534, opened 2018-08-23
add blankie plugin + CSP options

jaclynj:
This is still in progress but I wanted to get it in prelim for feedback

### DIFF
--- a/packages/mwp-app-server/src/util/getPlugins.js
+++ b/packages/mwp-app-server/src/util/getPlugins.js
@@ -1,5 +1,7 @@
 import Inert from 'inert';
 import CsrfPlugin from 'electrode-csrf-jwt/lib/csrf-hapi17';
+import CspPlugin from 'blankie';
+import ScooterPlugin from 'scooter'; // blankie dependency
 
 import config from 'mwp-config';
 import { plugin as loggerPlugin } from 'mwp-logger-plugin';
@@ -91,6 +93,20 @@ export function getCsrfPlugin(electrodeOptions) {
 		},
 	};
 }
+/**
+ * We are using Blankie as our Content Security Policy (CSP) plugin 
+ * Which uses Scooter to detect the user agent and apply the appropriate
+ * CSP header, usually Content-Security-Policy but some older browsers are
+ * slightly different. A CSP compatible browser will use the header to ignore
+ * scripts not whitelisted in our policy header.
+ * https://github.com/nlf/blankie 
+ */
+export function getCspPlugin(options) {
+	return {
+		plugin: CspPlugin,
+		options,
+	};
+}
 
 export function getAppRoutePlugin(options) {
 	return {
@@ -131,6 +147,16 @@ export default function getPlugins({ languageRenderers }) {
 		getCsrfPlugin({
 			headerName: CSRF_HEADER_NAME,
 			cookieName: CSRF_COOKIE_NAME,
+		}),
+		ScooterPlugin, // csp plugin (blankie) dependency
+		getCspPlugin({
+			defaultSrc: "'self' *.meetup.com beta2.dev.meetup.com:8001",
+			connectSrc: '*',
+			frameSrc: 'https://staticxx.facebook.com',
+			imgSrc: '*',
+			styleSrc: "* 'unsafe-inline'",
+			scriptSrc: "* 'unsafe-inline' 'unsafe-eval'",
+			generateNonces: 'false',
 		}),
 		requestAuthPlugin,
 		getActivityTrackingPlugin({ agent, isProdApi }),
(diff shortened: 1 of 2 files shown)


packages/mwp-app-server/src/util/getPlugins.js
```diff
@@ -132,6 +148,16 @@ export default function getPlugins({ languageRenderers }) {
...
 		}),
+		ScooterPlugin, // csp plugin (blankie) dependency
+		getCspPlugin({
+			defaultSrc: "'self' *.meetup.com beta2.dev.meetup.com:8001",
+			connectSrc: '*',
+			frameSrc: 'https://staticxx.facebook.com',
+			imgSrc: '*',
+			styleSrc: "* 'unsafe-inline'",
+			scriptSrc: "* 'unsafe-inline' 'unsafe-eval'",
+			generateNonces: 'false',
```
jaclynj: These are pretty permissive. I can probably find env variables for the urls

packages/mwp-app-server/src/util/getPlugins.js
```diff
@@ -91,6 +93,20 @@ export function getCsrfPlugin(electrodeOptions) {
...
+/**
+ * We are using Blankie as our Content Security Policy (CSP) plugin 
+ * Which uses Scooter to detect the user agent and apply the appropriate
+ * CSP header, usually Content-Security-Policy but some older browsers are
+ * slightly different. A CSP compatible browser will use the header to ignore
+ * scripts not whitelisted in our policy header.
+ * https://github.com/nlf/blankie 
+ */
+export function getCspPlugin(options) {
+	return {
```
mmcgahan: Looking good so far. In order to add the remaining headers, you'll need to register an `onPreResponse` handler that can call [`request.response.header(name, value)`](https://hapijs.com/api#-responseheadername-value-options)

This essentially means [writing a plugin](https://hapijs.com/tutorials/plugins?lang=en_US) with a `register` function that does something like

```js
server.ext('onPreResponse', (request, h) => {
  request.response.header('Foo', 'Bar');
  request.response.header('Baz', 'Qux');
  ...
});
```

Consider whether you would like to combine this with 'blankie' internally, rather than registering it as a separate plugin, e.g. in the plugin's `register` method, you could have something like

```js
export function register(server: HapiServer, options: ?Object) {
  server.ext('onPreResponse', (request, h) => {
    request.response.header('Foo', 'Bar');
    request.response.header('Baz', 'Qux');
    ...
  });
  return blankie.register(server, options);
```

That might not be a good idea, but something to think about.

packages/mwp-app-server/src/util/getPlugins.js
mmcgahan: permissive is probably fine for now. For the dev url, you'll probably need to get the subdomain from packageConfig:

```js
import { env, package as packageConfig } from 'mwp-config';
const { subdomain } = packageConfig;
const { asset_server } = env;

...

defaultSrc = ['self', '*.meetup.com', `${subdomain}.meetup.com:${asset_server.port}`].join(' '),
```

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## [17.2]
+
+Create mwp-csp-plugin which adds some useful security headers
+
+## [17.1]
```
jaclynj: This hasn't been merged yet but I'll wait until it is

.travis.yml
```diff
@@ -26,6 +26,7 @@ before_install:
   - echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" > packages/mwp-core/.npmrc
   - echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" > packages/mwp-config/.npmrc
   - echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" > packages/mwp-consumer/.npmrc
+  - echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" > packages/mwp-csp-plugin/.npmrc
```
jaclynj: I think this is right according to the Readme. Is there anything anything I might be missing?

mmcgahan: nope, that should be all you need 👍

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
...
+
+Create mwp-csp-plugin which adds some useful security headers
```
mmcgahan: just for consistency, `-  **New feature** 'mwp-csp-plugin'...`

packages/mwp-csp-plugin/src/README.md
```diff
@@ -0,0 +1,8 @@
+# Content Security Policy
+
+Sets the following headers:
+- Strict-Transport-Security: Tells browser that it should only be accessed using HTTPS & how long to remember that
```
mmcgahan: format header name as bold or code block?

PR #536, opened 2018-08-27
Add ux-capture.min.js to Dom.jsx

sadafie:


packages/mwp-app-render/src/components/Dom.jsx
```diff
@@ -72,13 +73,18 @@ const DOM = props => {
...
+	const uxCaptureJS = fs.readFileSync(uxCaptureFilename, 'utf-8');
+
 	return (
 		<html lang={htmlLang}>
 			<head>
 				{head.title.toComponent()}
 				{head.meta.toComponent()}
 				{head.link.toComponent()}
 				{head.script.toComponent()}
+				{uxCaptureJS &&
```
mmcgahan: don't think you need this conditional - the file is required to be there

sadafie: removed. should we handle errors returned by `fs.readFileSync` ?

PR #541, opened 2018-08-29
Enforce node and yarn versions

sadafie:
If you're on the wrong version of node, you'll get:

```bash
sadaf: ~/git_repo/meetup-web-platform [WP-781_engine_versions] [!] $ yarn
yarn install v1.9.4
[1/5] 🔍  Validating package.json...
error meetup-web-platform@0.0.1: The engine "node" is incompatible with this module. Expected version "^8.11.4".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

If you're on the wrong version of yarn, you'll get:

```bash
sadaf: ~/git_repo/meetup-web-platform [WP-781_engine_versions] $ yarn
yarn install v1.8.0
[1/5] 🔍  Validating package.json...
error meetup-web-platform@0.0.1: The engine "yarn" is incompatible with this module. Expected version "1.9.4".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

### DIFF
--- a/packages/mwp-tracking-plugin/package.json
+++ b/packages/mwp-tracking-plugin/package.json
@@ -4,7 +4,8 @@
   "description": "Data collection for MWP apps, including activity and click tracking",
   "engineStrict": true,
   "engines": {
-    "node": ">=8.1.0"
+    "node": "^8.11.4",
+    "yarn": "1.9.4"
   },
   "main": "./lib/index.js",
   "scripts": {
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: node_js
 node_js:
-  - "8.9"
+  - "8.11"
 cache:
   directories:
     - node_modules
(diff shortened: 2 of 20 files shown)


packages/mwp-tracking-plugin/package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Data collection for MWP apps, including activity and click tracking",
   "engineStrict": true,
   "engines": {
-    "node": ">=8.1.0"
+    "node": "^8.11.4",
+    "yarn": "1.9.4"
```
mmcgahan: do you think it would be good to be a little looser on the yarn version? e.g. `^1.9.4`, or do you feel like it's pretty fragile?

sadafie: I was considering that. It's probably safe to allow patch updates. I was hoping to keep everyone on the same version as travis uses, but maybe that's not as necessary.

PR #549, opened 2018-09-17
Mp 2157 upgrade nr agent

chenrui333:


packages/mwp-api-proxy-plugin/package.json
```diff
@@ -34,11 +34,12 @@
     "ajv": "5.5.2",
     "joi": "13.4.0",
     "mwp-auth-plugin": ">=0.0.1",
-    "mwp-logger-plugin": ">=0.0.1",
     "mwp-config": ">=0.0.1",
+    "mwp-logger-plugin": ">=0.0.1",
+    "newrelic": "^4.8.1",
```
sadafie: is there a reason we're adding newrelic as a dependency for mwp-api-proxy-plugin?

chenrui333: yeah, I think I had a warning when I tried to upgrade the main `package.json`.

chenrui333: This is what I have got from `pro-web` side
```
warning "mwp-app-server > mwp-api-proxy-plugin@17.1.2646" has incorrect peer dependency "newrelic@^3.3.1".
```

sadafie: ah, okay, i see. you don't need to add a new dependency here. if you look up at about line 30 in this file, you'll see:

```
   "peerDependencies": {
     "newrelic": "^3.3.1",
     "rxjs": "^5.4.3"
   },
```

Just update the peerDep version there.

chenrui333:
@sadafie Thanks for pointing out the right place! I have also updated the `peerDep` for mwp-core

chenrui333:
Thanks @sadafie @daniel-hayes , merging now!

PR #550, opened 2018-09-18
Move UX Capture React components to mwp-app-render

sadafie:
These need to be shared across repos, so moving out of mup-web.

**No new code**


sadafie:
No new code - SV

PR #552, opened 2018-09-19
add member state to initial redux config

browne0:
Original tickets that caused need:

https://meetup.atlassian.net/browse/MG-2473
https://meetup.atlassian.net/browse/MG-2441

For segmentation and rollout purposes, it is extremely advantageous to have the member object available pre query to do split tests based off of things like the member's id or location. 

If the member is not logged in, this will return an empty object with a member id of 0.


packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -246,6 +246,7 @@ const makeRenderer = (
 					isProdApi: server.settings.app.api.isProd,
 					isQL: parseMemberCookie(state).ql === 'true',
 					memberId: parseMemberCookie(state).id,
+					memberState: parseMemberCookie(state).state,
```
mmcgahan: is this literally the state of the country that the member is in? With no PR description and no ticket, we don't have context for understanding the purpose of adding this property, and I have concerns about the maintainability of incrementally adding member cookie properties over time. We should consider loading all member cookie params into a `member` object here.

Please document!

browne0: Absolutely I'll document some stuff and pass it in via a member object. 

It is the state of the country the member is in for reference.

It's a longer discussion but i'll try giving a gist in a comment.

eilinora: I would also suggest renaming this, I would have initially thought at a first glance this referred to the state of the member vs members physical location

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -245,7 +245,20 @@ const makeRenderer = (
...
 					isQL: parseMemberCookie(state).ql === 'true',
-					memberId: parseMemberCookie(state).id,
+					/**
+					 * Original tickets that caused need:
+					 * 
+					 * @see https://meetup.atlassian.net/browse/MG-2473
+					 * @see https://meetup.atlassian.net/browse/MG-2441
+					 * 
+					 * For segmentation and rollout purposes, it is extremely advantageous
+					 * to have the member object available pre query to do split tests
```
mmcgahan: the member cookie is not the same as the member object returned by `/members/self`, so be careful if you are relying on it having the same properties in any downstream code

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -245,7 +245,20 @@ const makeRenderer = (
...
+					 * based off of things like the member's id or location.
+					 * 
+					 * If the member is not logged in, this will return an empty object
```
mmcgahan: no need to document what `parseMemberCookie` does here. Also, just nitpick, `{ id: 0 }` is not an 'empty object'.

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -245,7 +245,20 @@ const makeRenderer = (
...
+					 * with a member id of 0.
+					 */
+					member: parseMemberCookie(state),
```
mmcgahan: if you're removing `memberId`, then this PR is introducing a breaking change - I recommend leaving `memberId` as a standalone param for now, and potentially adding a `// deprecated - use member.id` comment to indicate that it should not be referenced. Since you're defining a new 'correct' place to look for `member.id`, make sure you look for and update places where `config.memberId` is currently being used elsewhere in the platform (e.g. https://github.com/meetup/meetup-web-platform/blob/master/packages/mwp-api-state/src/cache/index.js#L33)

You'll also want to update [this line](https://github.com/meetup/mup-web/blob/master/src/app/root/appSelectors.js#L103) in mup-web

Finally, you'll need to add a new type for `config.member` to the `MWPState` type: https://github.com/meetup/meetup-web-platform/blob/master/flow-typed/platform.js#L10

flow-typed/platform.js
```diff
@@ -16,6 +16,24 @@ declare type MWPState = {
 		enableServiceWorker: boolean,
 		requestLanguage: string,
 		supportedLangs: Array<string>,
+		member: {
```
mmcgahan: Probably good to make this an intersection of exact types, with one of the 'constituent types' being `{| id: 0 |}`

flow-typed/platform.js
```diff
@@ -16,6 +16,24 @@ declare type MWPState = {
...
+			id: number,
+			status: number,
+			timestamp: number,
+			bs: number,
+			// time zone
+			tz: string,
+			zip: number,
+			country: string,
+			city: string,
+			state: string,
```
mmcgahan: I would guess that `state` is not populated for all members, so is this param optional, i.e. `state?: string`?

Are there other optional keys?

flow-typed/platform.js
```diff
@@ -16,6 +16,24 @@ declare type MWPState = {
...
+			lat: number,
+			lon: number,
+			ql: boolean,
+			scope: string,
+			rem: number,
```
mmcgahan: code comment on what this is would be nice

flow-typed/platform.js
```diff
@@ -16,6 +16,24 @@ declare type MWPState = {
...
+			org: number,
```
mmcgahan: can this be any number, or is it an enum like `0 | 1 | 2` or something?

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## [17.2]
+- **NEW FEATURE** Add whole member cookie to initial redux configuration on server
```
mmcgahan: I'd like to be more specific here - we're not adding the actual cookie, which is a string, but the object encoded by the cookie.

The deprecation notice should probably be called out in it's own bullet point, e.g.

```md
- **DEPRECATED** `state.config.memberId` should not be used. Reference `state.config.member.id`
   instead, which is guaranteed to be populated
```

flow-typed/platform.js
browne0: its 1 or its not there

flow-typed/platform.js
```diff
@@ -16,6 +16,27 @@ declare type MWPState = {
 		enableServiceWorker: boolean,
 		requestLanguage: string,
 		supportedLangs: Array<string>,
+		member:
+			| {|
+					id: number,
+					status: number,
+					timestamp: number,
+					bs: number,
```
jaclynj: whats `bs`?

flow-typed/platform.js
```diff
@@ -16,6 +16,27 @@ declare type MWPState = {
...
+					country: string,
+					city: string,
+					state?: string,
+					lat: number,
+					lon: number,
+					ql: boolean,
+					scope?: string,
+					// whether the member selected remember me or not
+					rem?: number,
+					org?: 1,
```
jaclynj: what is org?

browne0: whether the member is an organizer or not. added a comment.

browne0: it'll be `org=1` or org won't be provided if not an org

PR #553, opened 2018-09-19
UXCaptureEventMark component

sadafie:
Fixes https://meetup.atlassian.net/browse/WP-802

Create a component that can take a child element and add the proper event UX Capture mark.

For example:

```jsx
<UXCaptureEventMark
     browserEvent="onLoad"
     mark="ux-image-onload-logo"
>
          <img src="/logo" />
</UXCaptureEventMark>
```

will output:

```html
<img src="/logo" onLoad="if(window.UX){ UX.mark('ux-image-onload-logo'); }" />
```

### Question

- Is the `browserEvent` mark prop necessary? Will it always be `onLoad`?


sadafie:
See https://github.com/meetup/meetup-web-components/pull/638 for real usage example.

sergeychernyshev:
It looks like you're trying to create abstract components which might give user wrong impression that they can fire any events at will on any element.

In reality, the fact that `` tag requires only two marks to be fired is not as obvious as it sounds at a first glance and was devised after ruling out quite a few different events - see [Souders' article](https://www.stevesouders.com/blog/2015/05/12/hero-image-custom-metrics/) where he documents it, scroll down to "Image Custom Metrics" section to see specific of how he picked the best method.

This means that 3 methods documented (foreground image, text without a font, handler) have a particular configuration of events and `` in particular requires a pair of specific events.

I wonder if
```jsx
<UXCaptureImage src="meetup_logo.jpg" alt="Meetup logo" label="logo"/>
```
would be better api to generate 

```html
<img  src="meetup_logo.jpg" alt="Meetup logo" onLoad="if(window.UX){ UX.mark('ux-image-onload-logo'); }" />
<script>if(window.UX){ UX.mark('ux-image-inline-logo'); }</script>
```
to enforce two things:
- metric naming pattern that includes "ux-" prefix and "image" method and "onload" & "inline" sub-events
- but also to enforce both events from the pair are included
- actual element is `` and not any other

Alternatively, it can be a generic image component (that can include more image management abstractions) with optional `UXCaptureLabel` prop that will trigger inclusion of the events and calculate marks:
```jsx
<Image src="meetup_logo.jpg" alt="Meetup logo" UXCaptureLabel="logo"/>
```

One downside of using partial mark names / labels / slugs that I see - it's harder to reference partials against the array of expected marks and might require another abstraction that does similar abstraction for `UX.expect()` calls like:
```javascript
UX.expect([
{
    label: "ux-destination-verified",
    marks: [UXImageMarkLabel('logo'), UXTextMarkLabel('group-title')]
}, {
    label: "ux-primary-action-available",
    marks: [UXTextMarkLabel('rsvp-label'), UXHandlerLabel('rsvp-click-handler')]
}
])
```

P.S. after writing it all up, I realized that I am "running ahead of the steam engine" (as we say it in Russia) and we probably should not discuss proper abstractions until we are further in this project. So it might be worth moving these deep abstract React API conversations into Confluence rather than keeping it in GitHub PRs and slack threads.

packages/mwp-app-render/src/components/uxcapture/UXCaptureEventMark.jsx
```diff
@@ -0,0 +1,22 @@
+// @flow
+import * as React from 'react';
+
+type Props = {
+	browserEvent: string,
```
mmcgahan: are all browser events allowed? I think I've only seen reference to `onLoad`, in which case `browserEvent` doesn't need to be passed in.

packages/mwp-app-render/src/components/uxcapture/UXCaptureEventMark.jsx
```diff
@@ -0,0 +1,22 @@
...
+	mark: string,
+	children: React$Element<*>,
```
mmcgahan: Is there a flow type that only allows HTML React elements, or only components that have an `onLoad` prop? We would like to ensure that `onLoad` is handled when injected.

packages/mwp-app-render/src/components/uxcapture/UXCaptureEventMark.jsx
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * takes only a *single* child element
+ * and injects event based UX.mark() call into the child
+ *
+ * @see example https://github.com/meetup/ux-capture#image-elements
+ */
+export default ({ browserEvent, mark, children }: Props) =>
+	React.cloneElement(children, {
+		[browserEvent]: generateUXCaptureJS(mark),
```
mmcgahan: this will also override any existing `[browserEvent]` props in the component, so I think we probably should be a bit more careful about 'merging' with an existing property if there is one, or providing a wrapper.

mmcgahan: One thing I was wondering is whether we could wrap the child in a `<div>` and set the `onLoad` event to trigger on _that_ rather than the (presumably ``) child - I'm not sure, but I think it's possible that a wrapper's `onLoad` will not fire until all children have loaded, so

```html
<div onload="...">
  <img ...>
</div>
```

might provide the same timings as

```html
<img onload="..." />
```

sergeychernyshev: I don't know if we should rely on `onLoad` event even works on `<div>`s

sergeychernyshev: I'd rather avoid nesting extra layers just for this.

packages/mwp-app-render/src/components/uxcapture/UXCaptureEventMark.jsx
sadafie: Yeah, that was my question in the description too :)

mmcgahan: let's just go with always using `onLoad`, then

packages/mwp-app-render/src/components/uxcapture/UXCaptureEventMark.jsx
sadafie: Admittedly I had to look this up, but Google says only these tags support `onload`:

```<body>
<iframe>
<img>
<input>
<link>
<script>
<style>```

I'll see if I can make this flow type stricter.

File: packages/mwp-app-render/src/components/uxcapture/UXCaptureEventMark.jsx (hunk 3 above)
Comment (mmcgahan) [reply packages/mwp-app-render/src/components/uxcapture/UXCaptureEventMark.jsx]: it would be relatively easy to test, and if it works, it would make the interface much more composable at the cost of an extra level of HTML indentation. The major benefit of that approach is that it keeps the implementation 'pure' - the child component is completely unchanged, which is better for predictability (e.g. no concerns about interacting with/overriding an existing `onLoad` prop.

IMO and extra level of HTML indentation is a pretty low cost to pay for predictability

File: packages/mwp-app-render/src/components/uxcapture/UXCaptureEventMark.jsx (hunk 2 above)
Comment (mmcgahan) [reply packages/mwp-app-render/src/components/uxcapture/UXCaptureEventMark.jsx]: ah, boo, so the `<div>` wrapper idea won't work. Okay then - just need to figure out how best to merge our onLoad with an existing onLoad prop

File: packages/mwp-app-render/src/components/uxcapture/UXCaptureEventMark.jsx (hunk 3 above)
Comment (sergeychernyshev) [reply packages/mwp-app-render/src/components/uxcapture/UXCaptureEventMark.jsx]: It looks like div onload just doesn't work so it has to be on `` HTML tag.

If test-ability is important, maybe custom `<Image>` JSX tag is a better approach?

Comment (mmcgahan) [reply packages/mwp-app-render/src/components/uxcapture/UXCaptureEventMark.jsx]: Yeah I like the idea of a dedicated `<Image>` component - we've talked about setting up something like that to encapsulate some more responsive-friendly behavior as well. If we don't want to get into that kind of structure in this PR and save it for a followup, that's fine with me.

File: packages/mwp-app-render/src/components/uxcapture/UXCaptureEventMark.jsx [hunk 4]
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * takes only a *single* child element
+ * and injects event based UX.mark() call into the child
+ *
+ * @see example https://github.com/meetup/ux-capture#image-elements
+ */
+export default ({ mark, children }: Props) => {
+	// if child has on onLoad prop
+	// don't do anything
+	if (children.props.onLoad) {
```
hsbacot: How come we want to ignore onLoad rather than merging it here?

sadafie: Path of least resistance? :)   TBH, I'm not sure any of us fully understand other instrumentation use cases yet to be able to predict implementation details. As Mike mentioned, we may way to make this an image specific tag (i.e., `UXCaptureImageEventMark`) and or other transformations.

sadafie: Oops, submitted that too soon. Until we have a use case where we need to support an image tag with an additional onLoad prop, I just chose to ignore it. I could merge here too, but not sure if it's necessary.

PR #554, opened 2018-09-20
Convert from Redux to Cookie-based click tracking

mmcgahan:
This PR moves click tracking history tracking from Redux state into cookies - instead of accumulating click data in Redux state and then writing a cookie that is sent to the BE, the cookie itself holds state over the course of the session, and gets sent automatically with every request instead of using the API middleware to conditionally send it.

The logic around reading and resetting the cookie has been moved to the server - the application routes have to opt-in to consuming the cookie and then clearing it.

This should allow us to capture clicks that correspond to navigation away from mup-web or opening new mup-web pages in a different tab, because the cookie will persist.


packages/mwp-api-proxy-plugin/src/routes.js
```diff
@@ -26,6 +27,18 @@ const getApiProxyRoutes = path => {
 				'electrode-csrf-jwt': {
 					enabled: true,
 				},
+				'mwp-tracking-plugin': {
+					click: request => {
+						// only consume click data when the querystring indicates that it should be consumed
+						// this allows the client to avoid invoking click parsing for lazy-loaded API requests
+						if (request.query.metadata) {
```
sadafie: Should this be `if (!request.query.metadata)`?

packages/mwp-api-proxy-plugin/src/routes.js
```diff
@@ -26,6 +27,18 @@ const getApiProxyRoutes = path => {
 				'electrode-csrf-jwt': {
 					enabled: true,
 				},
+				'mwp-tracking-plugin': {
+					click: request => {
+						// only consume click data when the querystring indicates that it should be consumed
+						// this allows the client to avoid invoking click parsing for lazy-loaded API requests
```
sadafie: This is a nitpick but I found this comment a bit confusing. I'd probably just remove that second line.

packages/mwp-app-route-plugin/src/route.js
```diff
@@ -50,6 +50,9 @@ export default (languageRenderers: { [string]: LanguageRenderer }) => ({
 			'electrode-csrf-jwt': {
 				enabled: true, // need to generate tokens on page request
 			},
+			'mwp-tracking-plugin': {
+				click: () => true, // always track clicks on server render
```
jaclynj: I can't find where/how this `click` function is used, where can I see that?

mmcgahan: The server plugin uses this `route.settings.plugins` value to determine whether or not the cookie should be consumed: https://github.com/meetup/meetup-web-platform/pull/554/files#diff-15492b88113fa366a51ac60655d9e077R14

I've also added docs in the README that should make it easier for future devs to follow

mmcgahan:
@sadafie 

> Are there new test cases you could write for the changes in click.js and click middleware?
Done, with integration tests using Hapi's `server.inject`

> I found the description of the changes in the PR description helpful. It's a bit lengthy for code comments, but can we update/add to a README
Done

> Does this need a CHANGELOG + minor/major version bump?
yes, done.

> Unrelated, I'm forgetting where the click tracking click actions get dispatched. I'm assuming that's existing code, but could you point me to it please?

It was sort of buried, so I put it in a more 'expected' place in `clickStore.js`: https://github.com/meetup/meetup-web-platform/blob/WP-793_click-track-cookie/packages/mwp-tracking-plugin/src/util/clickStore.js#L21

packages/mwp-tracking-plugin/README.md
```diff
@@ -68,23 +69,35 @@ with parallel behavior implemented in Meetup Classic.
...
+	config: {
+		plugins: {
+			'mwp-tracking-plugin': {
+				// only track clicks when `?trackClicks=true` in querystring
+				click: request => Boolean(request.query.trackClicks),
+			},
+		},
+	},
+};
+```
```
jaclynj: Nice. This is very helpful!

packages/mwp-tracking-plugin/src/click.test.js
```diff
@@ -0,0 +1,99 @@
...
+describe('onPreHandlerExtension', () => {
+	// try {
+	// 	const pluginSettings =
+	// 		request.route.settings.plugins[CLICK_PLUGIN_NAME] || {};
+	// 	if (pluginSettings.click && pluginSettings.click(request)) {
+	// 		return clickReader(request, h);
+	// 	}
+	// } catch (err) {
+	// 	request.server.app.logger.error({ err, context: request, ...request.raw });
+	// }
```
jaclynj: Does this comment need to be here?

mmcgahan: ah, no, was using it as a reference to set up tests. will clean up!

PR #556, opened 2018-09-25
move click tracking entirely out of redux

mmcgahan:
The connection between click tracking and Redux is tenuous at best, and makes it difficult to follow the data.

This refactor simplifies the behavior to be independent of redux, and solely provide an interface for activating click tracking in the browser - a small update to `mwp-app-render/BrowserApp` replaces the click middleware+enhancer setup that existed previously.


mmcgahan:
need to SV to check this out in mup-web

PR #559, opened 2018-09-27
`UXCaptureImageLoad` component

sadafie:
Replaces `UXCaptureEventMark` which was having problems with `React.cloneElement()`.


packages/mwp-app-render/src/components/uxcapture/UXCaptureImageLoad.jsx
```diff
@@ -0,0 +1,29 @@
+// @flow
+import * as React from 'react';
+
+type Props = {
```
mmcgahan: better to extend the built-in `HTMLImageElement` props:

```jsx
type Props = React$ElementConfig(HTMLImageElement) & { mark: string };
```

packages/mwp-app-render/src/components/uxcapture/UXCaptureImageLoad.jsx
```diff
@@ -0,0 +1,29 @@
...
+ * and a UXCapture `onLoad` handler
+ *
+ * @see example https://github.com/meetup/ux-capture#image-elements
+ */
+const UXCaptureImageLoad = (props: Props) => {
+	const { mark, src, ...other } = props;
+
+	// TODO: handle merging of onLoad prop
+
+	const onLoadCallback = () => {
```
mmcgahan: by convention, prefer keeping callback names parallel with their HTML equivalents - use `onLoad` instead of `onLoadCallback` for this.

mmcgahan: `onLoad` will fire with an `event` argument - you can type it with `ev: SyntheticEvent<*>`

packages/mwp-app-render/src/components/uxcapture/UXCaptureImageLoad.jsx
```diff
@@ -0,0 +1,29 @@
...
+		if (window.UX) {
+			window.UX.mark(mark);
+		}
+	};
```
mmcgahan: Looks like you've left a `TODO` comment, but I think it's pretty straightforward to make sure that you don't silently suppress any existing `onLoad` prop that was passed in

```js
if (props.onLoad) {
  props.onLoad(ev);
}
```

mmcgahan:
Left a comment in Slack, but it looks like we're going to need more `dangerouslySetInnerHTML` here

All event handlers in React are handled internally as ‘SyntheticEvents’ - when you add an event handler prop to a component, React adds that to its internal event system, e.g. any `onClick` you add to a button doesn’t actually render the `<button onclick="..." />` in the HTML.

In order to get server-rendered `onload` HTML attributes, we’re going to need to use `dangerouslySetInnerHTML`

https://reactjs.org/docs/handling-events.html

The downside is that the `onload` handler will be called twice - once when the HTML is parsed, and again when the client app hydrates and effectively re-renders the page.

this works, but `loaded` gets printed twice:

```jsx
<div
  dangerouslySetInnerHTML={{
    __html: `
      <img src="http://placekitten.com/200/300" onload="console.log('loaded')" />
      <script>console.log('loading')</script>
    `,
  }}
/>
```

In order to get that markup, we'll probably need to use template strings to effectively convert the props into actual attributes. This might be a little tricky because things like `aria` attributes will be specified in camelCase, but will have to be hyphenated in the markup (e.g. `ariaLabel="foo"` becomes `aria-label="foo"`). I don't think it will be terribly complicated, but it's something we'll have to account for.

I pretty confident there's no other way around the issue.

sadafie:
Left a few comments in slack as well, but Sergey made a point about dedup-ing the marks within this component, and I've been finding that challenging. 

In theory, we want to use `dangerouslySetInnerHTML` to get the `onload` HTML attribute to render on the `` tag on server render. Once we are confident that the image has loaded and before any hydrates/re-renders... we either 1) remove the onload event listener or 2) use a flag (state?) to test if the listener has been triggered before and simply do nothing. 

Perhaps I'm not understanding the React lifecycles as accurately as I should be and there's an api interface I'm missing, but I played around with this and have had no luck. Also, what we theoretically need to do seems unattainable just by the nature of networks and browsers.

At this point in time, this PR is in a non-working state.

@mmcgahan I've sent a pairing invite for Monday if you have time!

packages/mwp-app-render/src/components/uxcapture/UXCaptureImageLoad.jsx
```diff
@@ -0,0 +1,66 @@
...
+			// Any other special props needed? Add here.
+			// @see: https://reactjs.org/docs/dom-elements.html
+			if (prop === 'className') {
+				attribute = 'class';
+			}
+			if (prop === 'htmlFor') {
+				attribute = 'for';
+			} else {
+				// handles camelCased attributes like `onLoad`
+				attribute = attribute.toLowerCase();
```
mmcgahan: this won't work with aria attributes https://reactjs.org/docs/accessibility.html#wai-aria

sadafie: Not sure I follow for the aria attributes. Aria attributes are `aria-*` in JSX and remain the same in html.

sadafie:
> Wondering if there might be a reference lib for converting props to HTML attributes - the current implementation definitely has edge case bugs

I did some searching for this and couldn't find anything. Perhaps my search wasn't comprehensive enough.

sadafie:
Actually there's this - https://www.npmjs.com/package/react-attr-converter

PR #561, opened 2018-10-02
UXCapture New Relic instrumentation

sadafie:
This PR creates two new components:

**UXCaptureConfig:**
```
const onMark = (mark) => { // do something};
const onMeasure = (measure) => { //do something};
<UXCaptureConfig onMark={onMark} onMeasure={onMeasure} />
```

This generate the `ux-capture` library script in the form of:
```js
<script>
if (window.UX) {
    window.UX.config({
         onMark: (mark) => { // do something},
         onMeasure: (measure => { //do something},
}
</script>
```

**UXCapture:**
This is a wrapper component of `UXCaptureExpect` and `UXCaptureConfig`:
```
<UXCapture
    destinationVerified={[...]}
    primaryContentDisplayed={[...]}
    primaryActionAvailable={[...]}
    secondaryContentDisplayed={[...]}
    onMark={[...]}
    onMeasure={[...]} />
```
This will produce the following `ux-capture` library javascript:
```js
<script>
if (window.UX) {
    window.UX.expect({...});
}
</script>
<script>
if (window.UX) {
    window.UX.config({
         onMark: (mark) => { // do something},
         onMeasure: (measure => { //do something},
}
</script>
```


sadafie:
I think I'm going to leave it as is for now, but one consideration to keep in mind is that with this construct, each consumer will be defining it's own `onMark` and `onMeasure` callbacks. The alternative is to have the callbacks defined in the platform code here in the `UXCapture` component itself. I'm hesitant to do that quite yet because I'm uncertain how this library will be used moving forward; we may not want to tie this components to closely to meetup specific logic.

packages/mwp-app-render/src/components/uxcapture/UXCaptureConfig.jsx
```diff
@@ -0,0 +1,37 @@
...
+	// No need to inject `UX.config()` script if no callbacks provided
+	if (!onMark && !onMeasure) {
+		return null;
+	}
+
+	if (onMark) {
+		config['onMark'] = onMark.toString();
+	}
+
+	if (onMeasure) {
```
sadafie: Any thoughts on maybe using the Uglify JS API here - https://github.com/mishoo/UglifyJS2#api-reference - to minify, strip comments - here? Might make the rendered html output a bit more concise.

packages/mwp-app-render/src/components/uxcapture/UXCaptureConfig.jsx
```diff
@@ -0,0 +1,37 @@
...
+export default ({ onMark, onMeasure }: Props) => {
+	const config = {};
+
+	// No need to inject `UX.config()` script if no callbacks provided
+	if (!onMark && !onMeasure) {
+		return null;
+	}
+
+	if (onMark) {
+		config['onMark'] = onMark.toString();
```
sadafie: @mmcgahan We need to use a combination of `toString()` + `JSON.stringify()` in order to `__dangerouslySetInnerHTML` below. However, the result is script that has the callback as a string instead of a function:

```
<script>
if (window.UX){
    window.UX.config({
        "onMark": "function(mark){ ....}",
    });
}
</script>
```

I'm trying to think of solutions using `eval` (something else?) but coming up short. Any thoughts?

mmcgahan: I think this issue exposes a potential issue with `onMark` and `onMeasure` in general - since they are part of server-side markup and executed outside of the React application, they only data they can access is the `window` context (including `window.UX`) - they can't, for example, call `setState` or access props on their parent object when the callback executes, which is a _very_ subtle constraint that is going to be difficult to communicate to consumers. However, it also constrains the utility of `onMark` and `onMeasure` to the point that I suspect we don't need to expose them to consumers at all, in which case we probably don't need `UXCaptureConfig.jsx` at all.

To demo what I mean, there's a critical flaw in this code, but it's impossible to see without understanding the implementation details of `UXCaptureConfig`:

```jsx
const Foo = ({ doSomething }) => (
  <UXCaptureConfig onMark={() => doSomething()} />
);
```

This would be completely reasonable code in regular React, but the markup it would generate is something like:

```html
<script>window.UX.config({ onMark: () => doSomething() });</script>
```

when `onMark` was fired, `() => doSomething()` would throw an error because `doSomething` is undefined in the 'raw' HTML/`window` context.

Basically, the _only_ behavior that is allowed in `onMark` and `onMeasure` is behavior that directly interacts with vanilla `window`-based JS APIs, which is something that most frameworks (but particularly React) are designed to avoid.

SO: before getting too deep into implementation tradeoffs, I want to check that exposing 'global' `onMark` and `onMeasure` handlers to the consumer interface through `UX.config` is actually something we want to do. If I remember right, I _think_ that the main purpose of those callbacks is to send the data to New Relic, which I think can still work as long as we can access the `window.newrelic` object, but I think we might want to sort of hard-code that call into our toolset rather than leave it completely open to consumers.

packages/mwp-app-render/src/components/uxcapture/UXCaptureConfig.jsx
mmcgahan: Given the size of the HTML snippet (< 0.5kB), I don't think I would make the effort.

sergeychernyshev:
I agree, it's better to keep specifics of callback implementation outside of these components.

sadafie:
I've created a new `UXCaptureNewRelicConfig.jsx` component that specifies the `onMark` and `onMeasure` callbacks and passes it along to `UXCaptureConfig.jsx`. This way the newrelic logic can be encapsulated apart from the consumer code. @mmcgahan lmk what you think.

sadafie:
Update: I moved the New Relic callback logic into the `UXCapture` component. All other components such as `UXCaptureExpect`, `UXCaptureConfig`, `UXCaptureImageOnload`, and `UXCaptureInlineScript` are generic React bindings that can be exported into their own library at a later time.

The `UXCapture` component is the only Meetup specific component and essentially just combines `UXCaptureExpect` and `UXCaptureConfig` and passes along NewRelic `onMark` and `onMeasure` callbacks to `UXCaptureConfig`. This was the New Relic logic is encapsulated in a way that won't add a layer of complexity to the consumer apps.

packages/mwp-app-render/src/components/uxcapture/UXCapture.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+type Props = {|
+	destinationVerified?: Array<string>,
+	primaryContentDisplayed?: Array<string>,
+	primaryActionAvailable?: Array<string>,
+	secondaryContentDisplayed?: Array<string>,
+|};
+
+export const onMark = (mark: string) => {
+	if (window.newrelic) {
```
mmcgahan: nitpick, you can reduce nesting with a guard pattern

```js
if (!window.newrelic) {
  return;
}

// 'happy path' below can be left-aligned
```

packages/mwp-app-render/src/components/uxcapture/UXCapture.jsx
```diff
@@ -0,0 +1,64 @@
...
+		if (window.performance) {
+			// `window performance.timing.navigationStart` is the event that NR uses as 'start' of page load
+			const timeToMark = now - window.performance.timing.navigationStart;
+			window.newrelic.setCustomAttribute(mark, timeToMark);
+		}
+	}
+};
+
+export const onMeasure = (measure: string) => {
+	if (window.newrelic) {
```
mmcgahan: same - guard pattern probably useful here

packages/mwp-app-render/src/components/uxcapture/UXCaptureConfig.jsx
```diff
@@ -0,0 +1,30 @@
+// @flow
+import React from 'react';
+
+type callbackType = (label: string) => mixed;
```
mmcgahan: probably `void` return type? The return value can't be consumed by anything, so better to let consumers know that they shouldn't return anything.

packages/mwp-app-render/src/components/uxcapture/uxCaptureConfig.test.jsx
```diff
@@ -0,0 +1,51 @@
...
+import { shallow } from 'enzyme';
+import UXCaptureConfig from './UXCaptureConfig';
+
+const MOCK_PROPS = {
+	onMark: foo => {
+		if (foo) {
+			return foo;
+		}
+	},
+	onMeasure: (bar, noop) => {
```
mmcgahan: I think `onMeasure` should only get one argument?

sadafie: yeppp, not sure what happened here. :)

packages/mwp-app-render/src/components/uxcapture/UXCapture.jsx
```diff
@@ -0,0 +1,64 @@
...
+type Props = {|
+	destinationVerified?: Array<string>,
+	primaryContentDisplayed?: Array<string>,
+	primaryActionAvailable?: Array<string>,
+	secondaryContentDisplayed?: Array<string>,
+|};
+
+export const onMark = (mark: string) => {
+	if (window.newrelic) {
+		const now = new Date().getTime();
```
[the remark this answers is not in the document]
sadafie: thanks! I wasn't aware of this!

packages/mwp-app-render/src/components/uxcapture/UXCapture.jsx
```diff
@@ -11,44 +11,56 @@ type Props = {|
...
-	const now = new Date().getTime();
+	const markStartTime = window.performance
+		.getEntriesByType('mark')
+		.find(entry => entry.name === mark).startTime;
 
 	// Set a marker in the trace details
 	window.newrelic.addToTrace({
 		name: mark,
-		start: now,
+		start: markStartTime,
```
sergeychernyshev: Sadaf, are you sure this is the right time for NewRelic?

It is relative to navigationStart and not a full timestamp.

packages/mwp-app-render/src/components/uxcapture/UXCapture.jsx
```diff
@@ -11,44 +11,56 @@ type Props = {|
...
+	// TBD: not sure if we'll actually need this
+	window.newrelic.addToTrace({
+		name: `${measure}-endTime`,
+		start: performanceMeasure.startTime + performanceMeasure.duration,
+		type: 'UX Capture measure',
+	});
+
+	//  Add a custom attribute to the PageView & BrowserInteraction events in Insights
+	// `window performance.timing.navigationStart` is the event that NR uses as 'start' of page load
+	const timeToMeasure =
```
sergeychernyshev: This should just be `performanceMeasure.duration`, I believe.

sadafie: Is duration what you want tracked? How is this data going to be used in New Relic?
I was under the impression that we just wanted the startTimes of measure tracked.

packages/mwp-app-render/src/components/uxcapture/UXCapture.jsx
sadafie: Not sure what you mean by `It is relative to navigationStart and not a full timestamp.`

packages/mwp-app-render/src/components/uxcapture/UXCapture.jsx
sergeychernyshev: Sorry, I don't get it - `performanceMeasure.startTime` is always 0.

`performanceMeasure.duration` is the one that keeps the difference between `window.performance.timing.navigationStart` and `lastPerformanceMark.startTime` (e.g. those horizontal orange bars on UserTiming timeline in Chrome DevTools)

packages/mwp-app-render/src/components/uxcapture/UXCapture.jsx
sergeychernyshev: UserTiming mark's `startTime` only keeps a number of seconds (e.g. `103.79999998258427`) since navigationStart, which is a full timestamp, equal to number of seconds since epoch (e.g. `1539110917528`)

PR #565, opened 2018-10-12
explicitly handle 4 types of activity-generating requests

mmcgahan:
The `url` and `referrer` fields are not being consistently populated as expected in consumer apps - after discussions on Slack with the data team, this PR explicitly defines the 'referrer override' case for SPA navigation, and sets up unit tests to validate the records for other types of requests that generate activity tracking records.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [17.4]
+
+- **Refactor** `mwp-tracking-plugin` now explicitly handles each type of request
+  that results in an activity record to ensure that the expected `url` and
+  `referrer` values are populated
+
 ## [17.3]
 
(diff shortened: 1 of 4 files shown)


mmcgahan:
@sadafie unfortunately no clear indication of what changed, although from the code the _previous_ behavior was the buggy one. There were a number of PRs in mup-web related to activity tracking that merged right around the time that the behavior changed, and there was a click tracking update to MWP that merged at that time, which may have 'fixed' a tracking bug without realizing that it changed the activity tracking records, so it has been hard to tease out the specific code change that was responsible.

The unit tests and specific code comments should help us all understand and maintain the intended behavior going forward - the desired formatting has been ambiguous even to me for a long time.

PR #568, opened 2018-10-16
Use High Resolution timer for TTI

sadafie:


packages/mwp-app-render/src/components/PageWrap/index.jsx
```diff
@@ -28,20 +28,20 @@ class PageWrap extends React.Component {
...
 			window.newrelic.addToTrace({
 				name: 'appInteractive',
-				start: now,
+				start: window.performance.timing.navigationStart + now,
 				type: 'Browser app has rendered and is interactive',
 			});
-			// 2. Add a custom attribute to the PageView & BrowserInteraction events in Insights
+
+			// Add a custom attribute to the PageView & BrowserInteraction events in Insights
 			window.performance &&
```
sergeychernyshev: No need to check for window.performance again here 
```suggestion
```

sergeychernyshev: Trying this suggestion thing here, let me know if it works for you - it shows some weird markup for me, but the diff seems to be what I intended it to ;)

sadafie: I think it works well for small changes like this! Obviously we aren't running our commit hooks so things like prettier and eslint don't run on commit, so that could be an "issue". Otherwise, easy to do.

sergeychernyshev: Yeah, pre-commit hooks are kind-of important here...

sergeychernyshev: But it's probably the same as editing a file through GitHub UI, so nothing new.

packages/mwp-app-render/src/components/PageWrap/index.jsx
```diff
@@ -28,20 +28,19 @@ class PageWrap extends React.Component {
...
 			});
-			// 2. Add a custom attribute to the PageView & BrowserInteraction events in Insights
-			window.performance &&
-				window.newrelic.setCustomAttribute(
-					'timeToAppInteractive',
-					now - window.performance.timing.navigationStart // this is the event that NR uses as 'start' of page load
-				);
+
+			// Add a custom attribute to the PageView & BrowserInteraction events in Insights
+				window.newrelic.setCustomAttribute('timeToAppInteractive', now);
```
sergeychernyshev:
```suggestion
			window.newrelic.setCustomAttribute('timeToAppInteractive', now);
```

packages/mwp-app-render/src/components/PageWrap/index.jsx
```diff
@@ -40,7 +40,7 @@ class PageWrap extends React.Component {
 			});
 
 			// Add a custom attribute to the PageView & BrowserInteraction events in Insights
-				window.newrelic.setCustomAttribute('timeToAppInteractive', now);
+span class="pl-c1"> 			window.newrelic.setCustomAttribute('timeToAppInteractive', now);
```
sergeychernyshev: Arrgh, looks like they have a bug there and this code leaks into the diff

sadafie: yeah, it's still in there. i'll do the commit manually through my editor

PR #571, opened 2018-10-30
fallback to language code in referer url

jaclynj:


packages/mwp-language-plugin/src/util/getLanguage.js
```diff
@@ -29,7 +29,16 @@ export const getCookieLang: ParseRequestLang = (request: HapiRequest) => {
 export const getUrlLang: ParseRequestLang = (request: HapiRequest) => {
 	const { supportedLangs } = getServerSettings(request);
 	const urlLang = request.url.path.split('/')[1];
-	return supportedLangs.includes(urlLang) && urlLang;
+	const validRequestLang = supportedLangs.includes(urlLang) && urlLang;
+
+	// If request url language is invalid, look at referer for language code
+	// This is for cases when the request url is the api proxy path ie `/mu_api/`
+	if (!validRequestLang && request.headers['referer']) {
```
jgl2832: im not familiar with how the referer header gets set, but is it possible for this to be an external URL? if it is, might be worthwhile to only do this if the initial `request.url` contains /mu_api/ in the pattern we expect

jaclynj: Yeah I was thinking that - but the proxy endpoint might not necessarily be /mu_api/ because that's a variable you can set in the app config. I'll look into using that variable to check that here

packages/mwp-language-plugin/src/util/getLanguage.js
```diff
@@ -29,7 +29,16 @@ export const getCookieLang: ParseRequestLang = (request: HapiRequest) => {
...
+		const refererUrl = new URL(request.headers['referer']);
+		const refererLang = refererUrl.pathname.split('/')[1];
+		return supportedLangs.includes(refererLang) && refererLang;
+	}
+	return validRequestLang;
```
sadafie: General question, does the platform code then set a language cookie after it parses the language from a url? I'm going to assume yes, but if not, we should.

jgl2832: it does not- currently all the cookie setting happens from chapstick. i'd actually love to have a discussion soon generally about the language cookie/param/pref landscape and try to come up with a better overall approach

packages/mwp-language-plugin/src/util/getLanguage.js
jaclynj: Also FYI this function is also called when we do the initial server side render so it is possible in that case that the referrer will be an external url

mmcgahan: I think the conventional way to do this would be to allow routes to specify plugin-specific configuration - essentially we want this plugin to behave differently on the api proxy route than on the app render route.

We do that for a couple of other plugins in the [mwp-api-proxy-plugin route config](https://github.com/meetup/meetup-web-platform/blob/WP-871_native-app-platformagent/packages/mwp-api-proxy-plugin/src/routes.js#L29) so you can probably follow that pattern. It's not very intuitive from an implementation standpoint because you have to know how to access the route config values from the request (e.g. [like this](https://github.com/meetup/meetup-web-platform/blob/WP-871_native-app-platformagent/packages/mwp-logger-plugin/src/index.js#L20-L22)), but it's not too terrible.

This is definitely stuff that could use an architectural revamp with @jgl2832 at some point

mmcgahan: I think the more conventional place to check the referrer header is `request.info.referrer` https://hapijs.com/api#-requestinfo

packages/mwp-language-plugin/src/util/getLanguage.js
mmcgahan: I think that was the underlying cause of this bug - when the code was first written, I don't think there were cases where you could get to mup-web without having a MEETUP_LANGUAGE cookie, but now there are. Definitely worth a discussion to define what the expected behavior should be in different cases

jaclynj:
@mmcgahan Thanks for the feedback - this is ready for re-review if you could take a look.

packages/mwp-language-plugin/src/util/getLanguage.js
```diff
@@ -27,8 +27,17 @@ export const getCookieLang: ParseRequestLang = (request: HapiRequest) => {
  * get the URL-specified language - check for existing language path prefixes
  */
 export const getUrlLang: ParseRequestLang = (request: HapiRequest) => {
+	const { url, info: { referrer }, route } = request;
 	const { supportedLangs } = getServerSettings(request);
-	const urlLang = request.url.path.split('/')[1];
+	const { plugins = {} } = route.settings || {};
```
mmcgahan: I'm pretty sure `route.settings` will always be an object, and `route.settings.plugins` gets initialized to `{}` - no need for fallbacks here

I think `route.settings` is now just an alias to the (preferred) `route.options` object https://hapijs.com/api#-routeoptionsplugins

PR #572, opened 2018-11-05
UX Capture font instrumentation using font loader library

sergeychernyshev:
Solution for font loading instrumentation:
https://meetup.atlassian.net/browse/WP-851


packages/mwp-app-render/src/components/Dom.jsx
```diff
@@ -87,6 +89,7 @@ const DOM = props => {
 				{newrelicJS &&
 					<script dangerouslySetInnerHTML={getInnerHTML(newrelicJS)} />}
 				<script dangerouslySetInnerHTML={getInnerHTML(uxCaptureJS)} />
+				<script dangerouslySetInnerHTML={getInnerHTML(fontLoaderJS)} />
```
mmcgahan: looks like this will render

```html
<script>
  <script src="..." />
</script>
```

I think `fontLoaderJS` should probably be `fontLoaderSrc`, and then this line would be

```html
<script src={fontLoaderSrc} />
```

packages/mwp-app-render/src/components/uxcapture/UXCaptureFont.jsx
```diff
@@ -0,0 +1,28 @@
...
+                families: ['${fontFamily}']
+            },
+            active: function() {
+                window.performance.mark('${mark}');
+            },
+        });
+    </script>
+`;
+const UXCaptureFont = ({ fontFamily, mark }: Props) =>
+	<script
```
mmcgahan: similar here - `generateUXCaptureFontJS` should probably not specify `<script>` wrapping tag

packages/mwp-app-render/src/components/Dom.jsx
sergeychernyshev: For this one, I hope to actually replace it with inline code instead, once it all works out because timing for network request can skew this way too much.

I might need to look into how to minimize the size of the library first (although it's only 6K on the wire).

sergeychernyshev:
OK, we are now using a custom build of Font Loader with "custom" module only which applications can inline.

packages/mwp-app-render/src/components/uxcapture/UXCaptureFont.jsx
```diff
@@ -0,0 +1,44 @@
...
+    });
+`;
+
+// fontFamily attribute should include individual weights if separate files are used
+// Example: "Graphik Meetup:n4,n5,n6"
+// Attention, if no weight specified, only n4 (e.g. font-weight: 400) is tested for!
+// See https://github.com/typekit/webfontloader#events for more detail
+const UXCaptureFont = ({ fontFamily, mark }: Props) =>
+	<script
+		dangerouslySetInnerHTML={{
```
jaclynj: Are you sure this works? I don't think it will. This is why we were having problems when pairing. See here: https://github.com/facebook/react/issues/8838#issuecomment-314740580
```
script tags injected via innerHTML are not executed for security reasons:

https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML

Although this may look like a cross-site scripting attack, the result is harmless. HTML5 specifies that a <script> tag inserted via innerHTML should not execute.
```

PR #579, opened 2018-12-08
hapi17 more updates to signal sci

jaclynj:


packages/mwp-rasp-plugin/src/sigsci/index.js
```diff
@@ -536,8 +536,7 @@ Sigsci.prototype.onPostResponse = function(err /* , rpcResponse */) {
 Sigsci.prototype.hapi = function() {
 	var self = this;
 	return function(request, h) {
-		self.middlewareHapi(request, h);
-		return h.continue;
+		return self.middlewareHapi(request, h);
```
mmcgahan: with this approach (which I think is good), you can simplify to:

```js
Sigsci.prototype.hapi = function() {
  return this.middlewareHapi.bind(this)
}
```

packages/mwp-rasp-plugin/src/sigsci/index.js
```diff
@@ -533,24 +536,19 @@ Sigsci.prototype.onPostResponse = function(err /* , rpcResponse */) {
 // HAPI.JS Support
 // ------------------------------------
 
-Sigsci.prototype.hapi = function() {
```
mmcgahan: removed this function because it boils down directly into `Sigsci.middlewareHapi`

PR #580, opened 2018-12-10
MG-3302: make browserId and clientIP available 

samantids:


packages/mwp-core/src/util/cookieUtils.js
```diff
@@ -18,6 +22,15 @@ export const parseMemberCookie = state => {
 	return member;
 };
 
+export const parseBrowserIdCookie = state => {
+	if (!state[BROWSER_ID_COOKIE]) {
+		logger.warn('No browser id cookie found');
```
mmcgahan: probably no need for warning

packages/mwp-core/src/util/cookieUtils.test.js
```diff
@@ -38,6 +44,21 @@ describe('parseMemberCookie', () => {
 	});
 });
 
+describe('parseBrowserIdCookie', () => {
+	it('returns a browser id', () => {
+		const requestState = { [BROWSER_ID_COOKIE]: 'id=abc-def-g1234' };
```
mmcgahan: `MEETUP_BROWSER_ID` actually includes double quotes around it - this test should be

```suggestion
		const requestState = { [BROWSER_ID_COOKIE]: '"id=abc-def-g1234"' };
```

(which I think will fail, but not sure)

PR #581, opened 2018-12-21
UX-Capture: WP 856 SPA support for React bindings

mmcgahan:
Still need to get a better `componentDidUpdate` check on `UXCaptureStartView`, but otherwise the basic implementation is close

## TODO
- [ ] tests
- [ ] review/add docs

### DIFF
--- a/packages/mwp-app-render/src/components/uxcapture/UXCaptureStartView.jsx
+++ b/packages/mwp-app-render/src/components/uxcapture/UXCaptureStartView.jsx
@@ -8,56 +8,64 @@ type Props = {|
 	secondaryContentDisplayed?: Array<string>,
 |};
 
-export default ({
-	destinationVerified,
-	primaryContentDisplayed,
-	primaryActionAvailable,
-	secondaryContentDisplayed,
-}: Props) => {
-	const zones = [];
-
-	if (destinationVerified && destinationVerified.length > 0) {
-		zones.push({
+const makeZones = (props: Props) =>
+	[
+		{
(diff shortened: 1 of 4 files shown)


packages/mwp-app-render/src/components/uxcapture/UXCaptureStartView.jsx
```diff
@@ -8,56 +8,70 @@ type Props = {|
...
 
-	// Don't add UX.expect() to page is no zones specified
-	if (!zones.length) {
-		return null;
+export default class UXCaptureStartView extends React.Component<Props> {
+	zones = makeZones(this.props);
+	componentDidUpdate(prev: Props) {
+		// updated on client - if mark name has changed, clear old mark and trigger new
+		if (
+			prev.destinationVerified !== this.props.destinationVerified &&
```
sergeychernyshev: Should this trigger if only one of the props has changed and not all of them?

mmcgahan: I _think_ so, but it sort of depends on what kind of rules we think we can enforce about mark naming - I can definitely imagine different related views (e.g. tabs) re-using the zone definitions but changing only 1 or 2 expected mark names, so I think we need to look for pretty small prop changes

mmcgahan:
Update: SPA support will arrive through meetup/ux-capture in a monorepo - React bindings will be moved out of meetup/meetup-web-platform. This PR should stay open until that happens _but should not be merged_

PR #583, opened 2018-12-24
TAX-358 Update mwp-core to add a request geo location to LaunchDarkly user data

mkleymenov:
*Description*
Tax Collection project is going to be launched initally in just two jurisdictions: Australia and Utah, US. The feature availability is managed by several LaunchDarkly flags which currently have access only to Meetup member information. In order to enable separation by geo location required for beta launch, we need to add a country code and a region code resolved from a client request to a payload sent to LaunchDarkly. 

*What does this PR do?*
1. Introduced Flow type for LaunchDarkly user (request payload) based on the official documentation for Node SDK: https://docs.launchdarkly.com/docs/node-sdk-reference#section-users
2. Updated `mwp-app-router-plugin` to send an instance of LaunchDarklyUser instead of member state object to LaunchDarkly when fetching feature flags for a member.
3. Updated `mwp-core` to create a LaunchDarklyUser from a member state object and a request object.
4. Added unit tests for the mapping method

*How did I test this PR?*
With unit tests and on local pro-web setup. 
In order to test the changes in the components locally, I've put the modified components to `node_modules` of `pro-web` and started the app. I was simulating possible values of `X-Region` request header, normally set by Fastly, with Modify Headers Chrome plugin. By playing with various combinations of `X-Region` header value and a `pro-tax-checkout-form` feature flag configuration I was able to confirm that targeting works as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## [18.2]
+
+-   **Refactor** `mwp-app-route-plugin` defines a `LaunchDarklyUser` type for a LaunchDarkly user object
+    based on the official [documentation](https://docs.launchdarkly.com/docs/node-sdk-reference#section-users)
+-   **Refactor** `mwp-core` uses a `LaunchDarklyUser` object to call `getFlags` method of `mwp-app-route-plugin`.
+-   **New Feature** `mwp-core` adds a request country and region provided by Fastly to LaunchDarkly user
+    custom attributes.
+
 ## [18.1]
 
 -   Adds new rasp plugin which monitors and prevents traffic from bad actors.
--- a/flow-typed/platform.js
+++ b/flow-typed/platform.js
@@ -143,3 +143,23 @@ declare type CookieOpts = {
 	isSecure?: boolean,
 	encoding?: string,
 };
+
+// See https://docs.launchdarkly.com/docs/node-sdk-reference#section-users
+declare type LaunchDarklyUser$CustomAttributes = {
+	RequestCountry?: string,
+	RequestRegion?: string,
+	[string]: string | boolean | number | Array<string> | Array<boolean> | Array<number>
+};
+
+declare type LaunchDarklyUser = {|
+	key: string,
+	ip?: string,
+	firstName?: string,
+	lastName?: string,
+	country?: string,
+	email?: string,
+	avatar?: string,
+	name?: string,
+	anonymous?: boolean,
+	custom?: LaunchDarklyUser$CustomAttributes
+|};
--- a/packages/mwp-app-route-plugin/src/index.js
+++ b/packages/mwp-app-route-plugin/src/index.js
@@ -20,14 +20,13 @@ export function register(
 	const ldClient = LaunchDarkly.init(options.ldkey || LAUNCH_DARKLY_SDK_KEY, {
 		offline: process.env.NODE_ENV === 'test',
 	});
-	server.expose('getFlags', memberObj => {
-		const key = (memberObj && memberObj.id) || 0;
-		return ldClient.allFlags({ ...memberObj, key, anonymous: key === 0 }).then(
+	server.expose('getFlags', (user: LaunchDarklyUser) => {
+		return ldClient.allFlags(user).then(
 			flags => flags,
 			err => {
 				server.app.logger.error({
 					err,
-					member: memberObj,
+					launchDarklyUser: user,
 				});
 				return {}; // return empty flags on error
 			}
--- a/packages/mwp-core/src/renderers/server-render.jsx
+++ b/packages/mwp-core/src/renderers/server-render.jsx
@@ -15,6 +15,7 @@ import ServerApp from 'mwp-app-render/lib/components/ServerApp';
 import { parseMemberCookie } from 'mwp-core/lib/util/cookieUtils';
 
 import { getVariants, parseBrowserIdCookie } from '../util/cookieUtils';
+import { getLaunchDarklyUser } from '../util/launchDarkly';
 
 const DOCTYPE = '<!DOCTYPE html>';
 const DUMMY_DOMAIN = 'http://mwp-dummy-domain.com';
@@ -282,15 +283,18 @@ const makeRenderer = (
 		// feature flags can be selected based on member id,
 		// email, and other properties.
 		// feature flags based on member id are available before the store is populated.
-		const addFlags = (populatedStore, member) =>
-			// getFlags needs as much member info as possible, but particularly id and email
-			// in order to match on common targeting rules
-			request.server.plugins['mwp-app-route'].getFlags(member).then(flags =>
-				populatedStore.dispatch({
-					type: 'UPDATE_FLAGS',
-					payload: flags,
-				})
-			);
+		const addFlags = (populatedStore, member) => {
+			// Populate a LaunchDarklyUser object from member and request details
+			const launchDarklyUser = getLaunchDarklyUser(member, request);
+			return request.server.plugins['mwp-app-route']
+				.getFlags(launchDarklyUser)
+				.then(flags =>
+					populatedStore.dispatch({
+						type: 'UPDATE_FLAGS',
+						payload: flags,
+					})
+				);
+		};
 
 		const checkReady = state =>
 			state.preRenderChecklist.every(isReady => isReady);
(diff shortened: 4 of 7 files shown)


flow-typed/platform.js
```diff
@@ -143,3 +143,23 @@ declare type CookieOpts = {
 	isSecure?: boolean,
 	encoding?: string,
 };
+
+// See https://docs.launchdarkly.com/docs/node-sdk-reference#section-users
+declare type LaunchDarklyUser$CustomAttributes = {
+	RequestCountry?: string,
+	RequestRegion?: string,
```
mkleymenov: LaunchDarkly allows any custom attribute names, but I decided to reserve these two as fields in order to enforce naming convention (the same property names are used by Chapstick when passing resolved geo location to LaunchDarkly)

packages/mwp-app-route-plugin/src/index.js
```diff
@@ -20,14 +20,13 @@ export function register(
 	const ldClient = LaunchDarkly.init(options.ldkey || LAUNCH_DARKLY_SDK_KEY, {
 		offline: process.env.NODE_ENV === 'test',
 	});
-	server.expose('getFlags', memberObj => {
-		const key = (memberObj && memberObj.id) || 0;
-		return ldClient.allFlags({ ...memberObj, key, anonymous: key === 0 }).then(
+	server.expose('getFlags', (user: LaunchDarklyUser) => {
```
mkleymenov: I did a full text search for `getFlags` in the entire `meetup-web-components` and it looks like the only two usages are in [server-render.jsx](https://github.com/meetup/meetup-web-platform/pull/583/files#diff-d7329f1db146f21557dc646dcd0d2350). So hopefully there shouldn't be any unexpected breakages because of signature change.

packages/mwp-core/src/util/launchDarkly.js
```diff
@@ -0,0 +1,76 @@
...
+
+	const location = (headers && headers[HEADER_FASTLY_X_REGION]) || '';
+	const [country, region] = location
+		.split('/')
+		.map(value => (value ? value.toUpperCase() : value));
+
+	const custom: LaunchDarklyUser$CustomAttributes = {};
+
+	if (country) {
+		custom.RequestCountry = country;
```
andrew-bykovets:
```
const requestCountry = country && {RequestCountry: country}
 const custom: LaunchDarklyUser$CustomAttributes = {
 ...requestCountry,
...requestRegion,
};
```

mkleymenov:
```
string [1] is not an object.

     packages/mwp-core/src/util/launchDarkly.js
      66│ 	const requestCountry = country && { RequestCountry: country };
      67│ 	const requestRegion = region && { RequestRegion: region };
      68│
      69│ 	return {
      70│ 		...requestCountry,
      71│ 		...requestRegion,
      72│ 	};
      73│ };
      74│

     /tmp/flow/flowlib_3df97785/core.js
 [1] 323│     split(separator?: string | RegExp, limit?: number): Array<string>;
```

mkleymenov: However this works:
```
const requestCountry = country ? { RequestCountry: country } : {};
const requestRegion = region ? { RequestRegion: region } : {};

return {
	...requestCountry,
	...requestRegion,
};
```

The object expansion in the return statement works because `LaunchDarklyUser$CustomAttributes` is an inexact type, so flow doesn't care that the expansion can potentially add properties that are not explicitly defined in the type

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## [18.2]
```
mmcgahan: you'll need to bump the minor version in the Makefile to publish 18.2.x

packages/mwp-app-route-plugin/src/index.js
```diff
@@ -20,14 +20,13 @@ export function register(
...
+		return ldClient.allFlags(user).then(
 			flags => flags,
 			err => {
 				server.app.logger.error({
 					err,
-					member: memberObj,
```
mmcgahan: If I remember right, `member` is a magic keyword for this log that will allow stackdriver to annotate the error with the particular member ID that generated the error - you can just make this `member: user`

packages/mwp-core/src/util/launchDarkly.js
```diff
@@ -0,0 +1,73 @@
...
+		user.email = email;
+	}
+	if (country) {
+		user.country = country;
+	}
+
+	return user;
+};
+
+const getRemoteIp = (request: HapiRequest): ?string => {
```
mmcgahan: Since this isn't an LD-specific utility, could you put it in a new module, e.g. `util/requestinfo.js` or something similar? Definitely a useful function to have available

packages/mwp-core/src/util/launchDarkly.js
```diff
@@ -0,0 +1,73 @@
...
+	const fromQuery = query && query.__set_geoip && query.__set_geoip.toString();
+	const fromHeaders = headers && headers[HEADER_FASTLY_CLIENT_IP];
+	const fromInfo =
+		info && typeof info.remoteAddress === 'string'
+			? info.remoteAddress
+			: undefined;
+	return fromQuery || fromHeaders || fromInfo;
+};
+
+const getCustomAttributes = (
```
mmcgahan: code comment for what 'custom attributes' means in this context would be useful - maybe a link to LD docs?

packages/mwp-core/src/util/launchDarkly.js
```diff
@@ -0,0 +1,73 @@
...
+		.split('/')
+		.map(value => (value ? value.toUpperCase() : value));
+
+	const requestCountry = country ? { RequestCountry: country } : {};
+	const requestRegion = region ? { RequestRegion: region } : {};
+
+	return {
+		...requestCountry,
+		...requestRegion,
+	};
```
mmcgahan: this kind of composition seems more difficult to read than just optionally adding a property to an object:

```js
const customAttr = {};
if (country) {
  customAttr.RequestCountry = country;
}
if (region) {
  customAttr.RequestRegion = region;
}
return customAttr
```

It's slightly more verbose than the object spread composition, but it's _much_ more clear what is happening.

packages/mwp-app-route-plugin/src/index.js
mkleymenov: Done

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -31,15 +32,14 @@ function getHtml(el) {
 	return `${DOCTYPE}${htmlMarkup}`;
 }
 
-export function getRedirect(context) {
+export function getRedirect(context: { url?: string, permanent?: boolean }) {
```
mkleymenov: Fixes a Flow error about an untyped `context` argument

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -31,15 +32,14 @@ function getHtml(el) {
...
 		return;
 	}
 	// use `URL` to ensure valid character encoding (e.g. escaped emoji)
-	const isFragment = context.url.startsWith('/');
-	const urlToFormat = isFragment
-		? `${DUMMY_DOMAIN}${context.url}`
-		: context.url;
+	const url: string = context.url;
+	const isFragment = url.startsWith('/');
+	const urlToFormat = isFragment ? `${DUMMY_DOMAIN}${url}` : url;
```
mkleymenov: Fixes a Flow error that `context.url` cannot be cast to string because it can be undefined. I've also added a few [test cases](https://github.com/meetup/meetup-web-platform/pull/583/files#diff-d9666b70703fe4e60a7d86d081316807) to validate the behavior when `context.url` is not defined or empty

packages/mwp-core/src/util/launchDarkly.js
mkleymenov: Done

packages/mwp-core/src/util/launchDarkly.js
mkleymenov: @mmcgahan Good idea. I've moved parsing of remote IP and parsing geo location from `X-Region` header to a new module `requestUtils`

packages/mwp-core/src/util/launchDarkly.js
mkleymenov: @mmcgahan Got it, reverted back to using ifs

PR #584, opened 2019-01-02
WP 956: State Error in activity tracker plugin

sadafie:


packages/mwp-tracking-plugin/src/activity.js
```diff
@@ -149,12 +149,15 @@ const getOnPreResponse = cookieConfig => (request, h) => {
...
 	};
 
-	if (browserId) {
-		request.response.state(browserIdCookieName, browserId, FOREVER);
-	}
-	if (trackId) {
-		request.response.state(trackIdCookieName, trackId, FOREVER);
+	if (!request.response.isBoom) {
+		if (browserId) {
+			h.state(browserIdCookieName, browserId, FOREVER);
```
sadafie: `h.state` and `request.response.state` are near identical (request.response.state just returns `this`). Let's use the toolkit everywhere for consistency.

sadafie:
I'm working on trying to get a unit test written for this, otherwise, I don't know how to test in dev and we might have to verify in prod.

packages/mwp-app-server/package.json
```diff
@@ -29,7 +29,7 @@
   "dependencies": {
     "convict": "4.0.1",
     "electrode-csrf-jwt": "git+https://github.com/meetup/electrode-csrf-jwt.git#793431729d8fb4ee315f28709f323f198b25d460",
-    "hapi": "17.4.0",
+    "hapi": "17.8.1",
```
mmcgahan: tangent - any info on what this update includes?

packages/mwp-tracking-plugin/src/activity.js
mmcgahan: agreed. The redundancy between the two method implementations is surprising :/

https://github.com/hapijs/hapi/blob/master/lib/response.js#L428-L432
https://github.com/hapijs/hapi/blob/master/lib/toolkit.js#L165-L168

PR #585, opened 2019-01-03
Wp 964: Remove state.app support

sadafie:
https://meetup.atlassian.net/browse/WP-964


packages/mwp-api-state/src/sync/index.js
```diff
@@ -156,28 +119,15 @@ export const getFetchQueriesEpic = (findMatches, fetchQueriesFn) => {
 			.then(({ successes = [], errors = [] }) => {
 				// meta contains a Promise that must be resolved
 				meta.resolve([...successes, ...errors]);
-				const deprecatedSuccessPayload = getDeprecatedSuccessPayload(
-					successes,
-					errors
-				);
-				const deprecatedActions = [apiSuccess(deprecatedSuccessPayload)];
-				if (meta && meta.onSuccess) {
-					deprecatedActions.push(meta.onSuccess(deprecatedSuccessPayload));
```
mmcgahan: Looks like this handler is still used in mup-web [here](https://github.com/meetup/mup-web/blob/master/src/app/register/registerActionCreators.js#L76-L80) and [here](https://github.com/meetup/mup-web/blob/master/src/app/register/registerActionCreators.js#L96-L100) - from what I can tell, it's used essentially the way that `action.meta.request.then(...)` is used, but I'm not sure how to keep the interface the same for `facebookTie` and `registerTopicsPost` - the other places where `onSuccess/onError` are specified don't actually do anything so they can be cleaned up separately.

2 options:

1. keep this `deprecatedActions` handling intact for now
2. work with Super Groups (I think) to figure out how to refactor those action creators to work with the recommended [Promise interface for queries](https://github.com/meetup/meetup-web-platform/blob/master/packages/mwp-api-state/Queries.md#promise-interface-for-dispatched-queries)

sadafie:
@mmcgahan can I get another look at this ticket / sign off to merge?

PR #586, opened 2019-01-07
WP 805: Update activity and click tracking timestamps

sadafie:
New format: *2019-01-07T10:27:02.791-05:00*

Question - is there a way to test in dev aside from unit tests?


packages/mwp-tracking-plugin/src/util/clickReader.js
```diff
@@ -1,4 +1,7 @@
-import { parseIdCookie } from './idUtils'; // TODO: provide this info through new plugin
+import {
+	parseIdCookie,
+	getZonedDateTimeStringWithUTCOffset,
+} from './trackingUtils'; // TODO: provide this info through new plugin
```
jaclynj: Not sure what this TODO means but should we backlog a ticket for this? If so could you add a ticket?

sadafie: Yeah, that's a pre-existing todo that i'll need to ask @mmcgahan about.

mmcgahan: 👍 created ticket, feel free to link here https://meetup.atlassian.net/browse/WP-968

dumptruckusa:
as for testing in dev: I'm not sure, but I remember a tool that Gruber wrote that allowed you to point logs at your dev box to you could view them. I had an Android build that did this at one point, but not sure how for web.

packages/mwp-tracking-plugin/src/util/trackingUtils.js
```diff
@@ -52,3 +57,10 @@ export const parseIdCookie = (
 	}
 	return parsed.id;
 };
+
+// ZoneId.toString() adds the timezone in brackets, which we want to remove
+// e.g.: 2019-01-07T10:27:02.791-05:00[America/New_York]
+export const getZonedDateTimeStringWithUTCOffset = () =>
+	ZonedDateTime.now(ZoneId.of('America/New_York'))
+		.toString()
+		.replace(/\[.*\]$/, '');
```
mmcgahan: not a blocker, but a slightly more efficient regex would be

```js
/\[[^\]+]\]$/
```

because the parser doesn't have to backtrack, but it's a bit harder to read so whatevs

packages/mwp-tracking-plugin/src/util/trackingUtils.js
```diff
@@ -52,3 +57,10 @@ export const parseIdCookie = (
 	}
 	return parsed.id;
 };
+
+// ZoneId.toString() adds the timezone in brackets, which we want to remove
+// e.g.: 2019-01-07T10:27:02.791-05:00[America/New_York]
+export const getZonedDateTimeStringWithUTCOffset = () =>
```
mmcgahan: sort of an optimization, but this seems like a utility that should ideally be able to accept an arbitrary `ZonedDateTime` - what do you think about decomposing this into a function like

```js
const getISOString = zonedDateTime => zonedDateTime.toString().replace(/\[.*\]$/, '');
// 'Now' timezone defaults to NYC
const getISOStringNow = (zoneId = ZoneId.of('America/New_York')) =>
  getISOString(ZonedDateTime.now(zoneId))
```

I think those primitives might be easier to apply to more applications without making this primary usecase more awkward

sadafie: Yeah I think this is fair to separate them out. I'm not sure how often these will get used, but makes sense to do so.

sadafie: I did it slightly differently. Instead of accepting a zoneId, it takes the tz string. This way all of the js-joda ZoneId logic can be encapsulated within the utility methods. You won't need a potentially extra import in places you're using this method.

PR #587, opened 2019-01-08
WP-692: Resolve circular dependencies

sadafie:
Not sure if I've resolved all of the circular dependencies yet, but here's a start / improvement.

- `yarn install` shows no errors
- `yarn check` shows no errors

All dev dependencies should be defined at the root level `package.json` file. Removed any directed mwp-* dependencies that are not used by individual packages.


sadafie:
Sorry, should have indicated this is still prelim. I'm still thinking over the `peerDependencies` approach. :) 

My aim is to simplify dependencies. The thought process is that though possible for consumers to mix-and-match packages, there are actually a minimum set of `mwp-*` dependencies that consumers must require in order for their application to run. For those dependencies, it makes sense to keep them as `peerDependencies` so that we don't potentially end up in situations where we have multiple versions. (That being said, our current package.json structures tend to use the ">=0.0.1" version construct and so we never end up with multi versions, but if that ever changes, we could.)

I actually think the more correct solution is a hybrid of `peerDependencies` and `dependencies`. The `mwp-*-plugin` packages should probably remain as direct `dependencies` while the remainder could live as `peerDependencies`. The key is to find the minimum set and define them. 

There probably is an argument to be made of the complexity of maintaining a hybrid approach - specifically that over time it will diverge. 

My next step is to test this branch in `mup-web` and inspect any obvious differences in either yarn.lock or with `yarn why *` and make a call. 

As for the actual circular dependencies, I did my best to manually verify that there aren't any top-level ones. I'm not getting the warnings from yarn workspaces anymore either. I haven't found any tools that can locate circular package dependencies, so that's not an option (would be nice!). I found a thread https://github.com/lerna/lerna/issues/1024 that more or less indicates what we already know.  It's also my understanding that yarn does it's best at allowing circular dependencies of the nature `a -> b -> a`, so the Warning would just be related to potential unexpected behavior, which I don't really think we've ever encountered. I'm just trying to do some sensible clean-up here while I'm at it. :)

mmcgahan:
@sadafie that sounds good. I think for the most part we're okay with installing multiple versions of dependencies in server code (it happens a lot through transitive dependencies, but there's not a concern about about bundle size on the server), just not in browser code, so I'd recommend avoiding `peerDependencies` unless there's a clear risk to loading multiple versions in a browser bundle.

sadafie:
I'm also going to close this ticket because it's old and out of sync.

PR #590, opened 2019-01-10
MG-4020 update list handle

andeliv:
D.o.D.:
- Add `isReverse` param to `merge` object
- Add new list from the start or existing if `isReverse = true`

### DIFF
--- a/packages/mwp-api-state/src/reducer.test.js
+++ b/packages/mwp-api-state/src/reducer.test.js
@@ -75,8 +75,9 @@ describe('getListState', () => {
 		).toEqual({});
 	});
 	it('returns a new object with dynamicRef', () => {
+		const ref = resp.query.list.dynamicRef;
 		expect(getListState(state, resp)).toEqual({
-			[resp.query.list.dynamicRef]: {
+			[ref]: {
 				value: resp.response.value,
 				query: resp.query,
 			},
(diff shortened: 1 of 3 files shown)


packages/mwp-api-state/src/reducer.test.js
```diff
@@ -93,6 +106,17 @@ describe('getListState', () => {
 			},
 		});
 	});
+	it('merges new response with existing dynamicRef adding new items at the top of the list', () => {
+		const value = ['qux'];
+		expect(
+			getListState({ [resp.query.list.dynamicRef]: { value } }, respWithReverse)
+		).toEqual({
+			[resp.query.list.dynamicRef]: {
```
sadafie: It feels as if there is a way to simplify/consolidate the unit test code, especially in regard to all the various resp* objects. As an example, I find it odd that we're using `resp` versus `respWithReverse` for things like `dynamicRef` and `response.value`, but perhaps I'm overlooking a nuance. This new test seems to conform with the others & refactoring is probably outside the scope of this ticket, unless you want to & feel like taking a stab at improving the test quality here :)

andeliv: Totally agree. Looks like not a big change, so will update here.

PR #592, opened 2019-01-28
Launch Darkly flag support for conditional CSS

sadafie:
As part of the Swarm updates, `@meetup/swarm-styles` will be publishing a new `globals.css` file that will be imported by `mup-web` and passed to the server render utilities in mwp-core. 

Server Render has access to the launch darkly flag values in initial state and will use that to determine whether or not render the "newCSSStyles".

In the case that the launch darkly flag value is true, we want to render BOTH the old CSS and the new CSS.

### DIFF
--- a/packages/mwp-core/src/renderers/server-render.jsx
+++ b/packages/mwp-core/src/renderers/server-render.jsx
@@ -338,31 +338,6 @@ const makeRenderer = (
 
 		return routesPromise.then(resolvedRoutes =>
 			initializeStore(resolvedRoutes).then(store => {
-				if ('skeleton' in request.query) {
-					// TODO: what is the skeleton? Do we need to also check the ld flags for css styles here too?
-					// Check launch darkly flags to determine if we should
-					// be rendering the new swarm global styles as well.
-					// The new global styles must be loaded after the old css styles
-					const initialState = store.getState();
-					const ldStyleFlag = initialState.flags['new-swarm-assets'];
-					if (ldStyleFlag) {
-						cssLinks = [...cssLinks, ...newCSSLinks];
-					}
-
-					// render skeleton if requested - the store is ready
-					return {
-						result: getHtml(
-							<Dom
-								basename={basename}
-								head={Helmet.rewind()}
-								initialState={store.getState()}
-								scripts={scripts}
-								cssLinks={cssLinks}
-							/>
-						),
-						statusCode: 200,
-					};
-				}
 				// the initial addFlags call will only be key'd by member ID
 				return addFlags(store, { id: parseMemberCookie(state).id })
 					.then(() => populateStore(store))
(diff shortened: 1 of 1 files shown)


packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -327,6 +339,16 @@ const makeRenderer = (
 		return routesPromise.then(resolvedRoutes =>
 			initializeStore(resolvedRoutes).then(store => {
 				if ('skeleton' in request.query) {
+					// TODO: what is the skeleton? Do we need to also check the ld flags for css styles here too?
```
sadafie: cc @mmcgahan --- I can't remember the purpose of `skeleton`. Do we need to include the LD new style conditions here as well, or should I remove?

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -156,6 +157,14 @@ const getRouterRenderer = ({
 		return redirect;
 	}
 
+	// Check launch darkly flags to determine if we should
+	// be rendering the new swarm global styles as well.
+	// The new global styles must be loaded after the old css styles
+	const ldStyleFlag = initialState.flags['new-swarm-assets'];
```
mmcgahan: I'd really like to avoid putting application-specific code into this function - e.g. any reference to a particular feature flag name -  because it doesn't solve for the general case that the platform should aim to support.

Instead, I was wondering if we could update how `cssLinks` is read, e.g. by allowing it to be _either_ an array _or_ a function. If it's a function (`typeof cssLinks === 'function'`), then it could be treated as a selector that takes `initialState` as an argument and returns an array of links. That would allow the caller to essentially compose `cssLinks` in whatever way it wants based on `state`, just like a `mapStateToProps` function.

There might be some other options, but essentially we want to be able to make it the responsibility of the caller to provide the right assets in one shot rather than composing internally if possible.

packages/mwp-core/src/renderers/server-render.jsx
mmcgahan: actually, if you could just remove the whole `if ('skeleton')` section, that's probably best - it's a vestige of some offline-first support that we experimented with early on, but it never took root so we should roll it back

sadafie:
@mmcgahan @hsbacot Can I get a final look over? We're thinking of merging this today. Thanks.

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -185,7 +192,7 @@ const makeRenderer$ = (renderConfig: {
 	middleware: Array<Function>,
 	scripts: Array<string>,
 	enableServiceWorker: boolean,
-	cssLinks: ?Array<string>,
+	cssLinks: ?(Array<string> | Function),
```
mmcgahan: usually it's good to avoid `Function` types since they're not very specific. Try

```js
?MWPState => Array<string>
```

hsbacot:
LGTM

PR #593, opened 2019-01-30
test for window.performance.getEntriesByType

sadafie:


sadafie:
@chenrui333 We're in progress moving these to a new mono repo and deprecating the functions there. This is a quick fix for now. We will definitely add complete tests in the new repo.

PR #596, opened 2019-01-30
Upgrade Lerna

sadafie:
Upgrade + hopefully fix the flakey npm publishing issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ after_success:
     - chmod 600 /tmp/github_deploy_key
     - ssh-add /tmp/github_deploy_key
     - >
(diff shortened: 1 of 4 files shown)


mmcgahan:
@sadafie major version bumps should probably get some comment on what was changed and why we're not making any other changes to account for the changes?

sadafie:
@mmcgahan Sorry all, I should have marked this prelim. The upgrade is not working/complete yet. This was a branch that Jackie and I created while pairing -- using travis to test `lerna publish` was the only way we could think of to test changes.

sadafie:
this doesn't work and is so old that i'm going to close it and hopefully we can revisit in the future.

PR #597, opened 2019-02-08
Mes 19 add sift sessionid mupweb

padsul:


packages/mwp-core/src/util/cookieUtils.js
```diff
@@ -30,6 +34,14 @@ export const parseBrowserIdCookie = state => {
 	return browserId;
 };
 
+export const parseSiftSessionCookie = state => {
+	if (!state[SIFT_SESSION_COOKIE]) {
+		return '';
+	}
+	const siftSessionId = querystring.parse(state[SIFT_SESSION_COOKIE]);
```
mmcgahan: Looks like tests are failing because you're using `querystring.parse` on a plain UUID string (it's not querystring-formatted). If you remove that, then the cleanest solution is

```js
export const parseSiftSessionCookie = state => state[SIFT_SESSION_COOKIE] || ''
```

padsul: Make sense. Update PR as suggested.

packages/mwp-core/src/util/cookieUtils.js
```diff
@@ -11,6 +11,10 @@ export const BROWSER_ID_COOKIE = appConfig.api.isProd
 	? 'MEETUP_BROWSER_ID'
 	: 'MEETUP_BROWSER_ID_DEV';
 
+export const SIFT_SESSION_COOKIE = appConfig.api.isProd
```
jaclynj: This is a session ID cookie specific to Sift? Maybe add a comment here as to what this is and how it's being used, why its Sift specific. Link to documentation would be helpful.

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -272,6 +276,9 @@ const makeRenderer = (
 					media: getMedia(userAgent, userAgentDevice),
 					browserId: parseBrowserIdCookie(state),
 					clientIp,
+					siftSessionId: parseSiftSessionCookie(state), // userSession cookie created in fastly and used by sift science to correlate spammy
+					// behavior to user activity; cookie has an expiration time of 4 hours and should be
+					// reset on logout
```
jaclynj: Adding a link to documentation or at least to Sift would be helpful.

mmcgahan: +1 - also let's try to keep 'sift' cookie documentation in one place if possible - add comments to function in `cookieUtils`, but don't specify cookie expiration, because that's 'owned'/specified elsewhere.

packages/mwp-core/src/util/cookieUtils.js
padsul: There has been already discussion on naming convention, for details please ref slack communication: https://meetuphq.slack.com/archives/C099CDFJS/p1549032521177100

packages/mwp-core/src/renderers/server-render.jsx
padsul: Make sense. Updated in recent commit.

packages/mwp-core/src/util/cookieUtils.js
mmcgahan: @padsul I think @jaclynj is requesting a link to docs in a code comment so that future devs understand where this is coming from.

packages/mwp-core/src/util/cookieUtils.js
```diff
@@ -11,6 +11,13 @@ export const BROWSER_ID_COOKIE = appConfig.api.isProd
 	? 'MEETUP_BROWSER_ID'
 	: 'MEETUP_BROWSER_ID_DEV';
 
+// SIFT_SESSION_ID cookie created in fastly and used by sift science to correlate spammy
+// behavior to user activity; cookie has an expiration time of 4 hours and should be
```
mmcgahan: this code doesn't really control the business rules (4 hour expiration, reset on logout), so in general it's better not to specify things like that separately from where they're defined - they inevitably will get out of sync. All this code really needs to know/communicate is how to parse the cookie.

PR #603, opened 2019-03-27
replace regex string replacement w/ formatter for offset formatting

chrislewis:


packages/mwp-tracking-plugin/src/util/trackingUtils.js
```diff
@@ -61,7 +61,7 @@ export const parseIdCookie = (
 // `toString()` adds the timezone in brackets, which we want to remove
```
mmcgahan: the code comments here are no longer relevant, but might be worth keeping an example of the expected output. The [js-joda API docs](https://js-joda.github.io/js-joda/class/src/format/DateTimeFormatter.js~DateTimeFormatter.html) don't currently mention `ISO_OFFSET_DATE_TIME`, so it might be hard for maintainers to know what is expected. I might try to submit a PR for those docs, because I definitely remember looking for a better approach to formatting this string when I first added this code...

chrislewis: Yeah - I was aware of the formatters because jsJoda is a port of the Java library, joda time ([relevant formatting instance](https://www.joda.org/joda-time/apidocs/org/joda/time/format/ISODateTimeFormat.html#basicDateTime--)). Formatting in jsJoda is [discussed](https://js-joda.github.io/js-joda/manual/formatting.html), but it's true that the provided formatters, namely [ISO_OFFSET_DATE_TIME](https://github.com/js-joda/js-joda/blob/master/src/format/DateTimeFormatter.js#L658-L662), aren't documented. That is a gap in the library - I'll look into how to contribute docs, or reach out requesting docs there.

chrislewis: Updated the comment. Sorry about that!

PR #606, opened 2019-04-11
fix: standardized_url and standardized_referrer not showing in activity logs

mmcgahan:
https://meetup.atlassian.net/browse/WP-1096

two issues:

1. the `matches` value is just an array, not an object with a `matched` property
2. the `standardized_url` fields were not being passed to the serializer.

Also discovered that the standardized urls were not being set correct for lazy requests, so I made the 'activity field setter' a config option on the route, which means that the special case handling for the api proxy route can be defined by the route itself - better separation of concerns.

### DIFF
--- a/packages/mwp-tracking-plugin/src/_activityTrackers.js
+++ b/packages/mwp-tracking-plugin/src/_activityTrackers.js
@@ -45,7 +45,7 @@ export const getTrackApiResponses: TrackGetter = trackOpts => request => (
  * 3. lazy-loaded data
  *    - url: proxy endpoint path (request.url.path)
  *    - referrer: current URL (request.referrer)
- * 4. tracking-only request:
+ * 4. tracking-only request
  *    - url: proxy endpoint path (request.url.path)
  *    - referrer: current URL (request.referrer)
  */
@@ -56,7 +56,7 @@ export const getTrackActivity: TrackGetter = () => (request: HapiRequest) => (
 	// proxy endpoint that should be tracked differently
 	const { getFields } =
 		request.route.settings.plugins[ACTIVITY_PLUGIN_NAME] || {};
-	console.log('request.route', request.route, getFields);
+	console.log('request.route', request.route.path, getFields);
 	const trackFields = getFields
 		? getFields(request, fields)
 		: {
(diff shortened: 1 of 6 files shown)


mmcgahan:
underlying issue fixed, but lazy requests aren't getting `standardized_referer`, and the `standardized_url` is for the referring page, not `/mu_api` or whatever is configured - that should be addressed here

packages/mwp-tracking-plugin/src/_activityTrackers.js
```diff
@@ -13,21 +13,22 @@ type ActivityFields = {
...
+	const { url = '', referrer = '', ...fieldLiterals } = fields;
 	return trackOpts.log(request, {
 		description: 'nav',
 		memberId: parseIdCookie(request.state[trackOpts.memberCookieName], true), // read memberId
 		browserId: updateId(trackOpts.browserIdCookieName)(request), // read/add browserId
 		trackId: updateId(trackOpts.trackIdCookieName)(request), // read/add trackId
 		referer: referrer,
 		url,
-		viewName,
-		subViewName,
```
jaclynj: Just checking, does it matter that these two values are no longer guaranteed to be here (if they aren't in the fields in the first place)

mmcgahan: good check, but yes it's fine if they are not returned

packages/mwp-tracking-plugin/src/_activityTrackers.js
```diff
@@ -48,29 +49,19 @@ export const getTrackApiResponses: TrackGetter = trackOpts => request => (
...
-		? {
-				url: requestReferrer, // original request referrer is passed through as URL
-				referrer: referrerOverride,
-			}
+	// route may specify a custom 'getFields', which usually means that it is a
+	// proxy endpoint that should be tracked differently
+	const { getFields } =
+		request.route.settings.plugins[ACTIVITY_PLUGIN_NAME] || {};
+	const trackFields = getFields
+		? getFields(request, fields)
```
sadafie: Just verifying that in this special case, we don't care about explicitly tracking the url and the referrer any longer?

mmcgahan: correct - `getFields` is expected to fully handle those values. I'll add some docs before merging

PR #607, opened 2019-04-16
Wp 1065 app context provider withintl

mmcgahan:
This is a refactor of how to make withIntl work with arbitrary child components.

The approach is to treat `state.config` in a special way by making it available outside of Redux, in a statically-defined `AppContext` context, which is directly set on the server through `<ServerApp appContext />` and indirectly on the browser by reading the value from `window.APP_RUNTIME.appContext`

This approach would also be useful for `connectWithMatchMedia`, which only relies on `state.config.media` - we can remove the Redux dependency by using `AppContext` in that HOC as well.


packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -171,13 +171,14 @@ const getRouterRenderer = ({
 	// so go ahead and assemble the full response body
 	const result = getHtml(
 		<Dom
-			basename={basename}
+			basename={appContext.basename}
```
mmcgahan: `basename` no longer a prop for `Dom` - still need to update implementation of `Dom` to remove `userAgent` prop as well - just read from `appContext`.

packages/mwp-core/src/renderers/browser-render.jsx
```diff
@@ -55,7 +55,7 @@ export function resolveAppProps(
 	return resolveComponents().then(() => ({
```
mmcgahan: add code comment about what this return object is for - BrowserApp props

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -132,14 +131,15 @@ const getRouterRenderer = ({
...
 	let appMarkup;
-	const staticContext: { url?: string, permanent?: boolean } = {};
+	const routerContext: { url?: string, permanent?: boolean } = {};
 
 	try {
 		appMarkup = ReactDOMServer.renderToString(
 			<ServerApp
-				basename={basename}
+				appContext={appContext}
+				basename={appContext.basename}
```
mmcgahan: this prop not needed for ServerApp - it reads from appContext

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -171,13 +171,14 @@ const getRouterRenderer = ({
...
-			basename={basename}
+			basename={appContext.basename}
 			head={sideEffects.head}
 			initialState={initialState}
+			appContext={appContext}
 			appMarkup={appMarkup}
 			scripts={scripts}
 			cssLinks={cssLinks}
-			userAgent={userAgent}
+			userAgent={appContext.userAgent}
```
mmcgahan: same here

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -205,75 +247,26 @@ const makeRenderer = (renderConfig: {
...
 }) => {
-	const {
-		routes,
-		reducer,
-		middleware,
-		scripts,
-		enableServiceWorker,
-		cssLinks,
-	} = renderConfig;
+	const { routes, reducer, middleware, scripts, cssLinks } = renderConfig;
```
mmcgahan: keep enableServiceWorker in this destructuring assignment

packages/mwp-i18n/src/withIntl.jsx
```diff
@@ -53,12 +51,24 @@ export default (
...
 	};
+
+	// Except in testing, wrap the component in an AppContext provider in order
+	// to pass in the correct 'requestLanguage'
 	const ConnectedWithIntl =
-		process.env.NODE_ENV === 'test' // avoid Redux context dependency in tests
+		process.env.NODE_ENV === 'test' // avoid AppContext dependency in tests
 			? WithIntl
-			: connect(mapStateToProps, mapDispatchToProps)(WithIntl);
+			: props => (
```
mmcgahan: define this function component at higher level

mmcgahan:
This pattern can also be used to refactor ServiceWorker.jsx (needs enableServiceWorker and requestLanguage values)

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -328,20 +326,19 @@ const makeRenderer = (renderConfig: {
 		return routesPromise.then(resolvedRoutes =>
 			initializeStore(resolvedRoutes).then(store => {
 				// the initial addFlags call will only be key'd by member ID
-				return addFlags(store, { id: parseMemberCookie(state).id })
+				return addFlags(store, { id: parseMemberCookie(request.state).id })
```
sadafie: was this a bug before? or am i missing where this change is coming from?

flow-typed/platform.js
```diff
@@ -7,44 +7,51 @@ declare type MatchMedia = {
 	isAtLargeUp: boolean,
 };
 
+declare type AppContext = {
+	apiUrl: string, // base API URL for queries
```
sadafie: Would it be useful to add `location` to AppContext as well?

packages/mwp-i18n/src/withIntl.jsx
```diff
@@ -53,16 +51,31 @@ export default (
...
-	const ConnectedWithIntl =
-		process.env.NODE_ENV === 'test' // avoid Redux context dependency in tests
-			? WithIntl
-			: connect(mapStateToProps, mapDispatchToProps)(WithIntl);
 
-	// modify display name to hide internal 'connect' implementation
+	// define component that consumes requestLanguage from context
+	const ContextEnhancedWithIntl = props => (
+		<AppContext.Consumer>
+			{appContext => {
```
sadafie: Does this need to be `props.appContext`?

packages/mwp-app-render/src/components/shared/PlatformApp.jsx
```diff
@@ -3,22 +3,19 @@ import { Provider } from 'react-redux';
...
-				</ServiceWorker>
-			</Provider>
-		);
-	}
-}
+// Special context provider for config values supplied by server
+export const AppContext = React.createContext();
+
+const PlatformApp = props => (
+	<AppContext.Provider value={props.context}>
```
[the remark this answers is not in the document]
mmcgahan: yeah I was debating that - once we're in the `PlatformApp`, it's clear we're in the 'app', but yeah naming consistency end to end is probably still valuable - will update

flow-typed/platform.js
mmcgahan: it might help reduce some argument passing in the server render code, but it would be misleading in the client-side application because it wouldn't be clear that `state.config.location` refers only to the server-rendered page, even when the user navigates elsewhere. I think it's probably better to leave it out

packages/mwp-core/src/renderers/server-render.jsx
mmcgahan: `state` was assigned as a `const` destructured from `request` in the old code (on line 228), but I got rid of that destructuring because it added a lot more variable names to keep track of in this scope, and `state` is ambiguous in code that touches both the Hapi `request` _and_ the Redux store - I wanted to make the reference more explicitly tied to the `request` parent.

PR #610, opened 2019-05-01
WP-1125 conventional commit linting

mmcgahan:
Conventional commit linting and other linting/githook updates/upgrades

- config moved to individual JS files where possible (commitlint, husky, lint-staged). This keeps `package.json` better defined, and the config more discoverable.
- minor updates to support breaking changes of eslint


sadafie:
Actually I just thought of a question -- should we be removing the `Makefile` pkg versions + corresponding logic from this?

mmcgahan:
@sadafie I thought about that, but then realized it was going to introduce a lot of effort figuring out our versioning strategy and workflow, so it would take a lot more time. I think it's good to have this in place for now and we can build on it when we refactor our packages in response to the fedx work

sadafie:
@mmcgahan Ah gotcha - so we're not doing the lerna publish step with conventional commits yet, then? Got it 👍

PR #612, opened 2019-05-06
WP 1178 SetCookie component for replatform redirect

mmcgahan:
**`Prelim` while I use `yarn link` to test locally**

In order to use mup-web as the source of backend truth for migrations from chapstick, we need to be able to set cookies on the server in addition to trigger redirects.

The new `SetCookie` component enables this behavior by storing new cookie settings in side effects during server rendering and applying corresponding `set-cookie` headers. In the browser, the cookie is set using `js-cookie`.

**Open question**: should there be some sort of `noOverwrite` prop to prevent the cookie from being re-set on every render/request? It would be relatively easy to implement, but I'm not sure how useful it would be - the behavior at the moment is to overwrite the cookie every time the app renders.


flow-typed/hapi.js
```diff
@@ -192,10 +235,21 @@ declare type HapiResponseToolkit = {
...
+		options?: HapiServerStateCookieOptions
+	) => HapiResponseObject,
 	unauthenticated: (error: Error, data?: HapiAuthenticationData) => void,
-	unstate: (key: string, options?: HapiServerStateCookieOptions) => HapiResponseObject,
-};
\ No newline at end of file
+	unstate: (
+		key: string,
+		options?: HapiServerStateCookieOptions
+	) => HapiResponseObject,
```
jaclynj: Did anything change in this file? It looks like just formatting but if not, please call out what changed.

flow-typed/platform.js
```diff
@@ -58,15 +58,33 @@ declare type FeatureFlags = { [string]: boolean | string };
 
 declare type Params = { [string]: string };
 
+type CookieOpts = {|
+	value: string,
+	ttl?: number | null, // milliseconds
+	isHttpOnly?: boolean,
+	isSecure?: boolean,
+	isSameSite?: false | 'Strict' | 'Lax',
```
jaclynj: Might be good to comment what the different options here mean

packages/mwp-cookie/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+#### Client-side-only cookie
+
+The `<SetCookie />` element is generally the wrong abstraction for setting cookies
+on the client, since client-side cookie setting is by definition done with
+callbacks (e.g. `componentDidMount`, `onClick`) that do not need to interact
+with app rendering.
+
+We recommend using [js-cookie](https://github.com/js-cookie/js-cookie/) in these
+cases, with its `SetCooke.set()` method.
```
jaclynj: Good docs!

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -66,6 +67,7 @@ const resolveSideEffects = () => ({
 	redirect: Redirect.rewind(),
 	forbidden: Forbidden.rewind(),
 	notFound: NotFound.rewind(),
+	cookie: SetCookie.rewind(),
```
jaclynj: You might want to also link to `https://github.com/gaearon/react-side-effect` in the comments above

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -97,23 +99,8 @@ const getMedia = (userAgent: string, userAgentDevice: string) => {
...
- * it sets the data needed by other parts of the DOM, such as `<head>`.
- * 2. `htmlMarkup`, which wraps `appMarkup` with the remaining DOM markup.
- * 3. `doctype`, which is just the doctype element that is a sibling of `<html>`
- *
- * @param {Object} renderProps
- * @param {ReduxStore} store the store containing the initial state of the app
- * @return {Object} the statusCode and result used by Hapi's response API
+ * HTML string and server response code, with optional cookies to write
  */
-type RenderResult =
```
jaclynj: If this type is removed here, what is the return type being used here https://github.com/meetup/meetup-web-platform/pull/612/files#diff-d7329f1db146f21557dc646dcd0d2350R112

flow-typed/hapi.js
mmcgahan: I think I added something here, saved it, and then later removed it, so just the Prettier formatting changes were left - I'll see if I can flip it back to the `master` version so it doesn't show in the diff.

flow-typed/platform.js
mmcgahan: This is sort of intentionally undocumented because it's only here to prevent Flow from yelling, but actually I think I can remove it again.

mmcgahan:
@jaclynj and any other potential reviewers: after trying out this code in mup-web, the `SetCookie` component _module_ was being instantiated twice, which made it impossible to only track the side effects from the intended component. Although I still don't fully understand the mechanism that caused the shadow instance, it _is_ a known risk when using `react-side-effect` and is related to the fact that state is created in a library-supplied HOC. Essentially, this exposes us to major problems if there are multiple versions of the library in the app, and apparently also when the app is bundled in a particular way.

To avoid practicing too many black magic incantations to make it 'just work' again, I am going to significantly refactor `<SetCookie />` to use React Context that will allow the cookie-setting function to be injected into the app at the top level and accessed from nested consumers.

**Please do not review until the `Prelim` label is removed**

packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -97,25 +97,12 @@ const getMedia = (userAgent: string, userAgentDevice: string) => {
 
 /**
  * Using the current route information and Redux store, render the app to an
- * HTML string and server response code.
```
mmcgahan: these comments haven't been valid for awhile - cleaning up

packages/mwp-core/src/renderers/server-render.jsx
mmcgahan: this is defined in `flow-typed/platform.js` - looks like this is just an old re-declaration of the same type available globally

mmcgahan:
It works! TODO: docs

mmcgahan:
Ready for final review

PR #615, opened 2019-06-07
mwp-config upgrade to babel 7

sadafie:
Some root level updates as well. Installs, builds, and test pass. 

One thing I'm unsure about is if this will be a breaking change in mwp-cli and mup-web (specifically conflicts with babel 6). I'm going to try updating and testing locally to see. I'd like to be able to merge this independently of the mwp-cli and mup-web upgrades if possible.


packages/mwp-config/package.json
```diff
@@ -15,15 +15,15 @@
...
-        "babel-preset-react": "6.24.1",
-        "babel-preset-stage-2": "6.24.1",
+        "@babel/core": "7.4.4",
+        "@babel/plugin-proposal-class-properties": "7.4.4",
+        "@babel/plugin-transform-flow-strip-types": "7.4.4",
+        "@babel/plugin-transform-runtime": "7.4.4",
+        "@babel/preset-env": "7.4.4",
+        "@babel/preset-flow": "7.0.0",
+        "@babel/preset-react": "7.0.0",
+        "@babel/preset-stage-2": "7.0.0",
```
mmcgahan: For presets, we should no longer need `preset-stage-2`, and just use the presets + plugins used for MWP in [`babel.config.js`](https://github.com/meetup/meetup-web-platform/blob/master/babel.config.js)

- `preset-env` 
- `preset-react`

for `preset-flow`, you can _either_ used the preset _or_ the `@babel/plugin-transform-flow-strip-types` plugin, but you don't need both https://babeljs.io/docs/en/babel-preset-flow

PR #624, opened 2019-07-16
feat: add `isBot` to state

bgruber:


packages/mwp-core/src/renderers/server-render.jsx
```diff
@@ -213,6 +214,7 @@ const getAppContext = (request: HapiRequest, enableServiceWorker: boolean) => {
 		browserId: parseBrowserIdCookie(state),
 		clientIp,
 		siftSessionId: parseSiftSessionCookie(state),
+		isBot: isBot(request.headers['user-agent']),
```
mmcgahan: the one place where it would be helpful to add this type is in https://github.com/meetup/meetup-web-platform/blob/master/flow-typed/platform.js#L48

```jsx
  isBot: boolean
```

PR #625, opened 2019-07-18
feat: add mupweb packages path

andeliv:
Add only `lib` folder from each mupweb-package to babel-loader include config instead of whole `/packages/` folder


mmcgahan:
The code change looks fine, but I'd like more context on the motivation for this change - was there an error, or does this improve performance in some way?

PR #626, opened 2019-07-23
Separate DIY edge from Shared edge request builders

mmcgahan:
https://www.pivotaltracker.com/story/show/167407231

This PR separates the 'request arg builder' functions for DIY edge (non-api.meetup.com) and Shared edge (api.meetup.com) requests, since the shared edge needs to support features that are not supported by the shared API/chapstick.

In particular, this separation includes a feature that strips the 'click-track' cookie from the request so that the request payload can be significantly smaller - this will avoid errors in variants service requests.

Also, remove unused `jar` code for allowing API to set cookies on the client - that was originally set up for the `/sessions` endpoint for login, but that feature was never replatformed and we don't need to carry the tech debt.

### DIFF
--- a/packages/mwp-api-proxy-plugin/src/util/receive.js
+++ b/packages/mwp-api-proxy-plugin/src/util/receive.js
@@ -2,7 +2,6 @@ import isBot from 'isbot';
 import querystring from 'qs';
 import { logger } from 'mwp-logger-plugin';
 
-import { API_PROXY_PLUGIN_NAME } from '../config';
 import { coerceBool, toCamelCase } from './stringUtils';
 
 // match escpaed unicode characters that are treated as newline literals in JS
@@ -173,11 +172,7 @@ export const makeParseApiResponse = query => ([response, body]) => {
  *
  * @param {Object} apiResponse JSON-parsed api response data
  */
-export const makeApiResponseToQueryResponse = query => ({
-	value,
-	error,
-	meta,
-}) => ({
+export const makeApiResponseToQueryResponse = query => ({ value, error, meta }) => ({
 	type: query.type,
 	ref: query.ref,
 	value,
@@ -188,7 +183,11 @@ export const makeApiResponseToQueryResponse = query => ({
 export const makeLogResponse = request => ([response, body]) => {
 	// `response` may contain private (user-specific) data. Log with caution
 	const {
-		request: { headers, method, uri: { href: url } },
+		request: {
+			headers,
+			method,
+			uri: { href: url },
+		},
 		statusCode,
 	} = response;
 	const logBase = {
@@ -204,9 +203,7 @@ export const makeLogResponse = request => ([response, body]) => {
 			// don't log errors from bots - e.g. for deleted groups/events/whatever
 			return;
 		}
-		const logError = (statusCode < 500 ? logger.warn : logger.error).bind(
-			logger
-		);
+		const logError = (statusCode < 500 ? logger.warn : logger.error).bind(logger);
 		let errorMessage;
 		try {
 			// well-behaved API errors return a JSON object with an `errors` array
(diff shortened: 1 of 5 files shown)


packages/mwp-api-proxy-plugin/src/util/receive.js
```diff
@@ -234,44 +231,15 @@ export const makeLogResponse = request => ([response, body]) => {
 	}
 };
 
-/**
- * When a tough-cookie cookie jar is provided, forward the cookies along with
- * the overall /api response back to the client
- */
-export const makeInjectResponseCookies = request => ([response, _, jar]) => {
```
mmcgahan: no longer need to inject response cookies

packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -8,7 +8,7 @@ import externalRequest from 'request';
 import config from 'mwp-config';
 
 export const API_META_HEADER = 'X-Meta-Request-Headers';
-const FULL_URL_PATTERN = /^https?:\/\//;
+const IS_DIY_EDGE_URL_PATTERN = /^https?:\/\//;
```
mmcgahan: renaming to make the purpose more clear - we may want to update this at some point to specifically exclude https://api.meetup.com URLs, but there's no particular reason for people to specify the full URL in that case.

packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -82,54 +82,79 @@ function makeMockResponse(requestOpts, response = MOCK_RESPONSE_OK) {
...
+		const url = new URL(query.endpoint);
+
+		switch (externalRequestOpts.method) {
+			case 'patch':
+			case 'put':
+			case 'post':
+				if (externalRequestOpts.formData) {
+					break;
+				}
+				// assume all DIY edge endpoints handle JSON body encoding
```
mmcgahan: I think this is a fair assumption, but we'll need to watch e2e's

packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -333,22 +378,20 @@ export const makeMockRequest = (
  * Make a real external API request, return response body string
  */
 export const makeExternalApiRequest = request => requestOpts => {
-	return externalRequestPromise(requestOpts)
```
mmcgahan: no changes in this block - just Prettier doing its thing

packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -82,54 +82,79 @@ function makeMockResponse(requestOpts, response = MOCK_RESPONSE_OK) {
...
+ * Currently, cookies cannot be set by DIY edge endpoints. (e.g. using the
+ * `jar` interface in `request`)
  */
-export const createCookieJar = requestUrl => {
-	const parsedUrl = url.parse(requestUrl);
-	if (parsedUrl.pathname === '/sessions') {
-		return externalRequest.jar(); // create request/url-specific cookie jar
-	}
-	return null;
+const buildDIYEdgeRequestArgs = externalRequestOpts => {
```
sadafie: I'm just slightly confused with naming -

* `DIY Edge` = non api.meetup.com
* `Shared Edge` = api.meetup.com (chapstick?)
* `External Request` = all requests, whether DIY or Shared?

I ask because I see that `buildDIYEdgeRequestArgs()` and `buildSharedEdgeRequestArgs()` both take an `externalRequestOpts` parameter. `external` to me tends to mean non-meetup endpoints, but it seems to refer all outbound requests? If that's the case, a nitpick suggestion would be to rename `externalRequestOpts` to simply `requestOpts`.

sadafie: Tangent question for my own knowledge - where do non meetup endpoints fall in this request structure? Are the handled separately elsewhere?

packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -82,54 +82,79 @@ function makeMockResponse(requestOpts, response = MOCK_RESPONSE_OK) {
...
+					break;
+				}
+				// assume all DIY edge endpoints handle JSON body encoding
+				body = JSON.stringify(query.params);
+				break;
+			case 'delete':
+			case 'get':
+			default: {
+				// copy query object params into `url.searchParams`
+				for (const p of new URLSearchParams(query.params)) {
```
sadafie: Does having the `new` operator inside the for-of statement cause it to create a new object on every iteration of the loop? In examples, I've normally seen:

`const param = new URLSearchParams...; for (const p of param) ...`

packages/mwp-api-proxy-plugin/src/util/send.js
sadafie: Could you elaborate a bit on _why_ the DIY edge url only tests for https? Is it because the Shared Edge and other external requests don't use https? Or is it application implementation specific - `https://` full URLS are specified only for DIY Edge.

mmcgahan: I guess naming it DIY edge is actually more confusing because this pattern could be used for _any_ endpoint, even a third party one, so I'll update the name. Essentially, if you specify the full host (protocol + domain) in the `endpoint`, it means you do _not_ want to use the default `baseUrl` of `https://api.meetup.com`. This regex checks for the `https://` protocol at the start of the string to check whether the endpoint is pointing to a non-default host (essentially, if you specify the protocol, you're specifying a fully-qualified URL).

packages/mwp-api-proxy-plugin/src/util/send.js
mmcgahan: Probably a bit more clear after discussing the DIY Edge definition, but 'external' is actually referring to 'outside of the API proxy server' - _any_ API request is an external request. That `externalRequestOpts` object was initially named to correspond to the `request` package, which is imported as `externalRequest` in this module - it's the module that makes the HTTP request from the proxy to the network. I wish it was named less ambiguously...

packages/mwp-api-proxy-plugin/src/util/send.js
mmcgahan: it's not evaluating the second part of `of` every loop, no - same way it doesn't re-evaluate `param` in your example. Examples break it up into two lines so that you can see/re-use the `URLSearchParams` object after the loop, but it's intended to be ephemeral here

packages/mwp-api-proxy-plugin/src/util/send.js
mmcgahan: > where do non meetup endpoints fall in this request structure? Are the handled separately elsewhere?

They're handled here - the proxy makes no assumptions about the responses other than that they are JSON-encoded, so non-meetup requests get mixed in with meetup requests in the query response array.

PR #627, opened 2019-07-31
Hapi 18 upgrade

sadafie:
Following the release notes / migration steps here:

https://github.com/hapijs/hapi/issues/3871


packages/mwp-api-proxy-plugin/src/util/send.js
```diff
@@ -292,7 +292,7 @@ export function getTrackingHeaders(request) {
 	if (trackingParam) {
 		return {
 			'X-Meetup-External-Track': trackingParam,
-			'X-Meetup-External-Track-Url': `${host}${request.url.href}`,
+			'X-Meetup-External-Track-Url': `${host}${request.url.pathname}${request.url.search}`,
```
sadafie: in Hapi 17 `request.url.href` was usually, but not always, a relative path. In Hapi 18, following the WHATWG url api, `request.url.href` is always an absolute path. Therefore, changing to use `pathname` + `search` should give the equivalent of what we need.

packages/mwp-logger-plugin/src/logger.js
```diff
@@ -44,7 +44,7 @@ export const httpRequestSerializers = {
 			requestInfo.responseSize = request.response.headers['content-length'];
 			requestInfo.status = request.response.statusCode;
 			requestInfo.latency = formatDuration(
-				request.info.responded - request.info.received
+				request.info.completed - request.info.received
```
sadafie: In Hapi 18, `request.info.responded` is a  non-0 value (timestamp) only if the response was successfully sent.

> If you access request.info.responded (usually in logging logic), consider switching to request.info.completed if you want to know when the server finished processing and use the existence of request.info.responded to log if a response was successful.

packages/mwp-tracking-plugin/src/_activityTrackers.js
```diff
@@ -37,31 +37,30 @@ export const getTrackApiResponses: TrackGetter = trackOpts => request => (
...
 	// proxy endpoint that should be tracked differently
-	const { getFields } =
-		request.route.settings.plugins[ACTIVITY_PLUGIN_NAME] || {};
+	const { getFields } = request.route.settings.plugins[ACTIVITY_PLUGIN_NAME] || {};
 	const trackFields = getFields
 		? getFields(request, fields)
 		: {
 				...fields,
-				url: request.url.path,
+				url: request.url.pathname,
```
sadafie: `request.url.path` is no longer supported. use `request.path` or `request.url.pathname` instead

packages/mwp-tracking-plugin/src/click.test.js
```diff
@@ -40,7 +40,12 @@ describe('onPreHandlerExtension', () => {
 		const injectRequest = {
 			method: 'get',
 			url: '/foo',
-			credentials: 'whatever',
+			auth: {
+				credentials: {
+					user: 'whatever',
+				},
+				strategy: 'bypass',
```
sadafie: > In order to fully support request injection authentication as a production-ready feature, the injection options were changed to require passing the name of the strategy being simulated to authenticate the request. Previously, a special 'bypass' strategy name was set for all injected credentials. To use the new format simply wrap the credentials and optional artifacts with an auth object and add a new strategy key with a name matching a configured authentication strategy.

Basically, we still want to bypass auth, so updated the credentials object for every `server.inject()` to explicitly pass the `bypass` strategy. Not recommended for prod code, but fine for testing.

sadafie:
Yep, seems to be running well in `mup-web`! Going to merge.

PR #634, opened 2019-08-09
Fix language path redirects

sadafie:
In the original migration, I wrongly translated this. My expectation was that I was calling `.toString()` on a URL object, but obviously it's not an instance of a URL object so calling .toString() on `Object` returns [object%20Object]

### DIFF
--- a/packages/mwp-app-route-plugin/src/handler.js
+++ b/packages/mwp-app-route-plugin/src/handler.js
@@ -1,5 +1,4 @@
 // @flow
-import url from 'url';
 
 /*
  * This is the Hapi route handler that will be applied to the React application
@@ -21,7 +20,8 @@ export default (languageRenderers: { [string]: LanguageRenderer }): HapiHandler 
 ) => {
 	const pathname = request.getLangPrefixPath();
 	if (pathname !== request.url.pathname) {
-		return h.redirect(url.format({ ...request.url, pathname }));
+		// redirect to a host relative path, keeping query params intact
+		return h.redirect(`${pathname}${request.url.search}`);
 	}
 	const requestLanguage = request.getLanguage();
 	const renderRequest = languageRenderers[requestLanguage];
--- a/packages/mwp-app-route-plugin/src/handler.js
+++ b/packages/mwp-app-route-plugin/src/handler.js
@@ -1,4 +1,5 @@
 // @flow
+import url from 'url';
 
 /*
  * This is the Hapi route handler that will be applied to the React application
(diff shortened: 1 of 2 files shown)


packages/mwp-app-route-plugin/src/handler.js
```diff
@@ -20,7 +21,7 @@ export default (languageRenderers: { [string]: LanguageRenderer }): HapiHandler
 ) => {
 	const pathname = request.getLangPrefixPath();
 	if (pathname !== request.url.pathname) {
-		return h.redirect({ ...request.url, pathname }.toString());
```
mmcgahan: this is fine, but I think it might be a bit cleaner to just use the `URL` interface and modify `pathname` on a new copy, e.g.

```js
const newUrl = new URL(request.url); // make a copy
newUrl.pathname = pathname;
return h.redirect(newUrl.toString());
```

packages/mwp-app-route-plugin/src/handler.js
```diff
@@ -20,7 +21,7 @@ export default (languageRenderers: { [string]: LanguageRenderer }): HapiHandler
...
+		return h.redirect(url.format({ ...request.url, pathname }));
```
mmcgahan: I'm not sure this will keep querystrings intact - e.g. in the Node REPL

```js
> foo = new URL('https://example.com/foo?bar=baz')
> url.format({ ...foo, pathname: 'newpath' })
'newpath'
```

mmcgahan: I think we need some unit tests for this to cover different URL cases - with/without querystring, with/without existing `pathname`

mmcgahan: Looking at this more, I don't think you can validly use the spread operator on a `URL` instance

mmcgahan: investigating a bit more, there are a few constraints

1. the redirect should be a host-relative URL, not a fully-qualified URL (e.g. it should be '/foo?bar=baz' not 'https://beta2.dev.meetup.com:8000/foo?bar=baz'). This will ensure that the server's 'host' (which will probably be some sort of Fargate internal host) isn't part of the redirect
2. querystring parameters need to stay intact

It looks like the `url.format` function that I originally used a couple years ago for this function has been deprecated - it's now part of the 'Legacy URL object' (https://nodejs.org/docs/latest-v8.x/api/url.html#url_legacy_urlobject). The `url.format` function being used here is actually a different interface built for proper `URL` objects, but offering different features.

SO, I believe this should be updated to:

```js
import { Url } from 'url'; // legacy URL interface

// ...

// need a redirect URI that does not include host, so using legacy Url interface
const redirectUrl = Url.format({ search: request.url.search, pathname })
return h.redirect(redirectUrl);
```
```

sadafie:
Tested and verified in mup-web

PR #635, opened 2019-08-14
❗️MONOREPO CONFIGS ❗️

sadafie:
This is an in-progress PR meant to support the `mupweb-app` package monorepo work in `mup-web`.

The intent of this PR is to introduce NO BREAKING CHANGES so that `pro-web` will continue to work without any required upgrades. If/when `pro-web` is ready to migrate, the branched/duplicate logic in this PR can be cleaned up.

Reference `mwp-cli` branch: https://github.com/meetup/mwp-cli/pull/115


sadafie:
Closing this ticket since we didn't need a MWP branch

PR #638, opened 2019-08-23
[PRELIM] Node 10 upgrade

sadafie:


.travis.yml
```diff
@@ -1,7 +1,7 @@
 sudo: false
 language: node_js
 node_js:
-    - '8.11'
+    - 10.16.3
```
mmcgahan: I'd like to at least allow patch updates, which can include important security updates, so could this be `'10.16'` (I think the quotes are necessary for YAML to parse it as a string instead of a float if there's only one decimal point)

.travis.yml
```diff
@@ -25,7 +25,7 @@ before_install:
 before_script:
     - lerna run build
     - git diff --exit-code yarn.lock
-    - yarn check
+    # - yarn check
```
mmcgahan: should this be uncommented before merge?

PR #644, opened 2019-10-16
Version Tag Response Header

sadafie:
PR branch for testing.


packages/mwp-app-server/src/util/index.js
```diff
@@ -27,6 +27,13 @@ export function onRequestExtension(request, h) {
 	return h.continue;
 }
 
+export function onPreResponseExtension(request, h) {
+	// response header corresponding to the application version tag generated by CI
```
mmcgahan: since this function doesn't control _how_ the env var is set, maybe just say that a version tag can be passed in as an env var and returned as a header

packages/mwp-app-server/src/util/index.js
```diff
@@ -39,6 +46,11 @@ export function registerExtensionEvents(server) {
 		method: onRequestExtension,
 	});
 
+	server.ext({
```
mmcgahan: small suggestion - put this in a `if (process.env.VERSION_TAG)` block?

sadafie:
Going to verify in prod

PR #645, opened 2019-10-18
Cloudwatch Log Cleanup

sadafie:
Reduce log noise to save $.


sadafie:
SV merge. We can always do follow up tickets to optimize more or add things back in.

PR #655, opened 2019-11-11
167407231 suppress click header

mmcgahan:
https://www.pivotaltracker.com/n/projects/2409221/stories/167407231

requests to the variants service sometimes have extra-large headers that result in a 500 error. This PR eliminates the largest cookie that is currently being forwarded - `click-track` from _all_ proxied API requests, because the click tracking data is consumed by the `mwp-tracking-plugin` and is not needed for any other request handling.

There are two things that we should watch after launching this in mup-web:

1. dramatic reduction in click tracking data appearing in our analytics as the result of no longer sending the click-track cookie to API endpoints. My understanding is that the API endpoints are not doing anything with the click tracking cookie, and so it is safe to remove it, but I will double check with the data team..

2. Sending much less data per request should reduce our 'outgoing data' payload from the API proxy, which could reduce costs and improve response times.

### DIFF
--- a/packages/mwp-api-proxy-plugin/src/util/send.js
+++ b/packages/mwp-api-proxy-plugin/src/util/send.js
@@ -9,6 +9,7 @@ import config from 'mwp-config';
 
 export const API_META_HEADER = 'X-Meta-Request-Headers';
 const FULL_URL_PATTERN = /^https?:\/\//;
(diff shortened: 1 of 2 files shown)


chenrui333:
Right now, the PT-github integration does not quite work, so I manually attach the PR to the ticket.

PR #658, opened 2019-12-10
Double write activity records

chenrui333:


chenrui333:
Close in favor of https://github.com/meetup/meetup-web-platform/pull/659

PR #659, opened 2019-12-12
Use nodejs Kinesis SDK to Perform analytics tracking double write

ChengzhiZhao:


ChengzhiZhao:
@chenrui333 I'm not familiar with `yarn check` gave some errors like `"espree#acorn" not installed`, I have tried `rm -rf node_modules && yarn` but doesn't seem to be working. We can check it on Monday.

PR #662, opened 2019-12-19
fix(tracking): fix_global_AWS_initialization

ChengzhiZhao:
Move AWS from singleton to function to initialize on each call

### DIFF
--- a/packages/mwp-tracking-plugin/src/util/avro.js
+++ b/packages/mwp-tracking-plugin/src/util/avro.js
@@ -2,7 +2,6 @@
 const log = require('mwp-logger-plugin').logger;
 const avro = require('avsc');
 const uuidv1 = require('uuid/v1');
-const AWS = require('aws-sdk');
 
 const canUsePubSub =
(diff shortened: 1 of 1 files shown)


packages/mwp-tracking-plugin/src/util/avro.js
```diff
@@ -42,17 +41,40 @@ const getPlatformAnalyticsLog = (
...
+			sts.assumeRole(
+				{
+					RoleArn: process.env.ASSUME_ROLE_ARN,
+					RoleSessionName: 'awssdk',
+				},
+				function(err, data) {
+					if (err) {
+						// an error occurred
+						console.log('Cannot assume role');
+						console.log(err, err.stack);
```
chenrui333: print stacktrace is not very cloudwatch friendly, it would spread into multiple lines.

ChengzhiZhao: I will update once it proved work.

PR #670, opened 2020-01-21
chore(deps): deprecate newrelic

chenrui333:

PR #679, opened 2020-03-06
fix(renderer): bring renderer part back after NewRelic cleanup

ekobtsev:
Brings back part of the renderer which shouldn't be removed in NewRelic removal

Related to this: https://github.com/meetup/meetup-web-platform/pull/670


softprops:
Oh crap did the new relic removal break page rendering?

softprops:
What was the impact?

ekobtsev:
@softprops I think that change was never deployed

chenrui333:
This might solve the mup-web side issue, which has struggle with local integration testing.

chenrui333:
I currently have a open PR for it.

PR #680, opened 2020-03-10
fix: 171722586 pass empty config object to UXCapture.create

andeliv:
Fixing: https://www.pivotaltracker.com/story/show/171722586

UXCapture.create waits for config object param


softprops:
Caused by https://github.com/meetup/meetup-web-platform/pull/670

PR #682, opened 2020-04-14
logging serialized to standard out in case we fail to write to kinesis

mattDenaci:
In cases when we can't log to kinesis, log message we wanted to send to standard out. This way, if kinesis becomes unavailable we can reprocess any tracking records we couldn't write to kinesis by reading them from cloudwatch logs.

### DIFF
--- a/packages/mwp-tracking-plugin/src/util/avro.js
+++ b/packages/mwp-tracking-plugin/src/util/avro.js
@@ -81,8 +81,12 @@ const getLogAWSKinesis = (): (string => Promise<void>) => {
 		});
 
 		kinesis.putRecord(options, function(err, data) {
-			if (err) console.log(err, err.stack);
-			else console.log(data);
+			if (err) {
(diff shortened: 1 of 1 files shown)


packages/mwp-tracking-plugin/src/util/avro.js
```diff
@@ -81,8 +81,12 @@ const getLogAWSKinesis = (): (string => Promise<void>) => {
 		});
 
 		kinesis.putRecord(options, function(err, data) {
-			if (err) console.log(err, err.stack);
-			else console.log(data);
+			if (err) {
+				console.log("Error sending message to Kinesis: "+ serializedRecord);
+				console.log(err, err.stack);
+			} else {
+				console.log("Kinesis recieved message: " + serializedRecord);
```
mattDenaci, line 88: I put this line in so we can test that the serialized record written out to cw is in the format that we expect/we have all the data to reprocess. I'll do a follow-up pr to reduce the volume of logs in the case when we successfully write to flume

PR #693, opened 2020-08-28
Apollo provider integration 174322636

komlev:
Adding ApolloProvider to mwp renderers so the ssr can be enabled on the mup-web

### DIFF
--- a/packages/mwp-app-render/src/components/ApolloProvider.jsx
+++ b/packages/mwp-app-render/src/components/ApolloProvider.jsx
@@ -4,19 +4,32 @@ import fetch from 'isomorphic-fetch';
 import { ApolloProvider } from 'react-apollo';
 import { isServer } from '../util/isServer';
 
-const client = new ApolloClient({
-	ssrMode: isServer(),
-	name: 'mup-web',
-	cache: isServer()
-		? new InMemoryCache()
-		: new InMemoryCache().restore(window.__APOLLO_STATE__),
-	uri: 'https://api.meetup.com/gql',
-	credentials: 'include',
-	fetch,
-});
+let cachedClient;
+
+const getClient = () => {
+	if (!cachedClient || isServer()) {
+		const options = {
+			ssrMode: isServer(),
+			name: 'mup-web',
+			cache: isServer()
+				? new InMemoryCache()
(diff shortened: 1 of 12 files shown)


packages/mwp-app-render/src/components/ApolloProvider.jsx
```diff
@@ -0,0 +1,22 @@
...
+import { ApolloProvider } from 'react-apollo';
+import { isServer } from '../util/isServer';
+
+const client = new ApolloClient({
+	ssrMode: isServer(),
+	name: 'mup-web',
+	cache: isServer()
+		? new InMemoryCache()
+		: new InMemoryCache().restore(window.__APOLLO_STATE__),
+	uri: 'https://api.meetup.com/gql',
```
cranberyxl: This should probably be an ENV var so support testing against local gql

packages/mwp-app-render/src/components/ApolloProvider.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import ApolloClient, { InMemoryCache } from 'apollo-boost';
+import fetch from 'isomorphic-fetch';
+import { ApolloProvider } from 'react-apollo';
+import { isServer } from '../util/isServer';
+
+const client = new ApolloClient({
+	ssrMode: isServer(),
+	name: 'mup-web',
+	cache: isServer()
```
cranberyxl: I think you need to make a new client on search server invocation to avoid a memory leak

packages/mwp-app-render/src/components/ServerApp.jsx
```diff
@@ -22,11 +23,13 @@ class ServerApp extends React.Component {
 				context={routerContext}
 			>
 				<HapiCookieProvider request={request} h={h}>
-					<PlatformApp
-						appContext={appContext}
-						store={store}
-						routes={routes}
-					/>
+					<ApolloProvider>
```
cranberyxl, line 26: maybe here you pass something to make a new client?

cranberyxl: instead of having `isServer` you can pass `server={true}` here

mkleymenov (approved):

PR #695, opened 2020-09-09
174334143: add SameSite policy to click-track cookie.

altuxovsergey:
Pivotal: https://www.pivotaltracker.com/n/projects/2409221/stories/174334143

**Context**
Chrome started enforcing SameSite cookie policy since version 80. Now any cookie that doesn't explicitly specify SameSite attribute is defaulted to SameSite=Lax, which prevents a browser from sending that cookie in cross-domain requests.

Added `SameSite` and `Secure` attributes for `click-track` cookie.

According to [this](https://github.com/js-cookie/js-cookie/issues/276) thread `js-cookie` lib should support custom attributes since version 2.2.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [25.1]
+
+-   **Improved** add `SameSite` policy to click-track cookie
+
 ## [25.0]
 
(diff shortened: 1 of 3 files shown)


mkleymenov (approved):

chenrui333 (approved):

chenrui333:
Bump version is needed for this PR (basically changelog update)

PR #700, opened 2020-09-25
fix: 174580981 lang order issue fix

kuhdm0:


andeliv (approved):

packages/mwp-config/locales.js
```diff
@@ -1,6 +1,6 @@
 module.exports = [
-	'de-DE',
 	'en-US',
```
chenrui333, line 2: why is this an issue? just curious.

PR #709, opened 2020-11-23
Updates web-components deps node version

Mikodin:
Fixes:
https://www.pivotaltracker.com/story/show/175624596


chenrui333:
just reminder that the version needs to be bumped as well.

chenrui333:
https://github.com/meetup/meetup-web-platform/blob/master/Makefile#L2
and release note

PR #712, opened 2020-12-11
feat: 175933240 hide hreflang & change canonical for group locale

kuhdm0:
SEO optimization for groups pages when `group locale` is set:
1. Canonical URLs locale is taken from group locale, not from user locale https://www.pivotaltracker.com/story/show/175933240
2. Hreflang tag removed https://www.pivotaltracker.com/story/show/175933177


packages/mwp-app-render/src/util/seo/links.js
```diff
@@ -5,73 +5,89 @@ import locales from 'mwp-config/locales';
...
+	let result = [
 		<link
 			rel="canonical"
-			href={generateCanonicalUrl(baseUrl, localeCode, route)}
+			href={generateCanonicalUrl(baseUrl, localeCode, route, groupLocaleCode)}
 			key="canonical"
 		/>,
 	];
+
+	if (!groupLocaleCode) {
```
andeliv, line 43: Let's maybe use reverse logic here and return early if `groupLocaleCode` is set. That will remove unnecessary nesting

packages/mwp-app-render/src/components/SEOHead.jsx
```diff
@@ -42,13 +44,13 @@ export const SEOHeadComponent = ({
...
 	route,
 }) => {
-
 	const metaData = generateMetaData({
 		appPath: `meetup:/${route}`,
 		baseUrl,
 		pageDescription,
 		imageUrl,
 		localeCode,
+		groupLocaleCode,
```
andeliv, line 53: I couldn't find where signature of this method changed. Do we need to pass `groupLocaleCode` here?

packages/mwp-app-render/src/components/SEOHead.jsx
```diff
@@ -33,6 +34,7 @@ export const SEOHeadComponent = ({
 	imageUrl,
 	ldJson,
 	localeCode,
+	groupLocaleCode,
```
andeliv, line 37: What do think about naming this in a more abstract way like `forcedLocaleCode`?

andeliv (approved):

PR #723, opened 2021-05-26
update apollo graphql edge endpoint to use www. domain by default

softprops:
we've already adopted this approach in build-meetup with great success. 

this change improves page performance but reducing network requests, increases response time as we don't have to wait before we can start executing the intended requests and reduces invocation load on graphql-edge-api

### DIFF
--- a/packages/mwp-app-render/src/util/getClient.js
+++ b/packages/mwp-app-render/src/util/getClient.js
@@ -1,7 +1,7 @@
 import ApolloClient, { InMemoryCache } from 'apollo-boost';
 import fetch from 'isomorphic-fetch';
 
-const GRAPHQL_ENDPOINT = process.env.GRAPHQL_ENDPOINT || 'https://api.meetup.com/gql';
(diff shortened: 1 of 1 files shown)


packages/mwp-app-render/src/util/getClient.js
```diff
@@ -1,7 +1,7 @@
 import ApolloClient, { InMemoryCache } from 'apollo-boost';
 import fetch from 'isomorphic-fetch';
 
-const GRAPHQL_ENDPOINT = process.env.GRAPHQL_ENDPOINT || 'https://api.meetup.com/gql';
+const GRAPHQL_ENDPOINT = process.env.GRAPHQL_ENDPOINT || 'https://www.meetup.com/gql';
```
softprops, line 4: note: I've tried just setting `GRAPHQL_ENDPOINT`  in mup-web but it doesn't work for client side apollo requests

komlev (approved):

PR #733, opened 2021-09-28
Fe 10016 replace http by https

95kach:
https://app.shortcut.com/meetup/story/10016/replace-http-with-https-links-for-messages-page-pages-with-structured-data

### DIFF
--- a/packages/mwp-core/src/util/linkify.js
+++ b/packages/mwp-core/src/util/linkify.js
@@ -23,6 +23,16 @@ const getRelAttr = (followedExternalDomains, link, target) => {
 	}
 };
 
+export const getSubDomain = url => {
+	if (!url || !url.includes('//')) {
+		return 'www';
+	}
+
+	const [, domain] = url.split('//');
+	const [, , subdomain] = domain.split('.').reverse();
+	return subdomain || 'www';
+};
+
 /**
  * Generates HTML link tag element with target
  *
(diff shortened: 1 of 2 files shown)


packages/mwp-core/src/util/linkify.js
```diff
@@ -23,6 +23,16 @@ const getRelAttr = (followedExternalDomains, link, target) => {
 	}
 };
 
+export const getSubDomain = (url: string) => {
+	if (!url || !url.includes('//')) {
+		return 'www';
+	}
+
+	const [, domain] = url.split('//');
+	const [, , subdomain] = domain.split('.').reverse();
```
andeliv, line 32: Will it work if we just use first value from `domain.split('.')`? `const [subdomain] = domain.split('.');`

packages/mwp-core/src/util/linkify.js
```diff
@@ -37,8 +47,8 @@ const createLink = (options: Object, followedExternalDomains?: Array<string>) =>
 	const target = options.target || '';
 	const targetAttr = `target="${target}"`;
 	const hasProtocolRE = /^(?:https?:|ws:|ftp:)?\/\//;
-	const hasMeetupHttpLinkRE = /http:\/\/www.meetup/g;
-	const meetupHttps = 'https://www.meetup';
+	const hasMeetupHttpLinkRE = /https?:\/\/(.+?\.)?meetup/g;
```
andeliv, line 50: How about using more explicit check here: /https?:\/\/([a-z0-9-]+?\.)?meetup/g;

andeliv (approved):

PR #753, opened 2021-12-03
feat(context): remove duotone from api

atrout:
We are removing the dynamic scaler, which provides duotoned images; these images are no longer used on the site.


chenrui333 (approved):

chenrui333:
maybe we should also bump the major version (as this is a breaking change)

softprops (approved):
This is very exciting. Is this a signal that all of this might go away as well?

https://github.com/meetup/meetup/search?q=duotone&type=

chenrui333 (approved):
🚀  ship it

PR #791, opened 2022-02-16
docs: call the internal variant endpoint

ew-meetup:
The `-ext` endpoint is mostly intended for testing and special cases where going out through the internet is unavoidable (e.g. mobile clients making direct calls to variant).

You'll need to confirm that networking is set up from mup-web (this client) and the "data realtime services" account. Account details are in https://github.com/meetup/variant/#where-is-it

Thank you for considering switching off of variant-ext endpoint.


chenrui333 (approved):

ew-meetup:
@chenrui333 , is this now using the internal endpoint? (I only modified the docs as a way of requesting that it use the internal one).

PR #844, opened 2022-04-25
Update tracking query - debugging

meetup-anton-hulak:


packages/mwp-tracking-plugin/src/util/avro.js
```diff
@@ -25,6 +25,28 @@ const getCrossAccountCredentials = async () => {
...
+			Promise.resolve();
+		};
+	} else {
+		return async (serializedRecord: string) => {
+			try {
+				const encodedRecord = Buffer.from(serializedRecord).toString(
+					'base64'
+				);
+				const img = new window.Image();
+				img.src = `https://analytics-tracking.meetup.com/data?records=${encodedRecord}`;
```
softprops, line 41: 👍

packages/mwp-tracking-plugin/src/util/avro.js
```diff
@@ -194,6 +216,37 @@ const chapinEnvelopeDeserializer: Object => Deserializer = schema => {
...
+			schemaVersion: schema.doc,
+			source: analyticsSource,
+			data: JSON.stringify(record),
+		};
+		const payload = {
+			records: [analytics],
+			viewId,
+			...metadata,
+		};
+		const encodingForNewPattern = JSON.stringify(payload);
```
softprops, line 237: not a blocker but just noting a repeating lesson we don't often seen to learn. "new" is a relative term and leaving New in a name will be very confusing for future engineers that might find this if when new becomes old

tcreedmeetup: that's a very good point 😁

